PR #39, opened 2018-10-02
Revise root Dockerfile

benjaminapetersen:
Cleanup & updating to match some other root Dockerfiles

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,41 +1,23 @@
-# WIP
-#
-# - how does this relate to the file in /tmp/build/ ?
-# - how should it relate to Make commands?
-#   - example(s):
-#     - https://github.com/openshift/cluster-image-registry-operator/blob/master/Makefile
-#   - possible commands:
-#     - make build  # build go app into binary
-#     - make build-image # build container image with included binary
-#     - make build-all # build binary + docker image?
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,41 +1,21 @@
...
 
-RUN useradd console-operator
-USER console-operator
+# finally to enable install integration
+#LABEL io.openshift.release.operator true
 
-ENTRYPOINT []
-CMD ["/usr/bin/console-operator"]
\ No newline at end of file
+# entrypoint specified in 03-operator.yaml as `console-operator`
```
spadgett: I'd probably leave `CMD` here

benjaminapetersen: Works for me, updated.

Dockerfile
```diff
@@ -1,41 +1,22 @@
...
+CMD console-operator
```
spadgett: Better to make it an array with the full path as before

benjaminapetersen: Its now building in the root dir I believe, but I can still update to array.

benjaminapetersen: ah no i did put it back in `/usr/bin`.

benjaminapetersen: testing this now

benjaminapetersen: Quotes fixed.  Heh.

PR #42, opened 2018-10-03
Update rbac for oauth and route creation

benjaminapetersen:
Adjusting `rbac.yaml` for deploying the operator.  `operator-sdk up local` permissions were a little more relaxed, deploying the operator via `oc create -f deploy/*` requires these changes for the oauth client & route to be created.

### DIFF
--- a/deploy/02-sa.yaml
+++ b/deploy/02-sa.yaml
@@ -1,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: console-operator
+  namespace: openshift-console
--- a/deploy/04-rbac-rolebinding.yaml
+++ b/deploy/04-rbac-rolebinding.yaml
@@ -1,0 +1,25 @@
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: default-account-console-operator
+  namespace: openshift-console
+subjects:
+- kind: ServiceAccount
+  name: console-operator
+roleRef:
+  kind: Role
+  name: console-operator
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: console-operator
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: console-operator
+subjects:
+- kind: ServiceAccount
+  name: console-operator
+  namespace: openshift-console
--- a/deploy/05-operator.yaml
+++ b/deploy/05-operator.yaml
@@ -13,6 +13,7 @@ spec:
       labels:
         name: console-operator
     spec:
+      serviceAccountName: console-operator
       containers:
       - name: console-operator
         image: openshift/origin-console-operator:latest
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,8 +12,10 @@ COPY --from=builder /go/src/github.com/openshift/console-operator/tmp/_output/bi
 # these manifests are necessary for the installer
 COPY deploy/00-crd.yaml \
     deploy/01-namespace.yaml \
-    deploy/02-rbac.yaml \
-    deploy/03-operator.yaml \
+    deploy/02-sa.yaml \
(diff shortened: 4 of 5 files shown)


deploy/02-rbac.yaml
```diff
@@ -31,9 +31,25 @@ rules:
...
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: console-operator
+rules:
+- apiGroups:
+  - oauth.openshift.io
+  resources:
+  - oauthclients
+  verbs:
+  - "*"
```
enj: No

benjaminapetersen: So pleasant Mo. :)

deploy/02-rbac.yaml
```diff
@@ -31,9 +31,25 @@ rules:
   - statefulsets
```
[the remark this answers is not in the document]
benjaminapetersen: I should trim this, a lot of it is default `sdk` generated.  Let me clean it up.

deploy/02-sa.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
```
benjaminapetersen: Creating a separate service account for `console-operator` and updating bindings below to use it.

benjaminapetersen:
@spadgett @enj ready for another look.
@enj I know you want me to rip out the `oauth` entirely & create it as a `yaml` so that the `operator` isn't responsible for its own `oauth` client.  I'll add that as a follow-on, I don't want to get in a non-working state for tomorrow's demo if that gets messy.

deploy/04-rbac-rolebinding.yaml
```diff
@@ -0,0 +1,27 @@
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: default-account-console-operator
+  namespace: openshift-console
+subjects:
+- kind: ServiceAccount
+  # name: console-operator
```
benjaminapetersen: the `sdk` struggles to do things if I change this, looking into it as a follow-on as well.

spadgett:
> I know you want me to rip out the oauth entirely & create it as a yaml so that the operator isn't responsible for its own oauth client.

@enj How is the OAuth client secret generated and how does the console know what the secret is?

deploy/05-operator.yaml
```diff
@@ -15,7 +15,9 @@ spec:
     spec:
       containers:
       - name: console-operator
-        image: openshift/origin-console-operator:latest
+        # image: openshift/origin-console-operator:latest
+        # image: hogwarts.io/gryffindor/expecto-patronus:v4.0.2
+        image: quay.io/benjaminapetersen/console-operator:v4.0.0
```
spadgett: Was this an accident?

pkg/operator/deployment.go
```diff
@@ -44,8 +48,9 @@ func newConsoleDeployment(cr *v1alpha1.Console) *appsv1.Deployment {
 				},
 				Spec: corev1.PodSpec{
 					// NodeSelector:  corev1.NodeSelector{},
-					RestartPolicy: "Always",
-					SchedulerName: "default-scheduler",
+					ServiceAccountName: ServiceAccountName,
```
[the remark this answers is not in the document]
spadgett: It's not the console that should use this service account, but the operator itself.

spadgett:
@enj We also need to be able to update the redirect URLs in the OAuth client when the hostname changes.

deploy/05-operator.yaml
benjaminapetersen: Yeah fixed right after I pushed. Sorry, I'll use a different file for my local dev/debugging.

pkg/operator/deployment.go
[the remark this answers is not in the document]
benjaminapetersen: Got it, yeah that makes sense.

benjaminapetersen:
@spadgett makes sense, my mistake.  It's working as expected again with `serviceAccountName: console-operator` added to the `operator.yaml`.

benjaminapetersen:
ill create an issue for the `oauth` client.

PR #43, opened 2018-10-03
Redact shared secret string in logging

benjaminapetersen:
A random secret string is generated & shared between the oauth client & a Secret.  This redacts most of the logging of the random bytes, but leaves a few characters for debugging purposes. 

screen shot 2018-10-02 at 11 16 36 pm

### DIFF
--- a/pkg/operator/oauth.go
+++ b/pkg/operator/oauth.go
@@ -96,14 +96,14 @@ func CreateOAuthClient(cr *v1alpha1.Console, rt *routev1.Route) (*oauthv1.OAuthC
 		logrus.Errorf("failed to create console oauth client : %v", err)
 		return nil, nil, err
 	} else {
-		logrus.Info("created console oauth client with secret ", randomBits)
+		logrus.Info("created console oauth client with secret")
 	}
 
 	if err := sdk.Create(authSecret); err != nil && !errors.IsAlreadyExists(err) {
 		logrus.Errorf("failed to create console oauth client secret : %v", err)
 		return nil, nil, err
 	} else {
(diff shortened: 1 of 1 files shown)


pkg/operator/oauth.go
```diff
@@ -96,14 +96,14 @@ func CreateOAuthClient(cr *v1alpha1.Console, rt *routev1.Route) (*oauthv1.OAuthC
 		logrus.Errorf("failed to create console oauth client : %v", err)
 		return nil, nil, err
 	} else {
-		logrus.Info("created console oauth client with secret ", randomBits)
+		logrus.Infof("created console oauth client with secret: %v...", randomBits[0:9])
```
spadgett: I'm worried someone might change the helper method to create a shorter secret at some point, and then we're accidentally logging the entire value. I think it's better to not log any part of the secret.

benjaminapetersen: Perhaps we have a `loglevel` toggle?  If they secrets mismatch, things don't work.  It would be a hassle to change source code for debugging.  I don't disagree, though it would probably be incorrect to shorten the secret.

spadgett: I don't think we should log any part of the secret regardless of log level.

They should always match here, right? Unless the create fails?

benjaminapetersen: they should always match, unless something goes funny.  Currently I think I always update them if anything is out of sync, w/ a new secret, every time.  So if someone meddled with one or the other they should re-sync.  

I'll nix it.

benjaminapetersen:
Updated

PR #49, opened 2018-10-11
Move /deploy to /manifests, update RBAC rules

benjaminapetersen:
- Moves `/deploy` to `/manifests`
- Leaves `/deploy/cr.yaml` in `/deploy` as it ought not be automatically created
- Updates RBAC to use specific/correct rules

@enj


manifests/02-operator.yaml
```diff
@@ -29,4 +29,4 @@ spec:
             fieldRef:
               fieldPath: metadata.namespace
         - name: OPERATOR_NAME
-          value: "console-operator"
+          value: "console-operator"
```
enj: Looks accidental.

manifests/00-rbac-role.yaml
```diff
@@ -1,40 +1,59 @@
...
   - console.openshift.io
   resources:
-  - "*"
+  - consoles
   verbs:
-  - "*"
+  - get
+  - list
+  - watch
+  - create
```
enj: Why?

manifests/00-rbac-role.yaml
```diff
@@ -1,40 +1,59 @@
...
+  - update
+  - delete
```
enj: Add TODO to drop all the deletes

manifests/00-rbac-role.yaml
benjaminapetersen: Why....? Whats the full question.

manifests/00-rbac-role.yaml
benjaminapetersen: added (above, line 7, as it applies to many)

PR #52, opened 2018-10-15
Add manifests/image-references file

benjaminapetersen:
I'm seeing this file crop up in the `/manifests` dir, so creating it, but need to verify its purpose before removing WIP.

An example:
- https://github.com/openshift/cluster-image-registry-operator/blob/master/deploy/image-references

### DIFF
--- a/manifests/image-references
+++ b/manifests/image-references
@@ -1,0 +1,8 @@
+kind: ImageStream
+apiVersion: image.openshift.io/v1
+spec:
+  tags:
+  - name: console-operator
+    from:
+      kind: DockerImage
+      name: docker.io/openshift/origin-console-operator:latest


benjaminapetersen:
/retest

curious if these are still flaking.

spadgett:
> I'm seeing this file crop up in the /manifests dir, so creating it, but need to verify its purpose before removing WIP.

@bparees any insight?

bparees:
it's something the installer/CVO needs.  Check w/ the installer team if you want a validation/sign-off on the content.

benjaminapetersen:
@spadgett should be good for a review then.

manifests/image-references
```diff
@@ -0,0 +1,8 @@
+kind: ImageStream
+apiVersion: image.openshift.io/v1
+spec:
+  tags:
+  - name: console-operator
+    from:
+      kind: DockerImage
+      name: docker.io/openshift/openshift/origin-console-operator:latest
```
spadgett: Looks like you have an extra `openshift` in there

PR #56, opened 2018-10-16
Create Console CR when Operator starts if not present

benjaminapetersen:
Ensures the operator will immediately create a console if a `Console` resource is not found.

### DIFF
--- a/cmd/console-operator/main.go
+++ b/cmd/console-operator/main.go
@@ -23,6 +23,7 @@ import (
 
 	// operator
 	consoleapi "github.com/openshift/console-operator/pkg/apis/console/v1alpha1"
+	"github.com/openshift/console-operator/pkg/operator"
 )
 
 func printVersion() {
@@ -41,7 +42,8 @@ func init() {
 }
 
 func main() {
-	flag.Parse() // suppress an annoying notification :/
+	defaultConsoleFlag := flag.Bool("create-default-console", false, "Create a Console custom resource if `true`")
+	flag.Parse()
 	printVersion()
 	// prometheus metrics
 	sdk.ExposeMetricsPort()
@@ -60,7 +62,12 @@ func main() {
 	watch(appsv1.SchemeGroupVersion.String(), "Deployment", namespace, resyncPeriod)
 
 	watch(consoleapi.SchemeGroupVersion.String(), "Console", namespace, resyncPeriod)
-	// i should be watching the oauth client i made :)
+
+	if *defaultConsoleFlag {
+		if _, err := operator.ApplyConsole(); err != nil {
+			logrus.Fatalf("Failed to create default console %v", err)
+		}
+	}
 
 	sdk.Handle(
 		stub.NewFilteredHandler(
--- a/manifests/05-operator.yaml
+++ b/manifests/05-operator.yaml
@@ -16,12 +16,14 @@ spec:
       serviceAccountName: console-operator
       containers:
       - name: console-operator
-        image: openshift/origin-console-operator:latest
+        image: docker.io/openshift/origin-console-operator:latest
         ports:
         - containerPort: 60000
           name: metrics
         command:
         - console-operator
+        args:
+        - "-create-default-console=true"
         imagePullPolicy: Always
         env:
         - name: WATCH_NAMESPACE
--- a/pkg/operator/console.go
+++ b/pkg/operator/console.go
@@ -1,0 +1,35 @@
+package operator
+
+import (
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/operator-framework/operator-sdk/pkg/sdk"
+
+	"github.com/openshift/console-operator/pkg/apis/console/v1alpha1"
+)
+
+func defaultConsole() *v1alpha1.Console {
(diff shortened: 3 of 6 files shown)


benjaminapetersen:
@spadgett @deads2k @enj

alecmerdler:
AFAIK, this is not how Operators are philosophically supposed to work. _Users_ create CRs, as this is the API for the application. This couples the installation of the Operator with the creation of the underlying app. Think of it like AWS: when they add a new service like RDS, a database isn't immediately provisioned for you, you as the user make that decision.

At the very least it should be a command-line flag (`--create-default-console=true`).

benjaminapetersen:
@alecmerdler I think you have it right, though I believe for our use case of using operators to stamp out the platform itself, the console is one of the "special" ones that is supposed to ensure something runs right away.  I'll make sure to double check this.

jwforres:
As ben says, some of the platform "cluster" operators are special in this regard, there is only one of them and we want there to be a CR by default on install. But I have no problem with Alec's suggestion to make this a flag, since there have been discussions about letting people use the console operator to spin up test consoles on a cluster.

enj:
If you add a flag, it will need some default value or need to be required.

spadgett:
Maybe default to false and add the arg in the operator deployment manifest?

manifests/05-operator.yaml
```diff
@@ -16,12 +16,13 @@ spec:
...
       containers:
       - name: console-operator
-        image: openshift/origin-console-operator:latest
+        image: docker.io/openshift/origin-console-operator:latest
         ports:
         - containerPort: 60000
           name: metrics
         command:
         - console-operator
+        args: ["-create-default-console=true"]
```
benjaminapetersen: `oc create -f manifests/` will now deploy an operator, which will automatically create a console if it does not exist.

pkg/operator/console.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+func ApplyConsole() (*v1alpha1.Console, error) {
+	console := defaultConsole()
+	if err := sdk.Get(console); errors.IsNotFound(err) {
+		if err = sdk.Create(console); err != nil {
+			return nil, err
+		}
+	}
+	return console, nil
```
spadgett: I'm not familiar with the SDK... Will this return the right value if the console already exists?

pkg/operator/console.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/operator-framework/operator-sdk/pkg/sdk"
+
+	"github.com/openshift/console-operator/pkg/apis/console/v1alpha1"
+)
+
+func defaultConsole() *v1alpha1.Console {
+	console := &v1alpha1.Console{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "console.openshift.io/v1alpha1",
```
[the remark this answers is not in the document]
benjaminapetersen: I don't know when we want to bump to `beta`.  Open to doing it anytime as a follow as I'd prob have to update all the packages.

pkg/operator/console.go
benjaminapetersen: Flow here is:
- `defaultConsole()` provides enough to make a `Get` request
- if got, it mutates `console` (dislike this about the api) and returns it.
- if not, it `Create`'s a `console`, still mutates `console`, and returns it.

I think the mutation is confusing and dislike that about the API.

cmd/console-operator/main.go
```diff
@@ -60,7 +62,10 @@ func main() {
 	watch(appsv1.SchemeGroupVersion.String(), "Deployment", namespace, resyncPeriod)
 
 	watch(consoleapi.SchemeGroupVersion.String(), "Console", namespace, resyncPeriod)
-	// i should be watching the oauth client i made :)
+
+	if *defaultConsoleFlag {
+		operator.ApplyConsole()
```
spadgett: Error handling?

pkg/operator/console.go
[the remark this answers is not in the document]
benjaminapetersen: Looked at a few other operators, all still `v1alpha1` at present.

manifests/05-operator.yaml
benjaminapetersen: I'll do that, agree!

cmd/console-operator/main.go
```diff
@@ -4,6 +4,7 @@ import (
 	// standard lib
 	"context"
 	"flag"
+	"github.com/openshift/console-operator/pkg/operator"
```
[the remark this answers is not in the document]
benjaminapetersen: will do, i need to clean up imports around here.

cmd/console-operator/main.go
```diff
@@ -60,7 +62,12 @@ func main() {
 	watch(appsv1.SchemeGroupVersion.String(), "Deployment", namespace, resyncPeriod)
 
 	watch(consoleapi.SchemeGroupVersion.String(), "Console", namespace, resyncPeriod)
-	// i should be watching the oauth client i made :)
+
+	if *defaultConsoleFlag {
+		if _, err := operator.ApplyConsole(); err != nil {
```
benjaminapetersen: @spadgett addresses the error handling. thx!

benjaminapetersen: I think this might be the last item.

PR #59, opened 2018-10-19
Add ManagementState to handle paused, unpaused, deleted

benjaminapetersen:
Add ManagementState to handle paused, unpaused, deleted  as managed, unmanaaged, removed.

### DIFF
--- a/examples/cr.yaml
+++ b/examples/cr.yaml
@@ -1,29 +1,23 @@
-# the console-operator will create a console if one does not exist
-# this is an example of a CR for a console resource
+# The console-operator will create a console automatically on startup if one does not exist.
+# This file is an example of a console resource.
+# You may edit this file and `oc apply -f cr.yaml` to make changes to the console.
 apiVersion: "console.openshift.io/v1alpha1"
 kind: "Console"
 metadata:
-  # do not change the name, this is required to match within the operator
+  # Do not change the name, the operator explicitly requires this name to match as the console is a singleton.
   name: openshift-console
-  # namespace: openshift-console
-  # for dev:
   namespace: openshift-console
 spec:
-  paused: false
-  # what needs to be customized here?
-  # fewest knobs possible!
-  # this will need to be reflected in the
-  # pkg/apis/console/v1alpha1/types.go file
-  # count is # of replicas
+  # openshift/api/operator/v1alpha1/types.go
+  # managementState indicates whether the operator is:
+  # - Managed indicates the operator fully manages the state of the console and its sub-resources
+  # - Unmanaged indicates that the operator should idle and ignore the console and its sub-resources
+  # - Removed indicates that the operator should delete all of the sub-resources of the console
+  managementState: Managed
+  # there should be only one console at this time
   count: 1
   # this is the console image
   baseImage: docker.io/openshift/origin-console
-  # note: version with a @sha does not work, @ is not valid starting a string in yaml
   version: latest
-  # location of the master should NOT be needed, the console should automatically use
-  # the service for the cluster API w/o us having to declare anything
-  # masterPublicURL:
-  # the log level for the operator itself.
-  # note: should there be a lvl for the components as well?
   logging:
     level: 0
--- a/pkg/apis/console/v1alpha1/types.go
+++ b/pkg/apis/console/v1alpha1/types.go
@@ -5,6 +5,7 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
 	routev1 "github.com/openshift/api/route/v1"
 
 	"github.com/operator-framework/operator-sdk/pkg/sdk"
@@ -16,7 +17,6 @@ const (
 )
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
-
 type ConsoleList struct {
 	metav1.TypeMeta `json:",inline"`
 	metav1.ListMeta `json:"metadata"`
@@ -24,7 +24,6 @@ type ConsoleList struct {
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
-
 type Console struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata"`
@@ -35,7 +34,8 @@ type Console struct {
 // if/when changed, be sure to regenerate generated code:
 // 	operator-sdk generate k8s
 type ConsoleSpec struct {
-	Paused bool `json:"paused"`
+	// ManagementState must be one of the above values to describe how the operator should behave
+	ManagementState operatorv1alpha1.ManagementState `json:"managementState"`
 	// Count is the number of Console replicas
 	Count     int32  `json:"count,omitempty"`
 	BaseImage string `json:"baseImage"`
@@ -45,8 +45,8 @@ type ConsoleSpec struct {
 	// https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/generate/podtemplatespec.go#L21
 	Logging *LoggingConfig `json:"logging"`
 }
+
 type ConsoleStatus struct {
-	Paused bool `json:"paused"`
 	// the hostname assigned by the cluster after the route is created
 	Host string `json:"host"`
 }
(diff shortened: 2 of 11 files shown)


benjaminapetersen:
/assign @enj 

@jhadvig just FYI (will start pinging you on these for context)

pkg/operator/route.go
```diff
@@ -65,3 +65,11 @@ func ApplyRoute(cr *v1alpha1.Console) (*routev1.Route, error) {
 	}
 	return rt, nil
 }
+
+func DeleteRoute(cr *v1alpha1.Console) error {
+	rt := newConsoleRoute(cr)
+	if err := sdk.Delete(rt); err != nil {
```
enj: `return sdk.Delete(...)`

examples/cr.yaml
```diff
@@ -9,7 +9,12 @@ metadata:
   # for dev:
   namespace: openshift-console
 spec:
-  paused: false
+  # openshift/api/operator/v1alpha1/types.go
+  # managementState indicates whether the operator is managing the CR (managed),
+  # ignoring the CR(paused) or if the CR is deleted (removed)
+  managementState: managed
```
alecmerdler: This makes more sense as a boolean. How would the `removed` value be used if the CR itself is deleted?

pkg/operator/reconcile.go
```diff
@@ -55,7 +55,29 @@ func ReconcileConsole(cr *v1alpha1.Console) error {
 	if _, _, err := ApplyOAuth(cr, rt); err != nil {
 		return err
 	}
-
 	return nil
+}
+
+func DeleteAllResources(cr *v1alpha1.Console) error {
```
enj: aggregate errors, do not exit early, filter out `NotFound`

pkg/operator/reconcile.go
```diff
@@ -55,7 +55,29 @@ func ReconcileConsole(cr *v1alpha1.Console) error {
...
+	if err := DeleteRoute(cr); err != nil {
+		return err
+	}
+	if err := DeleteConfigMap(cr); err != nil {
+		return err
+	}
+	if err := DeleteDeployment(cr); err != nil {
+		return err
+	}
+	if err := DeleteOAuthSecret(cr); err != nil {
```
enj: Remove, you did not create, you do not delete.

pkg/operator/oauth.go
```diff
@@ -62,6 +64,31 @@ func ApplyOAuth(cr *v1alpha1.Console, rt *routev1.Route) (*oauthv1.OAuthClient,
 	return authClient, authConfigSecret, nil
 }
 
+func DeleteOAuthSecret(cr *v1alpha1.Console) error {
```
enj: Remove.

pkg/operator/deployment.go
```diff
@@ -191,3 +191,11 @@ func ApplyDeployment(cr *v1alpha1.Console) (*appsv1.Deployment, error) {
 	}
 	return d, nil
 }
+
+func DeleteDeployment(cr *v1alpha1.Console) error {
+	d := newConsoleDeployment(cr)
+	if err := sdk.Delete(d); err != nil {
```
enj: `return sdk.Delete(...)`

pkg/operator/configmap.go
```diff
@@ -162,3 +166,11 @@ func ApplyConfigMap(cr *v1alpha1.Console, rt *routev1.Route) (*corev1.ConfigMap,
 	}
 	return configMap, nil
 }
+
+func DeleteConfigMap(cr *v1alpha1.Console) error {
+	configMap := newConsoleConfigMap(cr, nil)
+	if err := sdk.Delete(configMap); err != nil {
```
enj: `return sdk.Delete(...)`

examples/cr.yaml
enj: Remove means clean up resources you created and pause.  Deleting the CR will cause the GC to clean up everything else.

alecmerdler: Why is there a distinction? That's the point of using `ownerReferences`.

benjaminapetersen: This is standardized across a bunch (all?) of the operators now.

The `ConsoleOperator` -> `Console` relationship is like a `Deployment` to a `Pod` relationship.  If you delete a pod, the deployment recreates it immediately.  the only way to actually delete is to set its `ManagementState` to `Removed`, then it will delete all the sub resources. A lot like `Replicas` `0`.

pkg/operator/reconcile.go
benjaminapetersen: I do create this secret.  I do not create the client.

pkg/operator/oauth.go
benjaminapetersen: I do create this secret & own it. 
I do not own the client.

enj: Sorry, side effect of reviewing too quickly.

pkg/apis/console/v1alpha1/types.go
```diff
@@ -15,6 +15,17 @@ const (
 	defaultVersion   = "latest"
 )
 
+type ManagementState string
```
enj: Why is this duped instead of imported from `openshift/api`?

benjaminapetersen: Because its just a const.  I can fix that, however.

pkg/apis/console/v1alpha1/types.go
```diff
@@ -83,6 +96,11 @@ func (c *Console) SetDefaults() bool {
 		c.Spec.Logging = &LoggingConfig{
 			Level: 0,
 		}
+		changed = true
+	}
+	if c.Spec.ManagementState == "" {
```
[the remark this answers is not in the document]
benjaminapetersen: will do.

pkg/operator/reconcile.go
```diff
@@ -2,6 +2,8 @@ package operator
 
 import (
 	"github.com/openshift/console-operator/pkg/apis/console/v1alpha1"
+	errors "k8s.io/apimachinery/pkg/api/errors"
```
enj: nit import separation

pkg/operator/reconcile.go
```diff
@@ -55,7 +57,31 @@ func ReconcileConsole(cr *v1alpha1.Console) error {
...
 		return err
 	}
-
 	return nil
+}
+
+func DeleteAllResources(cr *v1alpha1.Console) error {
+	var errs []error
+
+	if err := DeleteService(cr); errors.IsNotFound(err) {
```
enj: You have this backwards, you want `!errors.IsNotFound`

but better approach would be to always append, and then `return errutil.FilterOut(errutil.NewAggregate(errs), errors.IsNotFound)`

pkg/apis/console/v1alpha1/types.go
```diff
@@ -15,16 +16,23 @@ const (
 	defaultVersion   = "latest"
 )
 
-// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+const (
```
enj: I think you forgot to delete this

pkg/operator/console.go
```diff
@@ -19,6 +19,9 @@ func defaultConsole() *v1alpha1.Console {
 			Name:      OpenshiftConsoleNamespace,
 			Namespace: OpenshiftConsoleNamespace,
 		},
+		Spec: v1alpha1.ConsoleSpec{
+			ManagementState: "Managed",
```
enj: use the defined const

pkg/operator/oauth.go
```diff
@@ -62,6 +64,29 @@ func ApplyOAuth(cr *v1alpha1.Console, rt *routev1.Route) (*oauthv1.OAuthClient,
...
+	return sdk.Delete(authConfigSecret)
+}
+
+// deletes secret & eliminates redirectUris
+func NeutralizeOAuthClient(cr *v1alpha1.Console) error {
+	authClient := defaultOAuthClient(cr)
+	if err := sdk.Get(authClient); err != nil {
+		return err
+	}
+	nilSecret := ""
```
enj: I would just leave the secret set since it is better than having a public OAuth client.  You could set it to a random value if you wanted to be extra paranoid.

benjaminapetersen: Random value is better than empty string?

pkg/operator/reconcile.go
benjaminapetersen: derp. resolved.

PR #60, opened 2018-10-23
Update READMEs cr.yaml path

jhadvig:
Not really sure if it we should change the folder name or the README.md so starting from the end :rocket:

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ Finally, you can create an instance of your custom resource (CR).
 Once an instance exists, your operator will take over managing it.
 
 ```bash
-$ oc create -f deploy/cr.yaml
+$ oc create -f examples/cr.yaml
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,7 +54,7 @@ Finally, you can create an instance of your custom resource (CR).
 Once an instance exists, your operator will take over managing it.
 
 ```bash
-$ oc create -f deploy/cr.yaml
+$ oc create -f examples/cr.yaml
```
spadgett: I don't think this is needed at all anymore. @benjaminapetersen can you confirm?

jhadvig: No we dont... the instance is created automatically

jhadvig:
Closing since this will be fixed in https://github.com/openshift/console-operator/pull/48

PR #65, opened 2018-11-02
Add annotation to PodSpec on Deployment whenever ConfigMap changes

benjaminapetersen:
Add annotation to PodSpec on Deployment whenever ConfigMap changes ResourceVersion changes to ensure Pod is redeployed

### DIFF
--- a/pkg/operator/deployment.go
+++ b/pkg/operator/deployment.go
@@ -1,20 +1,25 @@
 package operator
 
 import (
+	// standard lib
 	"fmt"
-
+	// 3rd party
 	"github.com/sirupsen/logrus"
-
+	// kubernetes
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
-
+	// openshift
 	"github.com/operator-framework/operator-sdk/pkg/sdk"
-
+	// operator
 	"github.com/openshift/console-operator/pkg/apis/console/v1alpha1"
+)
+
+const (
+	configMapResourceVersionAnnotation = "console.openshift.io/configmapversion"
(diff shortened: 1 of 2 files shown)


pkg/operator/deployment.go
```diff
@@ -172,6 +180,26 @@ func consoleContainer(cr *v1alpha1.Console) corev1.Container {
 
 }
 
+func UpdateDeployment(cr *v1alpha1.Console, cm *corev1.ConfigMap) (*appsv1.Deployment, error) {
+	d := newConsoleDeployment(cr)
+	if err := sdk.Get(d); err != nil {
+		return nil, err
+	}
+
+	if len(d.Spec.Template.ObjectMeta.Annotations) == 0 {
```
enj: this should be nil check

pkg/operator/deployment.go
```diff
@@ -172,6 +180,26 @@ func consoleContainer(cr *v1alpha1.Console) corev1.Container {
...
+		d.Spec.Template.ObjectMeta.Annotations = map[string]string{}
+	}
+
+	// If the configmap is updated, we want to rollout a new pod.
+	// setting an annotation is sufficient for this task.
+	d.Spec.Template.ObjectMeta.Annotations[configMapResourceVersionAnnotation] = cm.ObjectMeta.GetResourceVersion()
+
+	if err := sdk.Update(d); err != nil {
```
enj: Only update if changed

pkg/operator/deployment.go
benjaminapetersen: will do.

pkg/operator/deployment.go
benjaminapetersen: can do.

pkg/operator/deployment.go
```diff
@@ -172,6 +180,35 @@ func consoleContainer(cr *v1alpha1.Console) corev1.Container {
 
 }
 
+func UpdateDeployment(cr *v1alpha1.Console, cm *corev1.ConfigMap) (*appsv1.Deployment, error) {
+	d := newConsoleDeployment(cr)
+	changed := false
```
benjaminapetersen: @enj tracking changed here.

PR #68, opened 2018-11-19
Status is not namespaced

benjaminapetersen:


benjaminapetersen:
@spadgett @enj tests pass

PR #73, opened 2018-11-28
Add block to README for local dev with correct operator token

benjaminapetersen:
Add block to README for local dev with correct operator service account token


benjaminapetersen:
/assign @enj @spadgett 

No code in this one, just some dev notes updates.

PR #75, opened 2018-11-28
Use correct informer for OAuth clients

enj:
Previously, the code incorrectly called the OAuth client authorizations informer.

Update the informer and client types to be more consistent.

Signed-off-by: Monis Khan <mkhan@redhat.com>

/assign @benjaminapetersen

PR #76, opened 2018-11-28
Rename pkg/console/subresource -> pkg/console/resource.

benjaminapetersen:
Subresource has specific meaning in `kube`.


enj:
/hold

Will conflict with #75

benjaminapetersen:
Yup, this one can go after, I'll rebase.

PR #79, opened 2018-12-04
Sync loop update

benjaminapetersen:
- Cleans up the `sync_v400` fn
   - breaks this function down into multiple smaller functions
- Fixes issue where `oauthclient` `redirectURI` might not be set correctly

/assign @enj

### DIFF
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -1,24 +1,28 @@
 package operator
 
 import (
+	"errors"
+	"fmt"
+
+	"github.com/openshift/console-operator/pkg/controller"
+
 	// 3rd party
 	"github.com/sirupsen/logrus"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	coreclientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	// kube
 	oauthv1 "github.com/openshift/api/oauth/v1"
 	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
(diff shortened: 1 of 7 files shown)


benjaminapetersen:
@spadgett fyi

pkg/console/operator/sync_v400.go
```diff
@@ -29,93 +33,63 @@ import (
...
-		logrus.Errorf("%q: %v \n", "deployment", depErr)
-		allErrors = append(allErrors, depErr)
+		return consoleConfig, depErr
 	}
 	toUpdate = toUpdate || depChanged
 
+	// TODO: when this is done, be sure that the
+	// returned errors are handled properly in OPERATOR.GO
+
 	// handy debugging block
```
benjaminapetersen: This will go away when I remove WIP

benjaminapetersen:
@spadgett @enj take a look if you get a chance.
The sync loop is still pretty noisy, but it comes down to this:

```bash
INFO[0005] sync_v400 complete. changes:
INFO[0005] service changed: false
INFO[0005] route is new: false
INFO[0005] configMap changed: false
INFO[0005] secret changed: false
INFO[0005] oauth changed: false
INFO[0005] deployment changed: true
```
I need to figure out why the deployment always things it changed. That said, its not rolling out new pods:
```bash
NAME                                 READY     STATUS             RESTARTS   AGE
openshift-console-67b8f48b9d-kpqvj   1/1       Running            0          26m
```

benjaminapetersen:
Sync loop runs once every 2.5 seconds.  much less frequent, but still a bit chatty.  Will look into this some more as a follow.

PR #81, opened 2018-12-05
Update Deployment sync to check configmap and secret resource versions

benjaminapetersen:
Updates the `SyncDeployment()` fn to take the `secret` and `configmap`, and test the resource versions against what is recorded on the pod template.  this will be used to force a pod rollout instead of a boolean and should be more reliable (for example, if one of the dependencies changed on a previous sync, but the sync died). 

@enj @spadgett


enj:
/lgtm

Getting closer.

Issue #83, opened 2018-12-05
console should have multiple replicas ootb


jwforres:
right now the console deployment is only creating a single console replica, historically we always went with the same replicas as the number of masters by default, so 3, bare minimum would be 2, we must be HA by default

benjaminapetersen:
Fixed.

PR #86, opened 2018-12-06
Ensure minimum 2 replicas of console

benjaminapetersen:
Fixes issue #83 

@jwforres @spadgett

### DIFF
--- a/examples/cr.yaml
+++ b/examples/cr.yaml
@@ -15,12 +15,9 @@ spec:
   # - Unmanaged indicates that the operator should idle and ignore the console and its sub-resources
   # - Removed indicates that the operator should delete all of the sub-resources of the console
   managementState: Managed
-  # there should be only one console at this time
-  # this could be used to support leader election if desired
-  count: 1
+  # there should be the same number of console replicas as the number of masters, minimum 2
+  count: 3
   logging:
     level: 0
   # artifact of base operator
   version: 4.0.0
-  # going away
-  value: "panda-bears-are-awesome-and-this-should-be-a-console-instead-woot"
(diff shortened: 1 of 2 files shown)


pkg/console/operator/operator.go
```diff
@@ -274,8 +274,8 @@ func (c *ConsoleOperator) defaultConsole() *consolev1alpha1.Console {
 				// if Verison is not 4.0.0 our reconcile loop will not pick it up
 				Version: "4.0.0",
 			},
-			// one replica is created
-			Count: 1,
+			// minimum of 2 replicas
+			Count: 2,
```
enj: Uh why is this being changed in two places?  And why is `2` instead of `3`?

examples/cr.yaml
```diff
@@ -15,12 +15,9 @@ spec:
   # - Unmanaged indicates that the operator should idle and ignore the console and its sub-resources
   # - Removed indicates that the operator should delete all of the sub-resources of the console
   managementState: Managed
-  # there should be only one console at this time
-  # this could be used to support leader election if desired
-  count: 1
+  # there should be the same number of console replicas as the number of masters, minimum 2
```
wking: Looks like this (and the PR title and commit message) missed [the bump from 2 to 3][1].

[1]: https://github.com/openshift/console-operator/pull/86#discussion_r239562790

benjaminapetersen: Thanks @wking!  Sorry about that, rushed it in for demo sake, should have stopped & updated the extra bits. :/

I'll follow a PR to fix the example file at least.

benjaminapetersen: https://github.com/openshift/console-operator/pull/101

PR #89, opened 2018-12-10
Update Console type to include custom routes, logoutPublicURL, etc

benjaminapetersen:
Additional fields needed for the console CR.
Once this is settled/merged, I'll make a PR to `openshift/api` to move this up, then a PR to remove it from the `console-operator` & use the upstream version.


benjaminapetersen:
/hold
Just to be cautious about demos :)

benjaminapetersen:
@spadgett if you know of any other updates needed.

benjaminapetersen:
@jwforres fyi

examples/cr.yaml
```diff
@@ -21,3 +21,9 @@ spec:
     level: 0
   # artifact of base operator
   version: 4.0.0
+  defaultRoute: true
```
benjaminapetersen: This is just keeping the example file current

pkg/apis/console/v1alpha1/types.go
```diff
@@ -19,27 +19,48 @@ type Console struct {
...
-	// https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/apis/imageregistry/v1alpha1/types.go#L91-L92
-	// DefaultRoute: T|F
-	// additional routes config w/secrets
-	// Route[]
+	// Replicas is the number of Console replicas
+	Replicas int32 `json:"count,omitempty"`
+	// default generated route
+	DefaultRoute bool `json:"defaultRoute,omitempty"`
+	// there may potentially be musltiple custom host names
+	Routes []ConsoleConfigRoute `json:"routes,omitempty"`
```
benjaminapetersen: Per discussion, we are going to eliminate `Routes`, and just support a single route.  Then it will simply be:

```
Spec
  PublicHostName: string
  BasePath: string
  LogoutPublicURL: string
```
The `Secret` for the TLS certificate will be a well-known named secret in `openshift-managed` at this point.  It should be named something like `console-route-secret` (need to decide the name).

I'll make the changes on this.
@spadgett

spadgett: I think this is reasonable and will let us start handling these values. We can reevaluate the exact API when the move this to a global config. We might want to eventually put items into separate stanzas, but that doesn't seem too difficult a change.

I'd make it `PublicHostname` (hostname is one word). It's also consistent with other names in openshift/api: https://github.com/openshift/api/search?utf8=%E2%9C%93&q=hostname&type=

benjaminapetersen: Sounds good.

spadgett:
Hey, @benjaminapetersen. I'm going to close since this looks outdated. Feel free to update / reopen if needed.

/close

PR #91, opened 2018-12-12
Improve sync loop logging feedback

benjaminapetersen:
A block from the sync loop will now look like this:

screen shot 2018-12-12 at 10 35 58 am

This PR only adds a few log statements & cleans up formatting (`\n`, consistent lower case, verbiage, etc).

### DIFF
--- a/pkg/console/operator/operator.go
+++ b/pkg/console/operator/operator.go
@@ -3,6 +3,7 @@ package operator
 import (
 	// standard lib
 	"fmt"
+	"time"
 
 	// 3rd party
 	"github.com/blang/semver"
@@ -152,6 +153,8 @@ type ConsoleOperator struct {
 }
 
 func (c *ConsoleOperator) Run(stopCh <-chan struct{}) {
+	logrus.Info("starting console operator")
+	defer logrus.Info("shutting down console operator")
 	// only start one worker because we only have one key name in our queue
 	// since this operator works on a singleton, it does not make sense to ever run more than one worker
 	c.controller.Run(1, stopCh)
@@ -160,7 +163,12 @@ func (c *ConsoleOperator) Run(stopCh <-chan struct{}) {
 // sync() is the handler() function equivalent from the sdk
 // this is the big switch statement.
 // TODO: clean this up a bit, its messy
-func (c *ConsoleOperator) sync(_ interface{}) error {
+func (c *ConsoleOperator) sync(key interface{}) error {
+	startTime := time.Now()
+	logrus.Infof("started syncing operator %q (%v)", key, startTime)
+	defer func() {
+		logrus.Infof("finished syncing operator %q (%v) \n\n", key, time.Since(startTime))
+	}()
 	// we ignore the passed in key because it will always be workQueueKey
 	// it does not matter how the sync loop was triggered
 	// all we need to worry about is reconciling the state back to what we expect
@@ -176,14 +184,14 @@ func (c *ConsoleOperator) sync(_ interface{}) error {
 
 	switch operatorConfig.Spec.ManagementState {
 	case operatorsv1alpha1.Managed:
-		fmt.Println("Console is in a managed state.")
+		logrus.Println("console is in a managed state.")
 		// handled below
 	case operatorsv1alpha1.Unmanaged:
-		fmt.Println("Console is in an unmanaged state.")
+		logrus.Println("console is in an unmanaged state.")
 		return nil
 	// take a look @ https://github.com/openshift/service-serving-cert-signer/blob/master/pkg/operator/operator.go#L86
 	case operatorsv1alpha1.Removed:
-		fmt.Println("Console has been removed.")
+		logrus.Println("console has been removed.")
 		return c.deleteAllResources(operatorConfig)
 	default:
 		// TODO should update status
@@ -218,7 +226,6 @@ func (c *ConsoleOperator) sync(_ interface{}) error {
 	switch {
 	// v4.0.0 or nil
 	case v311_to_401.BetweenOrEmpty(currentActualVersion):
-		logrus.Println("Sync-4.0.0")
 		outConfig, configChanged, err = sync_v400(c, outConfig)
 		errs = append(errs, err)
 		if err == nil {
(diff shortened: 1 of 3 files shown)


benjaminapetersen:
@jhadvig

spadgett:
> consistent lower case

It looks like there's still a mix of lowercase / uppercase messages. Is that intentional?

pkg/console/operator/operator.go
```diff
@@ -228,7 +235,7 @@ func (c *ConsoleOperator) sync(_ interface{}) error {
 			}
 		}
 	default:
-		logrus.Printf("Unrecognized version. Desired %s, Actual %s", desiredVersion, currentActualVersion)
+		logrus.Printf("unrecognized version. desired %s, actual %s", desiredVersion, currentActualVersion)
```
jhadvig: I understand why we lower-casing the first sentence's first work, but not sure if that should apply to all sentences.

pkg/console/operator/sync_v400.go
```diff
@@ -39,6 +39,8 @@ import (
 // This ensures the logic is simpler as we do not have to handle coordination between objects within
 // the loop.
 func sync_v400(co *ConsoleOperator, consoleConfig *v1alpha1.Console) (*v1alpha1.Console, bool, error) {
+	logrus.Println("Running sync loop 4.0.0")
```
jhadvig: lower-case first letter ?

benjaminapetersen:
Nope, got sidetracked before finishing that last block.  Updated!

pkg/console/operator/sync_v400.go
```diff
@@ -215,9 +228,10 @@ func SyncRoute(co *ConsoleOperator, consoleConfig *v1alpha1.Console) (*routev1.R
 	if len(rt.Spec.Host) == 0 {
 		// TODO STATUS
 		logrus.Errorf("%q: %v \n", "route", rtErr)
-		return nil, false, errors.New("Waiting on Route.Spec.Host")
+		return nil, false, errors.New("waiting on route.spec.host")
 	}
 	// only returns the route if we hit the happy path, we cannot make progress w/o the host
+	logrus.Println("route exists and is in the correct state")
```
jhadvig: Not really sure what "correct state" means. Should the state be printed ?

pkg/console/subresource/util/util.go
```diff
@@ -2,14 +2,15 @@ package util
...
+	"os"
+	"strings"
+
 	"github.com/openshift/console-operator/pkg/apis/console/v1alpha1"
 	"github.com/openshift/console-operator/pkg/controller"
 	"github.com/sirupsen/logrus"
-	"gopkg.in/yaml.v2"
-	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	yaml "gopkg.in/yaml.v2"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
jhadvig: I think this will be imported as `v1` anyway

pkg/console/operator/sync_v400.go
```diff
@@ -78,29 +80,31 @@ func sync_v400(co *ConsoleOperator, consoleConfig *v1alpha1.Console) (*v1alpha1.
...
-		logrus.Infof("Sync_v400: To update Spec: %v", toUpdate)
+		logrus.Infof("sync_v400: to update spec: %v", toUpdate)
 		// TODO: set the status.
 		// setStatus(consoleConfig.Status, svc, rt, cm, dep, oa, sec)
 	}
 
+	defer func() {
+		logrus.Printf("sync loop 4.0.0 complete:")
+		logrus.Printf("service changed: %v", svcChanged)
+		logrus.Printf("route is new: %v", rtChanged)
```
spadgett: Should this be "route changed"?

pkg/console/operator/sync_v400.go
```diff
@@ -78,29 +80,31 @@ func sync_v400(co *ConsoleOperator, consoleConfig *v1alpha1.Console) (*v1alpha1.
...
 	if toUpdate {
-		logrus.Infof("Sync_v400: To update Spec: %v", toUpdate)
+		logrus.Infof("sync_v400: to update spec: %v", toUpdate)
 		// TODO: set the status.
 		// setStatus(consoleConfig.Status, svc, rt, cm, dep, oa, sec)
 	}
 
+	defer func() {
+		logrus.Printf("sync loop 4.0.0 complete:")
+		logrus.Printf("service changed: %v", svcChanged)
```
jhadvig: Since those are listed items, maybe add tab:
```go
logrus.Printf("\t service changed: %v", svcChanged)
```

maybe even without the space

pkg/console/operator/sync_v400.go
benjaminapetersen: Yeah, thinking about that. Route has been unique as I haven't done an 'update', just a 'create' or 'get' -- we just wait for `host` to be applied. So its a bit intentional, but it may be better to just say "changed" anyway.  I'll update that.

benjaminapetersen: updated

benjaminapetersen:
And added `\n\n` at the end of each sync loop, just to improve readability.

screen shot 2018-12-12 at 3 45 32 pm

pkg/console/operator/sync_v400.go
benjaminapetersen: Prefer the `tab`?

screen shot 2018-12-12 at 4 11 57 pm

benjaminapetersen: Does it look odd with the `INFO[123]` far left?

jhadvig: yeah, not really sure. Its definitely more clear that its an item fro ma list :)

benjaminapetersen: Works for me.

screen shot 2018-12-13 at 10 42 50 am

benjaminapetersen:
Last for this PR:
screen shot 2018-12-13 at 10 46 15 am

benjaminapetersen:
No flakes. Woo! I'll charge double my current rate for changes at this point 😂

PR #92, opened 2018-12-13
Add Validate() fn to route

benjaminapetersen:
The sync loop should also ensure the correctness of each resource, at least in all of the fields that matter for a running console. This adds a `validate()` fn to the route `sync`, which will only trigger an `update` if `changed` is `true`.

### DIFF
--- a/pkg/console/subresource/route/route.go
+++ b/pkg/console/subresource/route/route.go
@@ -1,15 +1,18 @@
 package route
 
 import (
+
 	// kube
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+
 	// openshift
 	routev1 "github.com/openshift/api/route/v1"
 	routeclient "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
 	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
+
 	// operator
 	"github.com/openshift/console-operator/pkg/apis/console/v1alpha1"
 	"github.com/openshift/console-operator/pkg/console/subresource/util"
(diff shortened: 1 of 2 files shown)


benjaminapetersen:
@jhadvig fyi

benjaminapetersen:
This will be the first of several PRs, one for each, if needed (they may not all need an improved Validate, the current `Apply` does handle some of the resources quite well.

pkg/console/subresource/route/route.go
```diff
@@ -109,3 +100,72 @@ func Stub() *routev1.Route {
 		ObjectMeta: meta,
 	}
 }
+
+// we can't just blindly apply the route, we need the route.Spec.Host
+// and we don't want to trigger a sync loop.
+func Validate(route *routev1.Route) (*routev1.Route, bool) {
+	changed := false
+
+	labelsSame := equality.Semantic.DeepEqual(route.GetLabels(), util.SharedLabels())
```
benjaminapetersen: This is a tad repetitive, but its clear.

benjaminapetersen:
Error message looks like this:

screen shot 2018-12-13 at 10 33 36 am

Note there is a duplicate message.  This is annoying, because `logrus.Errorf` doesn't return a value, but DOES do the colored `ERRO[123]`.  Therefore I need to print twice.  I prefer this to not using the color output at all, it is much easier to find errors when scanning the logs.

benjaminapetersen:
Note also the second message has the `sync-queue failed with:` prefix.  Not everything kills the sync loop... I don't know how important that would be for debugging, but it can't hurt.

spadgett:
I don't know the semantics of `resourceapply.ApplyRoute`. If we leave the hostname out of the spec, does it do the right thing (i.e., leave the generated value alone)? If so, we could follow the same pattern for routes as other resources.

With deep comparisons, do we need to worry if new fields added to the route spec in future versions will cause the operator to thrash?

benjaminapetersen:
At this point there isn't a `resourceapply.ApplyRoute`, so I'm just doing a `routeClient.Routes(controller.TargetNamespace).Update(validatedRoute)`.  Since the `Validate()` fn is not concerned about `Host`, it is doing the right thing (`if len(rt.Spec.Host)` is tested prior to this step).

I believe `resourceapply.ApplyRoute` doesn't exist because it hasn't been needed yet, so I may end up writing it at some point & opening against `library-go`. 

If new fields are applied by the server in the future and we don't know about them, I believe we would get thrashing.  I'm not sure what the best solution to this is yet, other than comparing values one by one. That seems a bit tedious but worth discussing.

PR #93, opened 2018-12-14
Service/validation

benjaminapetersen:
Continuing with the existing pattern, an invalid state on the `service` will trigger an `update` and throw an error, killing the sync loop:

screen shot 2018-12-14 at 10 29 00 am

This change to the service will trigger a new sync loop which will start at the top and work its way through the resources (one thing at a time, if the sync loop reaches the end, this indicates that everything is in the correct state).

### DIFF
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -200,13 +200,31 @@ func SyncConfigMap(co *ConsoleOperator, consoleConfig *v1alpha1.Console, rt *rou
 // there is nothing special about our service, so no additional error handling is needed here.
 func SyncService(co *ConsoleOperator, consoleConfig *v1alpha1.Console) (*corev1.Service, bool, error) {
 	logrus.Printf("validating console service...")
-	svc, svcChanged, svcErr := resourceapply.ApplyService(co.serviceClient, servicesub.DefaultService(consoleConfig))
-	if svcErr != nil {
-		logrus.Errorf("%q: %v \n", "service", svcErr)
-		return nil, false, svcErr
+	svc, getErr := co.serviceClient.Services(controller.TargetNamespace).Get(servicesub.Stub().Name, metav1.GetOptions{})
+	if apierrors.IsNotFound(getErr) {
+		_, svcChanged, svcErr := resourceapply.ApplyService(co.serviceClient, servicesub.DefaultService(consoleConfig))
+		return nil, svcChanged, fmt.Errorf("service not found, creating new service, create err = %v", svcErr)
+	}
+	if getErr != nil {
+		logrus.Errorf("%q: %v \n", "service", getErr)
+		return nil, false, getErr
+	}
+
+	if validatedService, changed := servicesub.Validate(svc); changed {
+		// ideally we would use resourceapply.ApplyService, however the current iteration
+		// only handles selector, type & EnsureObjectMeta.
+		// we need ports & sessionAffinity as well.
+		// TODO: PR upstream to ApplyService() to fix the above, so we can ajust here.
+		if _, err := co.serviceClient.Services(controller.TargetNamespace).Update(validatedService); err != nil {
+			logrus.Errorf("%q: %v \n", "service", err)
+			return nil, false, err
+		}
+		errMsg := fmt.Errorf("service is invalid, correcting service state")
+		logrus.Error(errMsg)
+		return nil, changed, errMsg
 	}
 	logrus.Println("service exists and is in the correct state")
-	return svc, svcChanged, svcErr
+	return svc, false, getErr
 }
 
 // apply route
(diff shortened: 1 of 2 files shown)


pkg/console/operator/sync_v400.go
```diff
@@ -200,13 +200,27 @@ func SyncConfigMap(co *ConsoleOperator, consoleConfig *v1alpha1.Console, rt *rou
...
+		_, svcChanged, svcErr := resourceapply.ApplyService(co.serviceClient, servicesub.DefaultService(consoleConfig))
+		return nil, svcChanged, fmt.Errorf("service not found, creating new service, create err = %v", svcErr)
+	}
+	if getErr != nil {
+		logrus.Errorf("%q: %v \n", "service", getErr)
+		return nil, false, getErr
+	}
+
+	if validatedService, changed := servicesub.Validate(svc); changed {
+		if _, err := co.serviceClient.Services(controller.TargetNamespace).Update(validatedService); err != nil {
```
benjaminapetersen: We cannot use `resourceapply.ApplyService()` here as it does not thoroughly check the `Service.Spec`.  If used, it will simply return the `existing` service & never actually issue an `Update` (our loop will `sync` forever). 

`resourceapply.ApplyService()` does work as a model for `SyncService()`, however.

pkg/console/operator/sync_v400.go
```diff
@@ -200,13 +200,27 @@ func SyncConfigMap(co *ConsoleOperator, consoleConfig *v1alpha1.Console, rt *rou
...
+			logrus.Errorf("%q: %v \n", "service", err)
+			return nil, false, err
+		}
+		errMsg := fmt.Errorf("service is invalid, correcting service state")
```
benjaminapetersen: As with route, kill the sync loop here at the update.

benjaminapetersen:
@jhadvig fyi

pkg/console/operator/sync_v400.go
spadgett: Can you expand on what specifically it doesn't check that we need?

pkg/console/subresource/service/service.go
```diff
@@ -18,42 +20,78 @@ const (
...
+
+func Validate(svc *v1.Service) (*v1.Service, bool) {
+	changed := false
+
+	if svc.ObjectMeta.Annotations[ServingCertSecretAnnotation] != ConsoleServingCertName {
+		changed = true
+		svc.ObjectMeta.Annotations[ServingCertSecretAnnotation] = ConsoleServingCertName
+	}
+
+	if !equality.Semantic.DeepEqual(svc.Spec.Selector, util.LabelsForConsole()) {
```
spadgett: Do we care about labels?

pkg/console/subresource/service/service.go
```diff
@@ -18,42 +20,78 @@ const (
 	consolePortName        = "https"
 	consolePort            = 443
 	consoleTargetPort      = 8443
+	sessionAffinity        = "None"
+	publishingType         = "ClusterIP"
```
spadgett: serviceType ?

pkg/console/subresource/service/service.go
benjaminapetersen: no prob!

pkg/console/operator/sync_v400.go
benjaminapetersen: It checks if `selector` matches, `type` matches, and does `EnsureObjectMeta` (which sets `namespace`, `name` & does a `merge` of `labels` and `annotations`).  

So it does not look at `ports`, `type`, or `sessionAffinity`. 

This could potentially be an upstream PR to resolve if we want to do that.  I can add a comment here.

benjaminapetersen: Yup.  Shall we keep this for now while I work on the upstream?  I can swap back to `ApplyService` when it merges & I update dependencies.

benjaminapetersen: Made a comment to doc this. I'd prefer to get the operator functionality correct now, and then move the logic around, I think.

benjaminapetersen:
/retest
heh, the empty test failed.

benjaminapetersen:
letting this one linger, we probably could have kept it, but its not on the "approved" list (though opened & tests passed quite a while ago).

pkg/console/subresource/service/service.go
benjaminapetersen: We may want this now.

pkg/console/operator/sync_v400.go
benjaminapetersen: Thinking we may want this validation now?

benjaminapetersen:
Nevermind, this needs a rebase, not a priority atm.

spadgett:
Closing since this PR is out of date. If there are gaps in library-go `ApplyService`, we should try to address them there instead of rolling our own.

/close

PR #94, opened 2018-12-14
ConfigMap/validation

benjaminapetersen:
Looking at dropping the config string into the logging output:

screen shot 2018-12-14 at 1 27 38 pm

this is as currently as `yaml`, rather than `json`.


benjaminapetersen:
@spadgett @jhadvig so this is interesting, I can tamper with the configmap in a bunch of ways and the sync loop handles it as it is.  We don't really need any special validation on this.

spadgett:
I'd skip the extra logic if it's not needed.

benjaminapetersen:
Yeah I'm inclined to just log the `configmap` (for debugging) and call it good on this one.

benjaminapetersen:
closing since there is no extra validation needed.

PR #95, opened 2018-12-14
Improve log message for resource version annotation comparison

benjaminapetersen:
Attempting to make the logging of the resource version comparison clear to someone who is not intimately familiar with the inner workings of the operator:

screen shot 2018-12-14 at 1 35 54 pm

Knowing that these annotations are for the `spec.template.metadata.annotations`, for the `pod`, is helpful in the log output, should the information be needed to diagnost problems.


benjaminapetersen:
@jhadvig 

/assign @spadgett @enj

spadgett:
These functions are removed in #123

/close

PR #96, opened 2018-12-17
Inject service-ca.crt into the console pod from a config map

spadgett:
Annotate a config map to hold the service-ca.crt for the Prometheus proxy.

Requires https://github.com/openshift/console/pull/960

See https://github.com/openshift/service-serving-cert-signer#openshift-service-serving-cert-signer-operator

/cc @benjaminapetersen @enj

### DIFF
--- a/pkg/console/subresource/configmap/service_ca.go
+++ b/pkg/console/subresource/configmap/service_ca.go
@@ -1,0 +1,36 @@
+package configmap
+
+import (
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/openshift/console-operator/pkg/apis/console/v1alpha1"
+	"github.com/openshift/console-operator/pkg/console/subresource/util"
+)
+
+const (
+	// ServiceCAConfigMapName is the name of the config map that contains the service CA bundle.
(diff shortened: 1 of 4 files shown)


spadgett:
@brancz @kyoto fyi

benjaminapetersen:
So far so good.  Agree, some of the funcs are long for what they do, but its doing the right thing...

spadgett:
OK, metrics are now working with this change:

console-openshift-console apps spadgett devcluster openshift com_overview_ns_openshift-console

pkg/console/subresource/configmap/service_ca.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/openshift/console-operator/pkg/apis/console/v1alpha1"
+	"github.com/openshift/console-operator/pkg/console/subresource/util"
+)
+
+const (
+	// ServiceCAConfigMapName is the name of the config map that contains the service CA bundle.
+	ServiceCAConfigMapName = "service-ca"
```
enj: You need to add this name to the list of watched config maps.

PR #74, opened 2018-11-28
[WIP] Update cluster operator status 

benjaminapetersen:
Uncomments the `clusteroperatorstatus` blocks.

Note that `NewClusterOperatorStatusController` from `controller.go` in `library-go` is the old group `operatorstatus.openshift.io`:

```
clusterOperatorClient:    clusterOperatorClient.Resource(schema.GroupVersionResource{Group: "operatorstatus.openshift.io", Version: "v1", Resource: "clusteroperators"}).Namespace(namespace),
```
working on updating `library-go` to get this changed (there is an unrelated conflict in another bit of code).

PR #97, opened 2018-12-18
Version bump

jhadvig:
@benjaminapetersen here is the version bump.
PTAL


spadgett:
@jhadvig can you separate the vendor changes into a separate commit to make review easier?

benjaminapetersen:
@jhadvig agree, I can't load in browser, too many files changed, github falls over.

benjaminapetersen:
@jhadvig but initially, looks like a downgrade?   `kubernetes-1.11.1` -> `kubernetes-1.11.0`?  I think that is the wrong direction?

jhadvig:
@spadgett moved the bump to a separate commit, although it's not strictly speaking a bump since we needed to get the updated version of **library-go** that already contains the events.
@benjaminapetersen regarding the confusion `1.11.1 -> `1.11.0`. Was advised to use `1.11.0` version since master team is using that one for their operators. But it shouldn't matter.

PTAL

pkg/console/starter/starter.go
```diff
@@ -107,6 +110,9 @@ func RunOperator(clientConfig *rest.Config, stopCh <-chan struct{}) error {
 		oauthinformers.WithTweakListOptions(tweakOAuthListOptions),
 	)
 
+	// TODO: Replace this with real event recorder (use ControllerContext).
+	recorder := ctx.EventRecorder
```
benjaminapetersen: I need to look into this.  Does this swallow events (not a real recorder)?

benjaminapetersen:
/retest

Generally looks good to me, just checking over a bit more.  Nice job on that!

benjaminapetersen:
/retest

```
could not wait for pod to complete: could not wait for pod completion: the pod ci-op-fgtgii7k/e2e-aws failed after 29m34s (failed containers: setup, test):  unknown

Container setup exited with code 1, reason Error
Container test exited with code 1, reason Error
```

benjaminapetersen:
Using deprecated install config? Weird.

/retest

benjaminapetersen:
This will replace much of what was done in PR #74.

jhadvig:
@benjaminapetersen had to add extra rules the console-operator Role and an extra RoleBinding to be able to get configMap from the `kube-system` namespace.

jhadvig:
@spadgett @benjaminapetersen the tests finally passed :pray:

pkg/console/operator/operator.go
```diff
@@ -183,7 +189,7 @@ func (c *consoleOperator) Sync(obj metav1.Object) error {
 	switch {
 	// v4.0.0 or nil
 	case v311_to_401.BetweenOrEmpty(currentActualVersion):
-		outConfig, configChanged, err = sync_v400(c, outConfig)
+		outConfig, configChanged, err = sync_v400(c, c.recorder, outConfig)
```
benjaminapetersen: NIT: So since we added it to the `consoleOperator{ }` struct above, it may not need to be separated out as a separate argument here?  Ie could just be:

```go
// operator.go
// its already part of 'c'
sync_v400(c, outConfig)

// sync_v400.go
func sync_v400(co *consoleOperator, consoleConfig *v1alpha1.Console) {
   // so here we can pop it off to use as you are
   recorder := c.recorder
}
```

pkg/console/starter/starter.go
benjaminapetersen: Maybe add a bit more context to the comment here, quick explain how it differs from a "real" event recorder.

manifests/05-operator.yaml
```diff
@@ -41,6 +41,10 @@ spec:
           value: docker.io/openshift/origin-console:latest
         - name: OPERATOR_NAME
           value: "console-operator"
+        - name: POD_NAME
```
benjaminapetersen: Do you happen to remember why this was needed?  We talked about it on Bluejeans, I just forget atm.

benjaminapetersen: I believe you mentioned it does affect CI.

manifests/02-rbac-role.yaml
```diff
@@ -20,6 +20,7 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - pods
```
benjaminapetersen: So we don't directly manipulate `pods`, why do we need to add this?

manifests/02-rbac-role.yaml
```diff
@@ -35,6 +36,7 @@ rules:
   - apps
   resources:
   - deployments
+  - replicasets
```
benjaminapetersen: We also don't directly manage `replicasets`.

manifests/03-rbac-rolebinding.yaml
```diff
@@ -11,6 +11,20 @@ subjects:
 - kind: ServiceAccount
   name: console-operator
 ---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
```
benjaminapetersen: Pondering this as well, It's not immediately clear why the addition.  Quick comment to clarify? Thanks!

manifests/05-operator.yaml
jhadvig: It's because of the Recorder. The envar itself is explained [here](https://github.com/openshift/library-go/blob/af91f0cfd1aad763ce48455b4d34e927dc782f04/pkg/operator/events/recorder.go#L24-L26)

Because of that we needed to add the `pods` && `replicaset` to the rbac rule, so in order to be able to record events, the console-operator pod (or better to say it's service account) needs to be able to get that specified pod and also replicaset.

manifests/03-rbac-rolebinding.yaml
jhadvig: We need to be able to read existing `configmap/extension-apiserver-authentication` in *kube-system* namespace. 
*extension-apiserver-authentication* is the name of ConfigMap in the kube-system namespace holding the root certificate bundle to use to verify client certificates on incoming requests before trusting usernames in headers specified by --requestheader-username-headers. This is created in the cluster by the kube-apiserver.
Explained [here](https://github.com/deads2k/kubernetes/blob/1105e4e0d1c0e78e2a203a136e9f8bcaff5c36ab/staging/src/k8s.io/apiserver/pkg/server/options/authentication.go#L235-L239).

manifests/02-rbac-role.yaml
jhadvig: Explained in comment above.

manifests/02-rbac-role.yaml
jhadvig: Explained in comment above.

jhadvig:
@benjaminapetersen fixed the NIT and added comments to explain the changes, sorry for not adding them as part of description.
Hope that makes things clear :)

manifests/03-rbac-rolebinding.yaml
enj: Turn off the delegated authn/authz (it is useless to you) and drop this RBAC change.

manifests/05-operator.yaml
enj: That would mean you need read only access to those resources, not read/write.

pkg/console/starter/starter.go
enj: This one looks like the real thing.

benjaminapetersen: twas my thinking as well :)

manifests/05-operator.yaml
benjaminapetersen: Great, sounds good.

jhadvig:
@enj comments addressed.

@benjaminapetersen PTAL

benjaminapetersen:
Great, pulling it down to test.

benjaminapetersen:
/lgtm
though need the e2e tests to pass.... ☹️

PR #99, opened 2018-12-19
[WIP] Manifests/env/pod name and operator image

benjaminapetersen:
@jhadvig fyi 

- Add POD_NAME and OPERATOR_IMAGE env vars, 
- update image-references to :v4.0 instead of :latest

WIP on this until we are certain about the correct tag to use.


benjaminapetersen:
Huh. Same fails.

/retest

benjaminapetersen:
Going to close this, but open just the `env` var changes as a single item.  Per convo with @spadgett, since the `images` are overwritten when we ship, this is probably unnecessary to update.

PR #101, opened 2018-12-21
Update incorrect comment in example/cr.yaml 

benjaminapetersen:
Update incorrect comment in example/cr.yaml regarding minimum number of console replicas

### DIFF
--- a/examples/cr.yaml
+++ b/examples/cr.yaml
@@ -15,7 +15,8 @@ spec:
   # - Unmanaged indicates that the operator should idle and ignore the console and its sub-resources
   # - Removed indicates that the operator should delete all of the sub-resources of the console
   managementState: Managed
-  # there should be the same number of console replicas as the number of masters, minimum 2
+  # there should be the same number of console replicas as the number of masters, minimum 3,
+  # unless deploying in a dev environment with a unique configuration, such as a single node setup.
   count: 3
   logging:
     level: 0


examples/cr.yaml
```diff
@@ -15,7 +15,8 @@ spec:
   # - Unmanaged indicates that the operator should idle and ignore the console and its sub-resources
   # - Removed indicates that the operator should delete all of the sub-resources of the console
   managementState: Managed
-  # there should be the same number of console replicas as the number of masters, minimum 2
+  # there should be the same number of console replicas as the number of masters, minimum 3,
+  # unless deploying in a dev environment with a unqiue configuration, such as a single node setup.
```
jhadvig: r/unqiue/unique

benjaminapetersen: Got it, thx!

Since this is a comment in what is basically a docs file, I would assume it is not subject to the feature freeze - @spadgett ?

spadgett: Correct, this doesn't need an exception.

Is the updated comment correct? Right now we hardcode 3 replicas regardless of number of masters.

benjaminapetersen: Well, it's an example file, intended to copy, paste, tweak as desired.  So the "should be the same number of console replicas as the number of masters" is sort of a guideline. If you have a single node cluster, probably should reduce the replicas to 1.

PR #102, opened 2019-01-09
Add node selector to operator deployment to deploy on master node

benjaminapetersen:


manifests/05-operator.yaml
```diff
@@ -13,6 +13,8 @@ spec:
       labels:
         name: console-operator
     spec:
+      nodeSelector:
+        "node-role.kubernetes.io/master": ""
```
jhadvig: NIT: I think it should be
```
node-role.kubernetes.io/master: ''
```

benjaminapetersen: Yeah, it's a weird annotation, has no value.  But its consistent with elsewhere. 🤷‍♂️

spadgett: > Just out of curiosity, why the value is empty ?

You have to match the label on the node, and that's what set for masters.

```yaml
  labels:
    beta.kubernetes.io/arch: amd64
    beta.kubernetes.io/instance-type: m4.large
    beta.kubernetes.io/os: linux
    failure-domain.beta.kubernetes.io/region: us-east-2
    failure-domain.beta.kubernetes.io/zone: us-east-2a
    kubernetes.io/hostname: ip-10-0-13-185
    node-role.kubernetes.io/master: ''
```

spadgett: We might want to just restrict the operator to linux if it doesn't matter if it runs on the master.

```suggestion
        beta.kubernetes.io/os: linux
```

jhadvig: > You have to match the label on the node, and that's what set for masters.

Have spoken to @mfojtik regarding the additional labels, and was told that in order to run the operator on one of the masters the given `nodeSelector` should be enough. 
Regarding specifying the linux machine, that shouldn't be necessary since every master is running on linux machine.

spadgett: Yeah, I'm not saying add both. I'm saying use that instead of the master node selector.

Let me follow up on this today. I'm not sure what the recommendation is for operators.

benjaminapetersen: Hmm... Yeah seems like we can loosen up to any Linux for the operator, so long as the console itself is on masters. Not a big deal if it gets destroyed and recreated.

spadgett: > From what I was told the recommendation is just the node selector, although there might be more to it.

Right, it's still a node selector. Just a different value if we don't require the operator to run on the masters.

jhadvig: Not sure why we don't would want to run operator on master.

spadgett: I checked with @derekwaynecarr, and he recommended letting the operator run on any linux node by default. We'll need to add a config option to let admins configure the node selector later, but for now, let's go with:

```yaml
nodeSelector:
  beta.kubernetes.io/os: linux
```

spadgett: I just realized that it's not possible to make operator node selector configurable since this is a static manifest that the CVO applies. I raised the issue with Derek.

spadgett: OK, change of plans. We've been asked to run the operator on the master :)

benjaminapetersen:
Updated.

spadgett:
Do we need a toleration? It looks like master nodes have a taint.

```yaml
  taints:
    - key: node-role.kubernetes.io/master
      effect: NoSchedule
```

spadgett:
Console deployment has it.

```yaml
      tolerations:
        - key: node-role.kubernetes.io/master
          operator: Exists
          effect: NoSchedule
```

jhadvig:
I guess we should add it
https://github.com/openshift/machine-api-operator/blob/bfcb1819b913e1ba1843b3c719804e6b2ed1f19e/test/integration/manifests/operator-deployment.yaml#L44-L47

benjaminapetersen:
Yup, this should do it.

benjaminapetersen:
/hold cancel

Tested and this works. Initially when I did an `oc apply -f manifests/05-console.yaml` I had some fluctuation with the `operator` `pods`, there was a `CrashLoopBackoff`, then naturally resolved into happy pods:

```bash 
oc get deployments
NAME                DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
console-operator    3         3         3            3           2h
openshift-console   3         3         3            3           2h

oc get pods
NAME                                READY     STATUS    RESTARTS   AGE
console-operator-5cf5f97b97-qrhzl   1/1       Running   0          11m
console-operator-5cf5f97b97-vjn6b   1/1       Running   0          12m
console-operator-5cf5f97b97-wj9wh   1/1       Running   0          12m
openshift-console-c99f58c6f-559rh   1/1       Running   0          2h
openshift-console-c99f58c6f-9lc26   1/1       Running   0          2h
openshift-console-c99f58c6f-g2pnp   1/1       Running   0          2h
```

PR #107, opened 2019-01-14
Console operator e2e test - Managed state

jhadvig:
Adding e2e tests for console-operator *Managed* state.

Added the suggested check if the console-operator is up. Then The test will remove the console-operator deployment and will check if the deployment is recreated.

to launch the tests, call `make test-e2e`, whre the **KUBECONFIG** envar has to be set

@benjaminapetersen PTAL

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -11,8 +11,10 @@ const (
 
 // consts to maintain existing names of various sub-resources
 const (
-	OpenShiftConsoleName      = "openshift-console"
-	OpenShiftConsoleShortName = "console"
-	OpenShiftConsoleNamespace = "openshift-console"
-	OAuthClientName           = OpenShiftConsoleName
+	OpenShiftConsoleName              = "openshift-console"
+	OpenShiftConsoleShortName         = "console"
+	OpenShiftConsoleNamespace         = "openshift-console"
+	OpenShiftConsoleOperatorNamespace = "openshift-console"
+	OpenShiftConsoleOperator          = "console-operator"
+	OAuthClientName                   = OpenShiftConsoleName
 )
--- a/pkg/testframework/clientset.go
+++ b/pkg/testframework/clientset.go
@@ -1,0 +1,62 @@
+package testframework
+
+import (
+	"fmt"
+	"testing"
+
+	clientappsv1 "k8s.io/client-go/kubernetes/typed/apps/v1"
+	clientcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	restclient "k8s.io/client-go/rest"
+
+	clientconfigv1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+	clientconsolev1alpha1 "github.com/openshift/console-operator/pkg/generated/clientset/versioned/typed/console/v1alpha1"
+)
+
+// Clientset is a set of Kubernetes clients.
+type Clientset struct {
+	clientcorev1.CoreV1Interface
+	clientappsv1.AppsV1Interface
+	clientconfigv1.ConfigV1Interface
+	clientconsolev1alpha1.ConsoleV1alpha1Interface
+}
+
+// NewClientset creates a set of Kubernetes clients. The default kubeconfig is
+// used if not provided.
+func NewClientset(kubeconfig *restclient.Config) (*Clientset, error) {
+	var err error
+	if kubeconfig == nil {
+		kubeconfig, err = GetConfig()
+		if err != nil {
+			return nil, fmt.Errorf("unable to get kubeconfig: %s", err)
+		}
+	}
+
+	clientset := &Clientset{}
+	clientset.CoreV1Interface, err = clientcorev1.NewForConfig(kubeconfig)
+	if err != nil {
+		return nil, err
+	}
+	clientset.AppsV1Interface, err = clientappsv1.NewForConfig(kubeconfig)
+	if err != nil {
+		return nil, err
+	}
+	clientset.ConfigV1Interface, err = clientconfigv1.NewForConfig(kubeconfig)
+	if err != nil {
+		return nil, err
+	}
+	clientset.ConsoleV1alpha1Interface, err = clientconsolev1alpha1.NewForConfig(kubeconfig)
+	if err != nil {
+		return nil, err
+	}
+	return clientset, nil
+}
+
+// MustNewClientset is like NewClienset but aborts the test if clienset cannot
+// be constructed.
+func MustNewClientset(t *testing.T, kubeconfig *restclient.Config) *Clientset {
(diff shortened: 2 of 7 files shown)


pkg/api/api.go
```diff
@@ -11,8 +11,10 @@ const (
 
 // consts to maintain existing names of various sub-resources
 const (
-	OpenShiftConsoleName      = "openshift-console"
-	OpenShiftConsoleShortName = "console"
-	OpenShiftConsoleNamespace = "openshift-console"
-	OAuthClientName           = OpenShiftConsoleName
+	OpenShiftConsoleName              = "openshift-console"
```
benjaminapetersen: Nit: If these all ought to match, should they just all `= OpenShiftConsoleName`?

pkg/client/client.go
```diff
@@ -0,0 +1,41 @@
...
+// Otherwise will assume running in cluster and use the cluster provided kubeconfig.
+//
+// Config precedence
+//
+// * KUBECONFIG environment variable pointing at a file
+//
+// * In-cluster config if running in cluster
+//
+// * $HOME/.kube/config if exists
+func GetConfig() (*rest.Config, error) {
```
benjaminapetersen: I'm guessing this is based on code from somewhere?  Is `client.go` the best package name? Just thinking about it as it only contains `GetConfig()`.

test/e2e/main_test.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		fmt.Println("failed waiting for operator to start")
+		os.Exit(1)
+	}
+
+	os.Exit(m.Run())
+}
+
+func waitForOperator() error {
+	depClient := kubeClient.AppsV1().Deployments(consoleapi.OpenShiftConsoleOperatorNamespace)
```
benjaminapetersen: Do we need to worry about the state of the deployment, perhaps for flakes?

test/e2e/managed_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	consoleapi "github.com/openshift/console-operator/pkg/api"
+	"github.com/openshift/console-operator/pkg/testframework"
+)
+
+func TestManaged(t *testing.T) {
+	client := testframework.MustNewClientset(t, nil)
```
benjaminapetersen: Do we assume each test in isolation, ie should call `err = waitForOperator()` when they start?

pkg/testframework/framework.go
```diff
@@ -0,0 +1,59 @@
...
+var (
+	// AsyncOperationTimeout is how long we want to wait for asynchronous
+	// operations to complete. ForeverTestTimeout is not long enough to create
+	// several replicas and get them available on a slow machine.
+	// Setting this to 5 minutes:w
+
+	AsyncOperationTimeout = 5 * time.Minute
+)
+
+// DeleteCompletely sends a delete request and waits until the resource and
```
benjaminapetersen: Can we use this for any resource?  It appears so, and trust it will always delete all dependents.

pkg/testframework/clientset.go
```diff
@@ -0,0 +1,62 @@
...
+func NewClientset(kubeconfig *restclient.Config) (clientset *Clientset, err error) {
+	if kubeconfig == nil {
+		kubeconfig, err = client.GetConfig()
+		if err != nil {
+			return nil, fmt.Errorf("unable to get kubeconfig: %s", err)
+		}
+	}
+
+	clientset = &Clientset{}
+	clientset.CoreV1Interface, err = clientcorev1.NewForConfig(kubeconfig)
```
benjaminapetersen: We want to swallow all of the`err` below?  Should we return the `nil, err` instead?

pkg/api/api.go
jhadvig: Was thinking about it but the semantics between name and namespace are a bit different. But agreeing on that :)

pkg/client/client.go
jhadvig: yes, its from https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/client/client.go#L35-L53

Was thinking about creating a helper for the e2e tests but this might come handy in the codebase as well. Wasnt really sure. But I guess I can put it into the testframework. Thoughts ? :)

test/e2e/main_test.go
jhadvig: I think it's good practice to check if the resource, which functionality we are testing is checked for availability :)

test/e2e/managed_test.go
jhadvig: hmm good question. I would say that initial check should be enough, if we are only going to manipulate with the state of the console-operator. On the other hand it can't  do any harm to check each time?

pkg/testframework/framework.go
jhadvig: yes it should

pkg/testframework/clientset.go
jhadvig: the `err` is not shallowed. Checked the function return values.
```go
(clientset *Clientset, err error)
```

they are going to be returned.

benjaminapetersen: Right. I think we prob want to avoid bare returns.  So far I have in this repo, generally seems the consensus is against bare returns (though they still exist in some older code in origin).

benjaminapetersen: I should not have said "swallowed", my bad.

test/e2e/managed_test.go
benjaminapetersen: I think we should probably strive to write each e2e test in isolation.

pkg/testframework/clientset.go
jhadvig: no worries ! ... if we dont use the default returns across the codebase lets just dont use them ! :+1:

test/e2e/managed_test.go
jhadvig: sorry for the confusion... I've just realize that the `err = waitForOperator()` that's called in the TestMain function, will be run for each test :)

jhadvig:
@benjaminapetersen comments address and the tests are green :)

pkg/client/client.go
benjaminapetersen: Wondering if we should just copy the whole file, if the 3-4 other functions help clarify intent.  Hmm.

pkg/api/api.go
benjaminapetersen: Yeah, I hear ya.  It just happens that the strings match.

benjaminapetersen:
@jhadvig great! looking forward to having some coverage.

/lgtm

PR #111, opened 2019-01-16
Set a 10m CPU request for console operator and console

spadgett:
/assign @jhadvig @benjaminapetersen


spadgett:
Neither our operator or the console itself is CPU intensive. The guidance is to set CPU request to 10m.

PR #112, opened 2019-01-21
Add e2e test for console operator Unmanaged and Removed states

jhadvig:
/assign @benjaminapetersen 

Adding Unmanaged and Removed test cases.
Each tests is testing several resources from the "openshift-console" namespace - Deployment/ConfigMap/Service/Route and their availability/unavailability for each operator state.
 
Also added `logs.go` that I've get [from](https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/testframework/logs.go) to get better logs.

### DIFF
--- a/pkg/testframework/console.go
+++ b/pkg/testframework/console.go
@@ -1,0 +1,152 @@
+package testframework
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	// operatorapi "github.com/openshift/api/operator/v1"
+	operatorsv1alpha1 "github.com/openshift/api/operator/v1alpha1"
+
+	consoleapi "github.com/openshift/console-operator/pkg/api"
+	v1alpha1 "github.com/openshift/console-operator/pkg/apis/console/v1alpha1"
+)
+
+func isOperatorManaged(cr *v1alpha1.Console) bool {
+	return cr.Spec.ManagementState == operatorsv1alpha1.Managed
+}
+
+func isOperatorUnmanaged(cr *v1alpha1.Console) bool {
+	return cr.Spec.ManagementState == operatorsv1alpha1.Unmanaged
+}
+
+func isOperatorRemoved(cr *v1alpha1.Console) bool {
+	return cr.Spec.ManagementState == operatorsv1alpha1.Removed
+}
+
+type operatorStateReactionFn func(cr *v1alpha1.Console) bool
+
+func ensureConsoleIsInDesiredState(t *testing.T, client *Clientset, state operatorsv1alpha1.ManagementState) error {
+	var cr *v1alpha1.Console
+	// var checkFunc func()
+	var checkFunc operatorStateReactionFn
+
+	switch state {
+	case operatorsv1alpha1.Managed:
+		checkFunc = isOperatorManaged
+	case operatorsv1alpha1.Unmanaged:
+		checkFunc = isOperatorUnmanaged
+	case operatorsv1alpha1.Removed:
+		checkFunc = isOperatorRemoved
+	}
+
+	err := wait.Poll(1*time.Second, AsyncOperationTimeout, func() (stop bool, err error) {
+		cr, err = client.ConsoleV1alpha1Interface.Consoles(consoleapi.OpenShiftConsoleOperatorNamespace).Get(consoleapi.ResourceName, metav1.GetOptions{})
+		if err != nil {
+			return false, err
+		}
+		return checkFunc(cr), nil
+	})
+	if err != nil {
+		DumpObject(t, "the latest observed state of the console resource", cr)
+		DumpOperatorLogs(t, client)
+		return fmt.Errorf("failed to wait to change console operator state to 'Removed': %s", err)
+	}
+	return nil
+}
+
+func ManageConsole(t *testing.T, client *Clientset) error {
+	cr, err := client.ConsoleV1alpha1Interface.Consoles(consoleapi.OpenShiftConsoleOperatorNamespace).Get(consoleapi.ResourceName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	if isOperatorManaged(cr) {
+		t.Logf("console operator already in 'Managed' state")
+		return nil
+	}
+
+	t.Logf("changing console operator state to 'Managed'...")
+
+	_, err = client.ConsoleV1alpha1Interface.Consoles(consoleapi.OpenShiftConsoleOperatorNamespace).Patch(consoleapi.ResourceName, types.MergePatchType, []byte(`{"spec": {"managementState": "Managed"}}`))
+	if err != nil {
+		return err
+	}
+	if err := ensureConsoleIsInDesiredState(t, client, operatorsv1alpha1.Managed); err != nil {
+		return fmt.Errorf("unable to change console operator state to 'Managed': %s", err)
+	}
+
+	return nil
+}
+
+func UnmanageConsole(t *testing.T, client *Clientset) error {
+	cr, err := client.ConsoleV1alpha1Interface.Consoles(consoleapi.OpenShiftConsoleOperatorNamespace).Get(consoleapi.ResourceName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	if isOperatorUnmanaged(cr) {
+		t.Logf("console operator already in 'Unmanaged' state")
+		return nil
+	}
+
+	t.Logf("changing console operator state to 'Unmanaged'...")
+
+	_, err = client.ConsoleV1alpha1Interface.Consoles(consoleapi.OpenShiftConsoleOperatorNamespace).Patch(consoleapi.ResourceName, types.MergePatchType, []byte(`{"spec": {"managementState": "Unmanaged"}}`))
+	if err != nil {
+		return err
+	}
+	if err := ensureConsoleIsInDesiredState(t, client, operatorsv1alpha1.Unmanaged); err != nil {
+		return fmt.Errorf("unable to change console operator state to 'Unmanaged': %s", err)
+	}
+
+	return nil
+}
+
+func RemoveConsole(t *testing.T, client *Clientset) error {
+	cr, err := client.ConsoleV1alpha1Interface.Consoles(consoleapi.OpenShiftConsoleOperatorNamespace).Get(consoleapi.ResourceName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	if isOperatorRemoved(cr) {
+		t.Logf("console operator already in 'Removed' state")
+		return nil
+	}
+
+	t.Logf("changing console operator state to 'Removed'...")
+
+	_, err = client.ConsoleV1alpha1Interface.Consoles(consoleapi.OpenShiftConsoleOperatorNamespace).Patch(consoleapi.ResourceName, types.MergePatchType, []byte(`{"spec": {"managementState": "Removed"}}`))
+	if err != nil {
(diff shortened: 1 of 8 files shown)


pkg/testframework/console.go
```diff
@@ -0,0 +1,162 @@
+package testframework
```
benjaminapetersen: Thinking about this now, it might be nice to split it up.  Right now it seems to be all the funcs needed just to test management state.  I'd prob rather separate out feature tests than keep adding to this one file:

```bash
/pkg
   |_ testframework
        |_ console
            |_ managementstate.go
            |_ otherfeature.go
            |_ yetanotherfeature.go
        |_ deployment
        |_ etc
```
Thoughts?

pkg/testframework/framework.go
```diff
@@ -19,6 +23,19 @@ var (
 	AsyncOperationTimeout = 5 * time.Minute
 )
 
+// Logger is an interface to report events from tests. It is implemented by
```
benjaminapetersen: I'm wondering if `framework.go` might grow into a kitchen sink of a file.  Maybe we can break this up now & encourage future growth into separate files to keep it tidy down the road.

benjaminapetersen: (I realize there isn't much here yet, so it could just be `/framework/logging.go` & `framework/resourceavailabilty.go` or something a bit like that.)

benjaminapetersen: I see you have logs file below already. Nice.  Was just thinking about `DumpObject`... hmm.

jhadvig: Yep thats a good question. But I think it mainly depends on how beefy we want our e2e tests for the console operator. If it should just test each operator state(maybe some extra tests, scaling operator down to 0 and back to 1), I wouldn't worry about the scale of the file :)

test/e2e/managed_test.go
```diff
@@ -2,19 +2,24 @@ package e2e_test
...
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/wait"
 
 	consoleapi "github.com/openshift/console-operator/pkg/api"
 	"github.com/openshift/console-operator/pkg/testframework"
 )
 
+// TestManaged sets console-operator to Managed state. After that "openshift-console
+// deployment is deleted after which the deploymnet is tested for availability, together
```
benjaminapetersen: `deploymnet` -> `deployment`

test/e2e/managed_test.go
```diff
@@ -28,18 +33,15 @@ func TestManaged(t *testing.T) {
...
-		if errors.IsNotFound(err) {
-			return false, nil
-		}
-		return false, err
-	})
-	if err != nil {
-		t.Fatal(err)
+
+	errChan := make(chan error)
+	go testframework.IsResourceAvailable(errChan, client, "ConfigMap")
```
benjaminapetersen: Do we want to delete each of these resources & then check each availability?

pkg/testframework/framework.go
```diff
@@ -57,3 +74,61 @@ func DeleteCompletely(getObject func() (metav1.Object, error), deleteObject func
 		return obj.GetUID() != uid, nil
 	})
 }
+
+// IsResourceAvailable checks if tested resource is available(recreated by console-ioerator),
```
benjaminapetersen: `ioerator` -> `operator`

test/e2e/unmanaged_test.go
```diff
@@ -0,0 +1,47 @@
...
+		func(deleteOptions *metav1.DeleteOptions) error {
+			return client.Deployments(consoleapi.OpenShiftConsoleOperatorNamespace).Delete(consoleapi.OpenShiftConsoleName, deleteOptions)
+		},
+	); err != nil {
+		t.Fatalf("unable to delete console deployment: %s", err)
+	}
+
+	t.Logf("waiting to check if the operator has not recreate console deployment...")
+	errChan := make(chan error)
+	go testframework.IsResourceAvailable(errChan, client, "ConfigMap")
```
benjaminapetersen: Should we be asking for a named resources?  For example, we have 2 configmaps now.

benjaminapetersen: (other resources are singletons, though there is also more than 1 secret in the namespace)

pkg/testframework/console.go
jhadvig: Makes sense to split it here, but Im not sure what the otherfeatures will be. The ones that pop to my head are, to start/stop the operator by scaling it up and down, but that can be on a `/pkg/testframework/operator.go` file.
Not sure what else at this state of the console-operator we should test.

test/e2e/managed_test.go
jhadvig: No only the `openshift-console` deployment is deleted. Wasnt sure if we want to delete all, or just some, or only one.

test/e2e/unmanaged_test.go
jhadvig: yeah, so here I wasn't sure if we want to nuke all the resources, or just some, or just one?
But since the operator should recreate any resource it should be enough to just test some resources. Thoughts ? :)

test/e2e/managed_test.go
benjaminapetersen: Right, only the `deployment` is deleted, but it checks if each of the other resources are still there.

pkg/testframework/framework.go
```diff
@@ -57,3 +74,61 @@ func DeleteCompletely(getObject func() (metav1.Object, error), deleteObject func
...
+		}
+		if counter == 10 {
+			return true, nil
+		}
+		return false, nil
+	})
+	errChan <- err
+}
+
+func getResource(client *Clientset, resource string) error {
```
benjaminapetersen: Thinking about this:

```golang
// current unnamed
getResource(client, 'ConfigMap')
// vs named?
getResource(client, 'configmap', 'console-config')
// vs specific-to-client funcs?
getConfigmapByName('console-config')
getSecretByName('oauth-config')
```

test/e2e/managed_test.go
jhadvig: yes, thinking about it now, I probably would be better to do it the opposite way. Delete more resource (eg.Deployment, Route, ConfigMap). Then check for all the resources.

pkg/testframework/console.go
```diff
@@ -0,0 +1,162 @@
...
+		cr, err = client.ConsoleV1alpha1Interface.Consoles(consoleapi.OpenShiftConsoleOperatorNamespace).Get(consoleapi.ResourceName, metav1.GetOptions{})
+		if err != nil {
+			return false, err
+		}
+		return checkFunc(cr), nil
+	})
+	if err != nil {
+		DumpObject(logger, "the latest observed state of the console resource", cr)
+		DumpOperatorLogs(logger, client)
+		return fmt.Errorf("failed to wait to change console operator state to 'Removed': %s", err)
```
benjaminapetersen: This branch is always `operator state to 'Removed'`?

pkg/testframework/framework.go
jhadvig: Makes sense to have resource based getters :+1:

pkg/testframework/console.go
```diff
@@ -0,0 +1,162 @@
...
+	return cr.Spec.ManagementState == operatorsv1alpha1.Unmanaged
+}
+
+func isOperatorRemoved(cr *v1alpha1.Console) bool {
+	return cr.Spec.ManagementState == operatorsv1alpha1.Removed
+}
+
+type operatorStateReactionFn func(cr *v1alpha1.Console) bool
+
+func ensureConsoleIsInDesiredState(logger Logger, client *Clientset, state operatorsv1alpha1.ManagementState) error {
```
benjaminapetersen: So `ensureConsoleIsInDesiredState`leads me to think ALL console state, not just `ManagementState`.   I would probably expect the world `ensure` does an `UPDATE` as well to set the state if it is not already correct, but it looks like the `PATCH` is in the functions wrapping this call. 

Thinking about this.

pkg/testframework/console.go
```diff
@@ -0,0 +1,162 @@
...
+		logger.Logf("console operator already in 'Managed' state")
+		return nil
+	}
+
+	logger.Logf("changing console operator state to 'Managed'...")
+
+	_, err = client.ConsoleV1alpha1Interface.Consoles(consoleapi.OpenShiftConsoleOperatorNamespace).Patch(consoleapi.ResourceName, types.MergePatchType, []byte(`{"spec": {"managementState": "Managed"}}`))
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return nil
```
benjaminapetersen: We are ok here if `errors.IsNotFound`?

pkg/testframework/console.go
jhadvig: nope, will update ! :)

pkg/testframework/console.go
jhadvig: no we can get rid of that :+1:

pkg/testframework/console.go
jhadvig: yeah, probably the wording is not the best.
What about `isConsoleInDesiredState` || `checkConsoleForDesiredState` ?

benjaminapetersen: or `isManagementStateManaged()`, `isManagementStateRemoved()`?

pkg/testframework/console.go
benjaminapetersen: Yeah I bet we want it to fail in this case.

test/e2e/unmanaged_test.go
benjaminapetersen: Agree, though it might be good to know for certain it is recreating all of the expected resources.  If our [listOpts funcs](https://github.com/openshift/console-operator/blob/master/pkg/console/starter/starter.go#L72) are ever updated, it could be possible the operator would stop paying attention to a resource it needs to manage.  Currently the funcs are `noop` except for the `OAuth` client func.

pkg/testframework/framework.go
```diff
@@ -19,6 +23,19 @@ var (
 	AsyncOperationTimeout = 5 * time.Minute
 )
 
+// Logger is an interface to report events from tests. It is implemented by
+// testing.T.
+type Logger interface {
```
benjaminapetersen: Should `Logger interface` be in the `logs.go` file?

benjaminapetersen: Chewing on the logger.  In the rest of the project, I've been using `logrus`.

benjaminapetersen: Mostly just asking "what do we gain doing it differently here".

pkg/testframework/framework.go
benjaminapetersen: Realized it doesn't actually get the resource, it returns only an error.  So if you don't want to make specific ones, since this is more of a helper, it could be something like `resourceExists()`

jhadvig:
@benjaminapetersen updated the PR based on our conversation. PTAL :)

pkg/testframework/console.go
benjaminapetersen: Oh, we will probably test the top level `console` config, which has, for example, `LogoutRedirect`. Also the `Operator` config itself has `Branding`, `DocumentationBaseURL`, as well as a `NodeSelector` override.  

Eventually we may support custom `Route` or perhaps a set of `Routes` (in discussion).

pkg/testframework/framework.go
benjaminapetersen: Probably related to the above comment, we have a set of other features to test, but uncertain how much it will grow over time (I dont expect tons).

jhadvig:
Last e2e-aws failure
```

jhadvig:
@benjaminapetersen we are green :)

benjaminapetersen:
/lgtm

awesome.

benjaminapetersen:
yay 🦄

PR #114, opened 2019-01-22
[WIP] Add NodeSelector override to Console CR

benjaminapetersen:
Per the conversation here https://github.com/openshift/api/pull/159#issuecomment-453244019.


benjaminapetersen:
/assign @spadgett 

I'm assuming we want an `Overrides` struct to wrap this. If we are good with it, I can add the logic to apply it.

benjaminapetersen:
We have a few values to update, not just `NodeSelector`.

PR #115, opened 2019-01-22
[WIP][DO_NOT_MERGE] Console/cr/update

benjaminapetersen:
Updates for the console CR.
/assign @spadgett 
Related to [global config](https://github.com/openshift/api/pull/159).


manifests/00-crd.yaml
```diff
@@ -1,13 +1,17 @@
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
-  name: consoles.console.openshift.io
+  name: consoleoperatorconfigs.console.operator.openshift.io
```
benjaminapetersen: @spadgett fyi updating the CRD to conform with the most current patterns.

manifests/00-crd.yaml
```diff
@@ -1,13 +1,17 @@
...
-    plural: consoles
-    singular: console
-  scope: Namespaced
-  version: v1alpha1
+    kind: ConsoleOperatorConfig
+    listKind: ConsoleOperatorConfigList
+    plural: consoleoperatorconfigs
+    singular: consoleoperatorconfig
+    categories:
+    - coreoperators
```
benjaminapetersen: `oc get coreoperators -o yaml` should include this, the console operator config.

manifests/00-crd.yaml
benjaminapetersen: `consoleoperatorconfigs` is what this represents now. `console.operator.openshift.io` is our group. This follows the pattern from [kube api server operator](https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/manifests/0000_10_kube-apiserver-operator_01_config.crd.yaml#L4).

spadgett:
RBAC policy will need to be updated, too

benjaminapetersen:
outdated.

PR #117, opened 2019-01-29
Test editing resources in Managed and Unmanaged console-operator state

jhadvig:
@benjaminapetersen adding promised tests.
Per our discussion, the tests are editing the resources separately, not bulk, since they have different semantics for the test.
The tests are testing ConfigMap, Service and Route. I've skipped the Deployment since the console operator is not managing changes user makes to it(which is kinda weird?).
 
Also made the `getResource()` public -> `GetResource()`, so it can be used on different places + now it's returning runtime.Object so different fields object field can be accessed.

PTAL

### DIFF
--- a/pkg/testframework/framework.go
+++ b/pkg/testframework/framework.go
@@ -5,8 +5,13 @@ import (
 	"testing"
 	"time"
 
+	routev1 "github.com/openshift/api/route/v1"
+	appv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	runtime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	consoleapi "github.com/openshift/console-operator/pkg/api"
@@ -27,8 +32,8 @@ func DeleteAll(t *testing.T, client *Clientset) {
 	for _, resource := range resources {
 		t.Logf("deleting console %s...", resource)
 		if err := DeleteCompletely(
-			func() (metav1.Object, error) {
-				return getResource(client, resource)
+			func() (runtime.Object, error) {
+				return GetResource(client, resource)
 			},
 			func(*metav1.DeleteOptions) error {
 				return deleteResource(client, resource)
@@ -39,8 +44,8 @@ func DeleteAll(t *testing.T, client *Clientset) {
 	}
 }
 
-func getResource(client *Clientset, resource string) (metav1.Object, error) {
-	var res metav1.Object
+func GetResource(client *Clientset, resource string) (runtime.Object, error) {
+	var res runtime.Object
 	var err error
 	switch resource {
 	case "ConfigMap":
@@ -76,7 +81,7 @@ func deleteResource(client *Clientset, resource string) error {
 
 // DeleteCompletely sends a delete request and waits until the resource and
 // its dependents are deleted.
-func DeleteCompletely(getObject func() (metav1.Object, error), deleteObject func(*metav1.DeleteOptions) error) error {
+func DeleteCompletely(getObject func() (runtime.Object, error), deleteObject func(*metav1.DeleteOptions) error) error {
 	obj, err := getObject()
 	if err != nil {
 		if errors.IsNotFound(err) {
@@ -85,7 +90,7 @@ func DeleteCompletely(getObject func() (metav1.Object, error), deleteObject func
 		return err
 	}
 
-	uid := obj.GetUID()
+	uid := getUID(obj)
 
 	policy := metav1.DeletePropagationForeground
 	if err := deleteObject(&metav1.DeleteOptions{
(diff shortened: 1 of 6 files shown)


benjaminapetersen:
legitimate test errors i think

pkg/testframework/framework.go
```diff
@@ -39,8 +44,8 @@ func DeleteAll(t *testing.T, client *Clientset) {
 	}
 }
 
-func getResource(client *Clientset, resource string) (metav1.Object, error) {
-	var res metav1.Object
+func GetResource(client *Clientset, resource string) (runtime.Object, error) {
```
benjaminapetersen: `runtime.Object` too generic?  I think we lose type safety if we use these.  Necessary?

pkg/testframework/framework.go
```diff
@@ -76,7 +81,7 @@ func deleteResource(client *Clientset, resource string) error {
 
 // DeleteCompletely sends a delete request and waits until the resource and
 // its dependents are deleted.
-func DeleteCompletely(getObject func() (metav1.Object, error), deleteObject func(*metav1.DeleteOptions) error) error {
+func DeleteCompletely(getObject func() (runtime.Object, error), deleteObject func(*metav1.DeleteOptions) error) error {
```
benjaminapetersen: `runtime.Object` same as above, too generic?

pkg/testframework/framework.go
```diff
@@ -109,16 +114,33 @@ func DeleteCompletely(getObject func() (metav1.Object, error), deleteObject func
 			return false, err
 		}
 
-		return obj.GetUID() != uid, nil
+		return getUID(obj) != uid, nil
 	})
 }
 
+func getUID(obj runtime.Object) types.UID {
+	configMap, ok := obj.(*corev1.ConfigMap)
```
benjaminapetersen: Wondering if passing a `runtime.Object` then casting it back is the best approach?

pkg/testframework/framework.go
```diff
@@ -134,12 +156,31 @@ func IsResourceAvailable(errChan chan error, client *Clientset, resource string)
 	errChan <- err
 }
 
+func IsResourceAvailable_(client *Clientset, resource string) error {
```
benjaminapetersen: Another way to do the same thing?

test/e2e/util.go
```diff
@@ -0,0 +1,103 @@
...
+
+// Basically each of these tests helpers are similar, they only vary in the
+// resource they are GETting, PATCHing and patch itself.
+// Since after the patch is done, the test needs to wait, till it can GET
+// patched object(or not, if the operator status is set to Managed).
+
+func patchAndCheckConfigMap(t *testing.T, client *testframework.Clientset) bool {
+	res, err := testframework.GetResource(client, "ConfigMap")
+	errorCheck(t, err)
+	configMap, ok := res.(*corev1.ConfigMap)
```
benjaminapetersen: The casting again concerns me a bit, I see it in all of these functions.  Wondering if its a signal the generalized `GetResource` ought to be a specific function for each resource type w/a specific return type?

test/e2e/util.go
```diff
@@ -0,0 +1,103 @@
...
+	route, ok = res.(*routev1.Route)
+	if !ok {
+		t.Fatalf("unable to type received object as Route")
+	}
+	newData := route.Spec.Port.TargetPort
+
+	return reflect.DeepEqual(originalData, newData)
+}
+
+func errorCheck(t *testing.T, err error) {
```
benjaminapetersen: I sympathize with the helper to get rid of the if block :)

So I think the recommended path is to always return the error, and handle them only at the top in one place.

pkg/testframework/framework.go
jhadvig: yeah its too generic and we dont need all of the interface features. Was speaking with @mfojtik and he advised me to use runtime.Object, since its more generic and there is a smaller chance you get an object that we cant handle

pkg/testframework/framework.go
jhadvig: yeah, just found out that the idiomatic way is to pass the runtime.Object into meta.Accessor which will return meta.Object

jhadvig: Using meta.Accessor is also more safe cause we can get error if passed object is not supported.

pkg/testframework/framework.go
jhadvig: As commented above :)

pkg/testframework/framework.go
jhadvig: yeah... sorry for that ... forgot to clean this :D

test/e2e/util.go
jhadvig: yeah was thinking about it and thought it would be better to have an single get function for all the types then multiple get functions for each type that would also cast object directly. Wasnt really sure which one is more idiomatic.

jhadvig:
@benjaminapetersen updated per our conversation

benjaminapetersen:
```
clusterversion.config.openshift.io/version condition met
hack/test-e2e.sh
[INFO] Running e2e tests
test/e2e/util.go:10:2: cannot find package "k8s.io/kubernetes/pkg/util/wait" in any of:
	/go/src/github.com/openshift/console-operator/vendor/k8s.io/kubernetes/pkg/util/wait (vendor tree)
	/usr/local/go/src/k8s.io/kubernetes/pkg/util/wait (from $GOROOT)
	/go/src/k8s.io/kubernetes/pkg/util/wait (from $GOPATH)
```
that one legitimate?

PR #119, opened 2019-01-31
New version bump

jhadvig:
@benjaminapetersen here is the version bump PR, splitted into separate commits.
Reporting the operator status will be following a followUp

@spadgett fyi


pkg/console/operator/sync_v400.go
```diff
@@ -114,10 +114,16 @@ func sync_v400(co *consoleOperator, consoleConfig *v1alpha1.Console) (*v1alpha1.
...
+		{
+			Group:          "apps/v1",
+			Resource:       "deployments",
+			Namespace:      defaultDeployment.Namespace,
+			Name:           defaultDeployment.Name,
+			LastGeneration: defaultDeployment.Generation,
+		},
 	}
-	deploymentGeneration := resourcemerge.ExpectedDeploymentGenerationV1alpha1(defaultDeployment, versionAvailability)
+	deploymentGeneration := resourcemerge.ExpectedDeploymentGeneration(defaultDeployment, versionAvailability)
```
jhadvig: @deads2k can you please confirm that this is the right usage of this function? cause I haven't found any other usage across other operator repositories. Thanks !

benjaminapetersen: Yeah, I found this one a touch tricky as well.

pkg/console/operator/sync_v400.go
```diff
@@ -114,10 +114,16 @@ func sync_v400(co *consoleOperator, consoleConfig *v1alpha1.Console) (*v1alpha1.
 func SyncDeployment(co *consoleOperator, recorder events.Recorder, consoleConfig *v1alpha1.Console, cm *corev1.ConfigMap, serviceCAConfigMap *corev1.ConfigMap, sec *corev1.Secret) (*appsv1.Deployment, bool, error) {
 	logrus.Printf("validating console deployment...")
 	defaultDeployment := deploymentsub.DefaultDeployment(consoleConfig, cm, serviceCAConfigMap, sec)
```
deads2k: nit, I prefer `required` or `requiredDeployment`

pkg/console/operator/sync_v400.go
```diff
@@ -114,10 +114,16 @@ func sync_v400(co *consoleOperator, consoleConfig *v1alpha1.Console) (*v1alpha1.
...
-	versionAvailability := &operatorv1alpha1.VersionAvailability{
-		Version: consoleConfig.Spec.Version,
+	versionAvailability := []operatorv1.GenerationStatus{
```
deads2k: you want this to come from your actual status.  Compare to https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master/pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go#L280

jhadvig: @deads2k we are using  "github.com/openshift/api/operator/v1alpha1" not "github.com/openshift/api/operator/v1" , where the operator status does not contain the `generations` field ->
https://github.com/openshift/api/blob/master/operator/v1alpha1/types.go#L105-L122
That's why are we need to constructing it this way.

benjaminapetersen: We will migrate to `operator/v1` in a follow-on if that is acceptable.

PR #120, opened 2019-02-01
[WIP] Add flag to handle brand

zherman0:
[WIP] This PR adds a runtime flag option to the operator that allows the user to set the brand value.  The command would look like:
`IMAGE=docker.io/openshift/origin-console:latest  console operator  --brand ocp`

### DIFF
--- a/pkg/apis/console/v1alpha1/types.go
+++ b/pkg/apis/console/v1alpha1/types.go
@@ -43,3 +43,8 @@ type ConsoleList struct {
 
 	Items []Console `json:"items"`
 }
+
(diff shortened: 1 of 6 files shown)


zherman0:
@spadgett @benjaminapetersen - Hey guys, this is a runtime solution for passing the brand.  It does not use GOFLAGS since Cobra is already being used.  This does not really solve having separate docker files unless we update the cmd args to include the brand flag but since they are commented out, I still need to think of an better overall solution.

zherman0:
Closing this for now since it is runtime based and we want a build time solution.

PR #121, opened 2019-02-01
bump *

benjaminapetersen:
Follow-up bump.
There is no explicit version change here, but we take `openshift/client-go` and `openshift/api` @ `master'. We need to pick up the following changes:

- https://github.com/openshift/client-go/pull/82
- https://github.com/openshift/api/pull/190

These two PRs merged after the previous version bump.


jhadvig:
quite a lot of changes if I take in mind that one bump was already done yesterday

benjaminapetersen:
@jhadvig I know right?! Five minutes after your bump, 10,000 lines of code change.

PR #103, opened 2019-01-11
[WIP] Updating v1alpha1/Console to v1/Console to fix cluster operator status

benjaminapetersen:

PR #124, opened 2019-02-04
Moving to use operator v1 api

jhadvig:
Migrating from `v1alpha1` towards `v1` for console-operator

Wasn't really sure what to do with `v1alpha1` fields that we have been using in the codebase. Adding comments to each occurrence, to discuss if that particular part is necessary.

@benjaminapetersen PTAL

@enj fyi


pkg/console/operator/operator.go
```diff
@@ -191,15 +186,8 @@ func (c *consoleOperator) Sync(obj metav1.Object) error {
 	case v311_to_401.BetweenOrEmpty(currentActualVersion):
 		outConfig, configChanged, err = sync_v400(c, outConfig)
 		errs = append(errs, err)
-		if err == nil {
```
jhadvig: here I wasnt really sure what to do with setting these status fields, which are not in `v1`, so im nuking them

pkg/console/operator/operator.go
```diff
@@ -191,15 +186,8 @@ func (c *consoleOperator) Sync(obj metav1.Object) error {
...
-			outConfig.Status.TaskSummary = "sync-4.0.0"
-			outConfig.Status.CurrentAvailability = &operatorsv1alpha1.VersionAvailability{
-				Version: desiredVersion.String(),
-			}
-		}
 	default:
-		logrus.Printf("unrecognized version. desired %s, actual %s", desiredVersion, currentActualVersion)
```
jhadvig: same as the previous comment

pkg/console/operator/operator.go
```diff
@@ -237,19 +225,17 @@ func (c *consoleOperator) deleteAllResources(cr *consolev1alpha1.Console) error
...
 			Namespace: api.OpenShiftConsoleNamespace,
 		},
-		Spec: consolev1alpha1.ConsoleSpec{
-			OperatorSpec: operatorsv1alpha1.OperatorSpec{
+		Spec: consolev1.ConsoleSpec{
+			OperatorSpec: operatorsv1.OperatorSpec{
 				// by default the console is managed
-				ManagementState: operatorsv1alpha1.Managed,
-				// if Version is not 4.0.0 our reconcile loop will not pick it up
-				Version: "4.0.0",
```
jhadvig: This field is not present in `v1` so removing as well 💣

pkg/console/operator/operator.go
benjaminapetersen: We can trim this down significantly.  I did it once before:

```golang

func (c *consoleOperator) Sync(obj metav1.Object) error {
	startTime := time.Now()
	logrus.Infof("started syncing operator %q (%v)", obj.GetName(), startTime)

	operatorConfig := obj.(*consolev1.Console)

	if err := c.handleSync(operatorConfig); err != nil {
		return err
	}
	return nil
}

func (c *consoleOperator) handleSync(config *consolev1.Console) error {

	switch config.Spec.ManagementState {
	case operatorv1.Managed:
		// handled below
	case operatorv1.Unmanaged:
		return nil
		// take a look @ https://github.com/openshift/service-serving-cert-signer/blob/master/pkg/operator/operator.go#L86
	case operatorv1.Removed:
		logrus.Println("console has been removed.")
		return c.deleteAllResources(config)
	default:
		// TODO should update status
		return fmt.Errorf("unknown state: %v", config.Spec.ManagementState)
	}

	// do we need the if(configChanged) update bits?
	outConfig, configChanged, err := sync_v400(c, config)
	if err != nil {
		return err
	}
	if configChanged {
		// TODO: this should do better apply logic or similar, maybe use SetStatusFromAvailability
		_, err = c.operatorClient.Update(outConfig)
		if err != nil {
			return err
		}
	}

	return nil
}

```

benjaminapetersen: See here, a previous pass at moving to the `v1`, before things changed: https://github.com/openshift/console-operator/pull/103/files#diff-69ecbb024dab972c8729678abdf9c0faR132

jhadvig:
Need to close this in favor of https://github.com/openshift/console-operator/pull/125

PR #125, opened 2019-02-05
Moving to use operator v1 api

jhadvig:
Migrating from `v1alpha1` towards `v1` for console-operator

Wasn't really sure what to do with `v1alpha1` fields that we have been using in the codebase. Adding comments to each occurrence, to discuss if that particular part is necessary.

@benjaminapetersen PTAL

@enj fyi


pkg/console/operator/operator.go
```diff
@@ -135,84 +133,52 @@ func (c *consoleOperator) Key() (metav1.Object, error) {
...
-// sync() is the handler() function equivalent from the sdk
-// this is the big switch statement.
-// TODO: clean this up a bit, its messy
 func (c *consoleOperator) Sync(obj metav1.Object) error {
 	startTime := time.Now()
 	logrus.Infof("started syncing operator %q (%v)", obj.GetName(), startTime)
 	defer logrus.Infof("finished syncing operator %q (%v) \n\n", obj.GetName(), time.Since(startTime))
 
-	operatorConfig := obj.(*consolev1alpha1.Console)
+	operatorConfig := obj.(*consolev1.Console)
```
benjaminapetersen: Pretty sure if you swap `Sync()` func here for:


```golang

func (c *consoleOperator) Sync(obj metav1.Object) error {
	startTime := time.Now()
	logrus.Infof("started syncing operator %q (%v)", obj.GetName(), startTime)

	operatorConfig := obj.(*consolev1.Console)

	if err := c.handleSync(operatorConfig); err != nil {
		return err
	}
	return nil
}

func (c *consoleOperator) handleSync(config *consolev1.Console) error {

	switch config.Spec.ManagementState {
	case operatorv1.Managed:
		// handled below
	case operatorv1.Unmanaged:
		return nil
		// take a look @ https://github.com/openshift/service-serving-cert-signer/blob/master/pkg/operator/operator.go#L86
	case operatorv1.Removed:
		logrus.Println("console has been removed.")
		return c.deleteAllResources(config)
	default:
		// TODO should update status
		return fmt.Errorf("unknown state: %v", config.Spec.ManagementState)
	}

	// do we need the if(configChanged) update bits?
	outConfig, configChanged, err := sync_v400(c, config)
	if err != nil {
		return err
	}
	if configChanged {
		// TODO: this should do better apply logic or similar, maybe use SetStatusFromAvailability
		_, err = c.operatorClient.Update(outConfig)
		if err != nil {
			return err
		}
	}

	return nil
}

```
It will eliminate the fields that we no longer care about.

benjaminapetersen: See here, a previous pass at moving to the `v1`, before things changed: https://github.com/openshift/console-operator/pull/103/files#diff-69ecbb024dab972c8729678abdf9c0faR132

benjaminapetersen: It drops the second switch statement entirely.  We dont need it yet.

pkg/console/operator/operator.go
```diff
@@ -135,84 +133,52 @@ func (c *consoleOperator) Key() (metav1.Object, error) {
...
+	case operatorsv1.Managed:
 		logrus.Println("console is in a managed state.")
 		// handled below
-	case operatorsv1alpha1.Unmanaged:
+	case operatorsv1.Unmanaged:
 		logrus.Println("console is in an unmanaged state.")
 		return nil
-	// take a look @ https://github.com/openshift/service-serving-cert-signer/blob/master/pkg/operator/operator.go#L86
-	case operatorsv1alpha1.Removed:
+		// take a look @ https://github.com/openshift/service-serving-cert-signer/blob/master/pkg/operator/operator.go#L86
```
benjaminapetersen: We can probably cut some of these old comments.

benjaminapetersen:
Fun fun...

benjaminapetersen:
500 errors

benjaminapetersen:
I was hopeful we could sneak one by CI 😅

spadgett:
looks like this needs to be rebased :/

benjaminapetersen:
boo.

jhadvig:
@spadgett rebased and pushed 👍

pkg/console/subresource/secret/secret_test.go
```diff
@@ -63,12 +64,20 @@ func TestStub(t *testing.T) {
 		{
 			name: "Test stubbing secret",
 			want: &corev1.Secret{
+<<<<<<< HEAD
```
benjaminapetersen: Wups.  Rebase stuff.....

PR #126, opened 2019-02-05
Add Goflag option to define branding at build time

zherman0:
Add Goflag option to define branding at build time.  Docker rhel file has been updated to provide ocp branding.
Usage: ADDITIONAL_GOTAGS="ocp" make WHAT="cmd/console" 
$ console version
ConsoleOperator v0.0.1
Git Commit: 6cf43775
Build Date: 2019-02-05T20:08:03Z
Current Brand Setting: ocp
Tag option: ocp
This solution uses tags in the build process.

Additionally, updated the console version helper tool to show current Git Commit value, build date, and the brand setting.

### DIFF
--- a/Dockerfile.rhel7
+++ b/Dockerfile.rhel7
@@ -1,7 +1,7 @@
 FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.10 AS builder
 WORKDIR /go/src/github.com/openshift/console-operator
 COPY . .
-RUN make build
+RUN ADDITIONAL_GOTAGS="ocp" make build WHAT="cmd/console"
 
 FROM registry.svc.ci.openshift.org/ocp/4.0:base
 RUN useradd console-operator
--- a/hack/lib/build/binaries.sh
+++ b/hack/lib/build/binaries.sh
@@ -241,7 +241,7 @@ os::build::internal::build_binaries() {
       if [[ ${#nonstatics[@]} -gt 0 ]]; then
         GOOS=${platform%/*} GOARCH=${platform##*/} go install \
           -pkgdir "${pkgdir}/${platform}" \
-          -tags "${OS_GOFLAGS_TAGS-} ${!platform_gotags_envvar:-}" \
+          -tags "${OS_GOFLAGS_TAGS-} ${!platform_gotags_envvar:-} ${ADDITIONAL_GOTAGS-}" \
           -ldflags="${local_ldflags}" \
           "${goflags[@]:+${goflags[@]}}" \
           "${nonstatics[@]}"
--- a/pkg/cmd/version/cmd.go
+++ b/pkg/cmd/version/cmd.go
@@ -3,14 +3,21 @@ package version
 import (
 	"fmt"
 	"github.com/blang/semver"
+	"github.com/openshift/console-operator/pkg/console/version"
 	"github.com/spf13/cobra"
 	"strings"
+
+	cm "github.com/openshift/console-operator/pkg/console/subresource/configmap"
 )
 
 var (
-	Raw     = "v0.0.1"
-	Version = semver.MustParse(strings.TrimLeft(Raw, "v"))
-	String  = fmt.Sprintf("ConsoleOperator %s", Raw)
+	Raw        = "v0.0.1"
+	VerInfo    = version.Get()
+	GitCommit  = VerInfo.GitCommit
+	BuildDate  = VerInfo.BuildDate
+	Version    = semver.MustParse(strings.TrimLeft(Raw, "v"))
+	BrandValue = cm.DEFAULT_BRAND
+	String     = fmt.Sprintf("ConsoleOperator %s\nGit Commit: %s\nBuild Date: %s\nCurrent Brand Setting: %s", Raw, GitCommit, BuildDate, BrandValue)
 )
 
 func NewVersion() *cobra.Command {
--- a/pkg/console/subresource/configmap/brand_ocp.go
+++ b/pkg/console/subresource/configmap/brand_ocp.go
@@ -1,0 +1,8 @@
+// +build ocp
+
+package configmap
+
+const (
+	DEFAULT_BRAND   = "ocp"
+	DEFAULT_DOC_URL = "https://docs.openshift.com/container-platform/4.0/welcome/index.html"
+)
--- a/pkg/console/subresource/configmap/brand_okd.go
+++ b/pkg/console/subresource/configmap/brand_okd.go
@@ -1,0 +1,8 @@
+// +build !ocp
+
+package configmap
+
+const (
+	DEFAULT_BRAND   = "okd"
+	DEFAULT_DOC_URL = "https://docs.okd.io/4.0/"
+)
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@ OUT_DIR = _output
 OS_OUTPUT_GOPATH ?= 1
 
 export GOFLAGS
+export ADDITIONAL_GOTAGS
 export TESTFLAGS
 # If set to 1, create an isolated GOPATH inside _output using symlinks to avoid
 # other packages being accidentally included. Defaults to on.
--- a/hack/lib/constants.sh
+++ b/hack/lib/constants.sh
@@ -54,11 +54,11 @@ function os::build::ldflags() {
 
   declare -a ldflags=()
 
-  ldflags+=($(os::build::ldflag "${OS_GO_PACKAGE}/pkg/version.majorFromGit" "${OS_GIT_MAJOR}"))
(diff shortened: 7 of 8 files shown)


zherman0:
@spadgett @benjaminapetersen - Created a build time solution using the goflags var and using tags.

Dockerfile
```diff
@@ -2,7 +2,8 @@ FROM registry.svc.ci.openshift.org/openshift/release:golang-1.10 AS builder
 # create a work dir and copy local files in
 WORKDIR /go/src/github.com/openshift/console-operator
 COPY . .
-RUN make build
+
+RUN make build WHAT="cmd/console" GOFLAGS="-brandokd"
```
spadgett: I would just have one `-ocp` flag, then fallback to OKD logo when that flag is not present. I don't think we need tags for every brand value.

zherman0: Ok, so when it comes to building the operator, we only want okd or ocp.  The other options will never be available through the build?

spadgett: Right, and I probably would only have one tag, `ocp`. Then we use OKD branding when `ocp` is not set.

zherman0:
@spadgett @benjaminapetersen - Simplified the flag situation to only have one -ocp flag.  The docker rhel file uses that flag otherwise everything defaults to okd.

hack/lib/build/binaries.sh
```diff
@@ -182,9 +182,15 @@ os::build::internal::build_binaries() {
     # Use eval to preserve embedded quoted strings.
     eval "goflags=(${OS_GOFLAGS:-})"
 
+    local brand=""
     local arg
+
     for arg; do
-      if [[ "${arg}" == -* ]]; then
+      if [[ "${arg}" == "-ocp" ]]
```
spadgett: Instead of using GOFLAGS and checking for a specific value, it would be better to define an entirely different env var for tags. Then we don't need the magic check here.

We should confirm that there isn't a way to set this already (OS_GOFLAGS_TAGS?)

zherman0: I did try a version where I had a new GOFLAG_TAGS.  I still had do checking on it and I had to add new code to the Makefile to handle another env flag var. It turned out to be easier to just use the existing GOFLAG and check it for the -ocp value.  I am open to discussion, but I thought it isolated changes to just the binaries.sh script.  
As for the OS_GOFLAG_TAGS (which is set automatically), if you mean forcing the brand based on os/arch I am worried that someone might want to have okd on RHEL. Or did you have something else in mind @spadgett ?

zherman0: That is the solution I tried at first, but the problem is the values in in the ADDITIONAL_GOTAGS get processed by the same code as goflags, and if there is no "-", the value is  considered a package to get built. So `make WHAT="cmd/console" GOFLAGS="" ADDITIONAL_GOTAGS="ocp"` gives an error : can't load package: package github.com/openshift/console-operator/ocp. I need to have the code above ignore items from the ADDITIONAL_GOTAGS but I was not sure of an easy way to do that.

spadgett: Ok I see the problem. Yeah, my suggestion is wrong. Let’s catch up on it tomorrow

spadgett: Does this work in the Dockerfile (without the Makefile changes)?

```
RUN ADDITIONAL_GOTAGS="-ocp" make build
```

zherman0: No, without the hack/build-go.sh $(WHAT) $(GOFLAGS) $(ADDITIONAL_GOTAGS) in the Makefile, that flag is ignored

benjaminapetersen: Also, does anyone know if its common to update scripts in `hack/lib` or at that point is the script typically copied out of `hack/lib` into `hack/` (top level) and tweaked as non-lib?  I'm guessing there isn't a best practice here (though all repos more or less use this.... and `lib` implies it should be generic).

pkg/cmd/version/cmd.go
```diff
@@ -3,14 +3,21 @@ package version
...
 	"strings"
+
+	cm "github.com/openshift/console-operator/pkg/console/subresource/configmap"
 )
 
 var (
-	Raw     = "v0.0.1"
-	Version = semver.MustParse(strings.TrimLeft(Raw, "v"))
-	String  = fmt.Sprintf("ConsoleOperator %s", Raw)
+	Raw        = "v0.0.1"
```
benjaminapetersen: Thanks for updating this!

I don't remember why there is a separate `pkg/console/version/version.go`.  Could be a relic of past versions of the operator.

hack/lib/build/binaries.sh
zherman0: Yeah, I did not want to update those files (hack/lib/build/binaries.sh) but there was no obvious mechanism to get additional flags into the tag field.

spadgett: @zherman0 It works for me. I did export ADDITIONAL_GOTAGS from the Makefile

```
~/go/src/github.com/openshift/console-operator goflags*
❯ ADDITIONAL_GOTAGS=ocp make
hack/build-go.sh
++ Building go targets for darwin/amd64: cmd/console
[INFO] hack/build-go.sh exited with code 0 after 00h 00m 05s

~/go/src/github.com/openshift/console-operator goflags*
❯ ./_output/local/bin/darwin/amd64/console version
ConsoleOperator v0.0.1
Git Commit: 275c5a58
Build Date: 2019-02-06T20:40:28Z
Current Brand Setting: ocp
```

spadgett: I opened a PR with the change: https://github.com/zherman0/console-operator/pull/1

benjaminapetersen: To answer on the original `hack/` scripts question, there does not seem to be a goal to keep them consistent across repos.  There are hard-coded bits that have to be changed for each repo.

Dockerfile.rhel7
```diff
@@ -1,7 +1,7 @@
 FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.10 AS builder
 WORKDIR /go/src/github.com/openshift/console-operator
 COPY . .
-RUN make build
+RUN ADDITIONAL_GOTAGS="ocp" make build WHAT="cmd/console"
```
benjaminapetersen: So the `Dockerfile` does not need to be updated due to the `build tag` below `+build !ocp` that will include `brand_okd.go` by default.

pkg/console/subresource/configmap/brand_ocp.go
```diff
@@ -0,0 +1,7 @@
+// +build ocp
+
+package configmap
+
+const (
+	BRAND = "ocp"
```
benjaminapetersen: for ref, this `BRAND` will be the fallback for `getBrand()` from:
https://github.com/openshift/console-operator/pull/134/files#diff-bc08d2d514ac5156db922f45487af976R46

pkg/console/subresource/configmap/brand_okd.go
```diff
@@ -0,0 +1,7 @@
+// +build !ocp
+
+package configmap
+
+const (
+	BRAND = "okd"
```
benjaminapetersen: [Same comment](https://github.com/openshift/console-operator/pull/126/files#r256120278)

Dockerfile.rhel7
zherman0: correct

zherman0:
flake
/test e2e-aws

benjaminapetersen:
@zherman0 this will need a rebase.  There is a small conflict now, its likely just updating the `getBrand()` func, to return your new `BRAND` const instead of `defaultBrand`.

pkg/console/subresource/configmap/brand_ocp.go
benjaminapetersen: Are you sure this is resolved?  I would have expected to see an update to `configmap.go` in this PR.

benjaminapetersen: Got me curious about the default docs URL as well... if that should be toggled for `ocp` vs `okd`.  Currently in `configmap.go` its just `"https://docs.okd.io/4.0/"`.

benjaminapetersen: @ahardin-rh helped dig up this alternative URL: https://docs.openshift.com/container-platform/4.0/welcome/index.html

I don't know yet if we should use or not (it is also password protected).  Fishing around for this.

benjaminapetersen: @zherman0 yes, I believe this is the correct URL to use. It is password protected for now as it is under active dev. It will not be later.

benjaminapetersen:
bam, merged.

PR #127, opened 2019-02-06
honor console and operator config

benjaminapetersen:
WIP
- add clients for `console.config` & `console.operator`
- honor both configurations
   - console.config is user/customer facing
   - console.operator is internal
- create default `console.config` & `console.operator` if not present
- honor values of `console.config` & `console.operator`

PR #128, opened 2019-02-06
Update CRD for operator.openshift.io Console (Operator Config)

benjaminapetersen:
- Partial from #127, breaking this up again into units.
- Related to [cluster-config-operator PR 11](https://github.com/openshift/cluster-config-operator/pull/11).
- CRD for operator config stays here, top level console config goes to cluster-config-operator
   - We don't own this config.  Its top level, so multiple operators may need to consume it.
   - We do still own our operator config.

### DIFF
--- a/manifests/00-crd-operator-config.yaml
+++ b/manifests/00-crd-operator-config.yaml
@@ -1,13 +1,18 @@
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
-  name: consoles.console.openshift.io
+  name: consoles.operator.openshift.io
 spec:
-  group: console.openshift.io
+  scope: Cluster
+  group: operator.openshift.io
   names:
     kind: Console
     listKind: ConsoleList
     plural: consoles
     singular: console
-  scope: Namespaced
-  version: v1alpha1
+  subresources:
+    status: {}
+  versions:
+  - name: v1
+    served: true
+    storage: true


manifests/00-crd-operator-config.yaml
```diff
@@ -0,0 +1,23 @@
+# operator.openshift.io Console is the private configuration for the
+# console operator.
+#
+# config.openshift.io Console is the public (user facing) configuration
+# for the console.  See this CRD
```
jhadvig: Lets swap these two comments and if https://github.com/openshift/cluster-config-operator/pull/11 got merged lets add link for the CRD :)

benjaminapetersen: This is confusing enough, I'll delete the comment here, but consider putting something in README or somewhere as a follow-up.  It should be clear how these work.

benjaminapetersen:
@enj for review

spadgett:
Won't this break the operator unless we change the corresponding operator code? This removes the old CRD, correct?

spadgett:
Yeah, this causes the operator to fail. We can't break the work up like this unfortunately.

> E0206 21:52:58.753243       1 reflector.go:134] github.com/openshift/console-operator/pkg/generated/informers/externalversions/factory.go:101: Failed to list *v1alpha1.Console: the server could not find the requested resource (get consoles.console.openshift.io)

benjaminapetersen:
Was trying to decide if I just add it along with the original CRD, then remove the original later, but not necessary.  #129 opened to replace.

PR #129, opened 2019-02-07
Console Config & Operator Config client additions

benjaminapetersen:
- Updatest mainfests/crd.yaml to crd-operator-config.yaml
- Adds clients, informers for both ConsoleConfig & ConsoleOperatorConfig
- Creates default ConsoleConfig, ConsoleOperatorConfig if either are not present 
- disables code-gen as we no longer maintain api resources in this repo

This PR clears the way for:
- Status
  - ConsoleConfig.Status.PublicHostname 
  - ConsoleOperatorConfig.Status (standard operator)
- Spec
  - ConsoleConfig.Spec.Authentication.LogoutRedirect
  - ConsoleOperatorConfig.Spec.Customization.Brand
  - ConsoleOperatorConfig.Spec.Customization.DocumentationBaseURL

Which can be done as follow-ons.

PR #130, opened 2019-02-08
Console Config & Operator Config client additions

benjaminapetersen:
Same as #129! With new and improve release payloads! (hopefully)

- Updatest mainfests/crd.yaml to crd-operator-config.yaml
- Adds clients, informers for both ConsoleConfig & ConsoleOperatorConfig
- Creates default ConsoleConfig, ConsoleOperatorConfig if either are not present 
- disables code-gen as we no longer maintain api resources in this repo

This PR clears the way for:
- Status
  - ConsoleConfig.Status.PublicHostname 
  - ConsoleOperatorConfig.Status (standard operator)
- Spec
  - ConsoleConfig.Spec.Authentication.LogoutRedirect
  - ConsoleOperatorConfig.Spec.Customization.Brand
  - ConsoleOperatorConfig.Spec.Customization.DocumentationBaseURL

Which can be done as follow-ons.


spadgett:
I'm still not seeing the config-operator console CRD in the release payload :(

hack/test-unit.sh
```diff
@@ -5,7 +5,6 @@ os::log::info "Running unit tests"
 
 
 PACKAGES_TO_TEST=(
-    "github.com/openshift/console-operator/pkg/apis/console/v1"
```
spadgett: It looks like the tests below might be fixed. Is this change still needed?

pkg/console/operator/operator.go
```diff
@@ -138,15 +135,23 @@ func (c *consoleOperator) Sync(obj metav1.Object) error {
...
 	defer logrus.Infof("finished syncing operator %q (%v) \n\n", obj.GetName(), time.Since(startTime))
 
-	operatorConfig := obj.(*consolev1.Console)
-	if err := c.handleSync(operatorConfig); err != nil {
+	operatorConfig := obj.(*operatorsv1.Console)
+
+	consoleConfig, err := c.consoleConfigClient.Get(api.ResourceName, metav1.GetOptions{})
+	if errors.IsNotFound(err) && CreateDefaultConsoleFlag {
+		logrus.Errorf("No console config found. Creating.")
+		c.consoleConfigClient.Create(c.defaultConsoleConfig())
```
spadgett: We need to at least log the error if the create fails. Otherwise we will have no way to debug failures creating the console config.

spadgett:
CRD manifest is still missing from the release payload unfortunately

spadgett:
The manifest is there!

/retest

hack/test-unit.sh
benjaminapetersen: Yup, this package no longer exists.

spadgett: Right! I see now, thanks

pkg/console/operator/operator.go
```diff
@@ -124,10 +121,12 @@ func NewConsoleOperator(
 
 // key is actually the pivot point for the operator, which is our Console custom resource
 func (c *consoleOperator) Key() (metav1.Object, error) {
-	operatorConfig, err := c.operatorClient.Get(api.ResourceName, metav1.GetOptions{})
-
+	operatorConfig, err := c.operatorConfigClient.Get(api.ResourceName, metav1.GetOptions{})
 	if errors.IsNotFound(err) && CreateDefaultConsoleFlag {
-		return c.operatorClient.Create(c.defaultConsole())
+		if _, err := c.operatorConfigClient.Create(c.defaultConsoleOperatorConfig()); err != nil {
+			logrus.Errorf("No console operator config found. Creating. %v \n", err)
```
benjaminapetersen: @spadgett agree, this is probably better.

pkg/console/operator/operator.go
```diff
@@ -138,15 +137,24 @@ func (c *consoleOperator) Sync(obj metav1.Object) error {
...
 	defer logrus.Infof("finished syncing operator %q (%v) \n\n", obj.GetName(), time.Since(startTime))
 
-	operatorConfig := obj.(*consolev1.Console)
-	if err := c.handleSync(operatorConfig); err != nil {
+	operatorConfig := obj.(*operatorsv1.Console)
+
+	consoleConfig, err := c.consoleConfigClient.Get(api.ResourceName, metav1.GetOptions{})
+	if errors.IsNotFound(err) && CreateDefaultConsoleFlag {
+		if _, err := c.consoleConfigClient.Create(c.defaultConsoleConfig()); err != nil {
+			logrus.Errorf("No console config found. Creating. %v \n", err)
```
benjaminapetersen: log & return the error.

benjaminapetersen:
It wasnt, then it was, then it wasn't.... Very weird

PR #131, opened 2019-02-08
Operator status

jhadvig:
@benjaminapetersen @spadgett @enj PTAL

### DIFF
--- a/pkg/console/operator/operator.go
+++ b/pkg/console/operator/operator.go
@@ -6,9 +6,11 @@ import (
 	"time"
 
 	// 3rd party
+	"github.com/golang/glog"
 	"github.com/sirupsen/logrus"
 
 	// kube
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -25,6 +27,7 @@ import (
 	"github.com/openshift/console-operator/pkg/api"
 	"github.com/openshift/console-operator/pkg/boilerplate/operator"
 	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
 
 	// informers
 	configinformerv1 "github.com/openshift/client-go/config/informers/externalversions/config/v1"
@@ -47,7 +50,8 @@ import (
 )
 
 const (
-	controllerName = "Console"
+	controllerName           = "Console"
+	workloadFailingCondition = "WorkloadFailing"
 )
 
 var CreateDefaultConsoleFlag bool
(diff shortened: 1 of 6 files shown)


pkg/console/starter/starter.go
```diff
@@ -136,46 +144,31 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		recorder,
 	)
 
+	clusterOperatorStatus := status.NewClusterOperatorStatusController(
+		"console",
+		[]configv1.ObjectReference{
+			{Group: "operator.openshift.io", Resource: "Console", Name: api.OpenShiftConsoleName},
```
enj: Kind != Resource

pkg/console/operator/operator.go
```diff
@@ -145,6 +147,16 @@ func (c *consoleOperator) Sync(obj metav1.Object) error {
 	}
 
 	if err := c.handleSync(operatorConfig, consoleConfig); err != nil {
+		v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorsv1.OperatorCondition{
```
enj: You need to unset this on non-failure right?  Also in general you need to guard any status updates to prevent getting your sync loop stuck in an infinite loop.

pkg/console/operator/operator.go
```diff
@@ -157,6 +177,24 @@ func (c *consoleOperator) handleSync(operatorConfig *operatorsv1.Console, consol
 		// handled below
 	case operatorsv1.Unmanaged:
 		logrus.Println("console is in an unmanaged state.")
+		v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorsv1.OperatorCondition{
+			Type:    operatorsv1.OperatorStatusTypeAvailable,
+			Status:  operatorsv1.ConditionUnknown,
+			Reason:  "Unmanaged",
```
benjaminapetersen: Shall we make a `const` for the `Reason` `"Unmanaged"`?  To reuse rather than copy/paste.

pkg/console/operator/operator.go
```diff
@@ -157,6 +177,24 @@ func (c *consoleOperator) handleSync(operatorConfig *operatorsv1.Console, consol
 		// handled below
 	case operatorsv1.Unmanaged:
 		logrus.Println("console is in an unmanaged state.")
+		v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorsv1.OperatorCondition{
```
benjaminapetersen: Ok, so if I set the `Console` to `Unmanaged`, all of theses statuses will be applied.  When I change it back to `Mananaged`, or to `Removed` or `Force`, we will need to clear these or update them to something else.

jhadvig:
cc'ing @bparees

pkg/console/operator/operator.go
bparees: That should already be happening since you should have already had something that was setting these under steady state (ie on resync as well as on resource updates).

benjaminapetersen: We don't yet, this is the first PR to do anything with Conditions.

benjaminapetersen:
#winning all tests pass... lol.

jhadvig:
@benjaminapetersen checked both tests logs for the console-operator pod and there are without any errors! We are probably good to merge this.
Anyway will go and try the changes locally one again following your README

bparees:
@jhadvig you need to create the empty clusteroperator object as part of your manifest, otherwise the CVO does not know to wait for it.  You need to create that object *after* your you create your operator deployment.  see:
https://github.com/openshift/cluster-image-registry-operator/tree/master/manifests
https://github.com/openshift/cluster-image-registry-operator/blob/master/manifests/07-operator.yaml
and
https://github.com/openshift/cluster-image-registry-operator/blob/master/manifests/08-clusteroperator.yaml

benjaminapetersen:
@jhadvig @bparees we have that manifests here in PR #132. It's fine to copy the file over here and close, since CI has flaked and it hasn't merged yet.

Thanks for the note on ordering, so we will want to create `06-clusteroperator.yaml` to follow our `05-operator.yaml`

benjaminapetersen:
Ok, pulled & deployed.

I see `console` in the list:
```
oc get clusteroperator
NAME                                    VERSION                  AVAILABLE   PROGRESSING   FAILING   SINCE
cluster-autoscaler-operator             v4.0.0-0.150.0.0-dirty   True        False         False     1d
cluster-node-tuning-operator                                     True        False         False     1d
console                                                          True        False         True      50s
kube-controller-manager                                          True        False         False     1d
machine-api-operator                    v4.0.0-0.150.0.0-dirty   True        False         False     1m
...
```
yay.

benjaminapetersen:
Oops, we have some conflicts to resolve now.

pkg/console/operator/sync_v400.go
```diff
@@ -38,60 +40,114 @@ import (
...
 	logrus.Println("running sync loop 4.0.0")
 	recorder := co.recorder
 
 	// track changes, may trigger ripples & update operator config or console config status
 	toUpdate := false
 
-	rt, rtChanged, rtErr := SyncRoute(co, operatorConfig)
-	if rtErr != nil {
-		return operatorConfig, consoleConfig, toUpdate, rtErr
+	rt, rtChanged, err := SyncRoute(co, operatorConfig)
```
benjaminapetersen: I'm thinking this over. It may actually break things.  We've been depending on the sync loop stopping as soon as it hits one error & starting over.  This has been a good thing, as the loop will resolve 1 problem and die. A change will happen, so the loop gets triggered again.  On the next pass, it will resolve the next problem & die. It will do this over and over until all errors are resolved & everything is working.  

If we go the `append` route, then an early error doesn't stop the loop.  That means a later `Sync*` func is probably trying to fix something it may not be able to fix.

For example, if the `route` does not yet have `Spec.Host` (this is provided by the cluster), then we don't really care if it even tries to create the configmap, the data will be wrong.  And we don't really care if it tries to make the deployment, it can't run without the correct configmap.

The `return` for each of these ensures that we only have to deal with the single relevant error.

benjaminapetersen: I wonder if we can use a couple funcs for this:

```golang
        // still kill the loop...
	if rtErr != nil {
                 // helper func to deal with the status update
		handleSyncErrorCondition(co, operatorConfig, fmt.Errorf("%q: %v", "route", rtErr))
		return operatorConfig, consoleConfig, toUpdate, rtErr
	}
	toUpdate = toUpdate || rtChanged

// function defined below...
func handleSyncErrorCondition(co *consoleOperator, operatorConfig *operatorv1.Console, message string) {
	v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorv1.OperatorCondition{
		Type:               workloadFailingCondition,
		Status:             operatorv1.ConditionTrue,
		Message:            message,
		Reason:             "SyncError",
		LastTransitionTime: metav1.Now(),
	})

	if _, err := co.operatorConfigClient.UpdateStatus(operatorConfig); err != nil {
		// handle
	}
	// hanlde
}
```
Still tinkering with this.

pkg/console/operator/operator.go
```diff
@@ -47,7 +50,8 @@ import (
 )
 
 const (
-	controllerName = "Console"
+	controllerName           = "Console"
+	workloadFailingCondition = "WorkloadFailing"
```
benjaminapetersen: Interesting. I thought there was a set of `Available`, `Progressing`, `Failing`, only. I didn't realize there is also `PrereqsSatisfied` & `Upgradeable`, and that we can add custom.

jhadvig: you can customize the failing one's ... the conditiona type just needs to be suffexed with `Failing`

benjaminapetersen: Nice

pkg/console/operator/sync_v400.go
```diff
@@ -105,16 +170,27 @@ func sync_v400(co *consoleOperator, operatorConfig *operatorv1.Console, consoleC
...
-	// the cluster operator status
-	return operatorConfig, consoleConfig, toUpdate, nil
+	return nil, nil, toUpdate, nil
 }
 
-func SyncConsoleConfig(co *consoleOperator, consoleConfig *configv1.Console, route *routev1.Route) (*configv1.Console, error) {
-	logrus.Printf("Updating console.config.openshift.io with hostname: %v \n", route.Spec.Host)
-	consoleConfig.Status.PublicHostname = util.HTTPS(route.Spec.Host)
-	return co.consoleConfigClient.UpdateStatus(consoleConfig)
+func handleSyncErrorCondition(operatorConfig *operatorv1.Console, message string) {
```
benjaminapetersen: So if you call this & the `condition` is added, but the `handleSyncErrorCondition` doesn't actually do an `Update()` request, its going to lose this change.

benjaminapetersen:
So this error:

`level=fatal msg="failed to initialize the cluster: Cluster operator console has not yet reported success"`

https://github.com/openshift/cluster-version-operator/blob/9335e43c21efc9df7eed48f05585997087fc04db/pkg/cvo/internal/operatorstatus.go#L126

Means we need `Available: true, Progressing: false, Failing: false` in order for CVO to be happy & tell the installer its happy.

benjaminapetersen:
Closing to clean up open PRs for feature freeze, #142 has a few other necessary components for operator status, such as renumbering manifests, etc.

PR #132, opened 2019-02-11
Add console ClusterOperator manifest

benjaminapetersen:
In order for CVO to be aware of the `console` `ClusterOperator`, we need to have a manifest.

Uncertain about name `console` vs `openshift-console` atm.  Will find out.

@jhadvig @zherman0

### DIFF
--- a/manifests/05-clusteroperator.yaml
+++ b/manifests/05-clusteroperator.yaml
@@ -1,0 +1,5 @@
+apiVersion: config.openshift.io/v1
+kind: ClusterOperator
+metadata:
+  name: console
+spec: {}
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -22,3 +22,14 @@ const (
 	OpenShiftConsoleRouteName         = OpenShiftConsoleShortName
 	OAuthClientName                   = OpenShiftConsoleName
 )
+
+// configs always have the name "cluster"
+const (
+	ConsoleConfigName         = "cluster"
+	ConsoleOperatorConfigName = ConsoleConfigName
+)
+
+// however, the ClusterOperator resource is named "console"
+const (
+	ClusterOperatorName = "console"
+)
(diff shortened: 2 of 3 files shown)


manifests/05-clusteroperator.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: config.openshift.io/v1
+kind: ClusterOperator
+metadata:
+  name: console
```
benjaminapetersen: `console` is the correct name for this resource.

pkg/api/api.go
```diff
@@ -22,3 +22,14 @@ const (
 	OpenShiftConsoleRouteName         = OpenShiftConsoleShortName
 	OAuthClientName                   = OpenShiftConsoleName
 )
+
+// configs always have the name "cluster"
```
benjaminapetersen: `cluster` is the correct name for our configs, however. 

- `config.openshift.io`, `Console`, name: `cluster`
- `operator.openshift.io`, `Console`, name: `cluster`

manifests/05-clusteroperator.yaml
benjaminapetersen: - `config.openshift.io`, `ClusterOperator`, name: `console`

pkg/api/api.go
```diff
@@ -22,3 +22,14 @@ const (
...
+const (
+	ConsoleConfigName         = "cluster"
+	ConsoleOperatorConfigName = ConsoleConfigName
+)
+
+// however, the ClusterOperator resource is named "console"
+const (
+	ClusterOperatorName = "console"
```
benjaminapetersen: - `config.openshift.io`, `ClusterOperator`, name: `console`

benjaminapetersen: But the `namespace` for the `operator` is `openshift-console-operator`, and the `operator` `deployment` is named `console-operator`. 

Just tacking that on here for record :)

benjaminapetersen:
@jhadvig is rolling this into his PR #131 
Closing.

PR #133, opened 2019-02-11
Rename Deployment openshift-console to console

benjaminapetersen:
Per an email conversation w/@spadgett.

Changing the `Deployment` name to `console` is consistent with the other resources & eliminates the extra `ConsoleShortName` variable.

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -11,14 +11,13 @@ const (
 
 // consts to maintain existing names of various sub-resources
 const (
-	OpenShiftConsoleName              = "openshift-console"
-	OpenShiftConsoleShortName         = "console"
+	OpenShiftConsoleName              = "console"
 	OpenShiftConsoleNamespace         = TargetNamespace
 	OpenShiftConsoleOperatorNamespace = "openshift-console-operator"
 	OpenShiftConsoleOperator          = "console-operator"
 	OpenShiftConsoleConfigMapName     = "console-config"
 	OpenShiftConsoleDeploymentName    = OpenShiftConsoleName
-	OpenShiftConsoleServiceName       = OpenShiftConsoleShortName
-	OpenShiftConsoleRouteName         = OpenShiftConsoleShortName
+	OpenShiftConsoleServiceName       = OpenShiftConsoleName
+	OpenShiftConsoleRouteName         = OpenShiftConsoleName
 	OAuthClientName                   = OpenShiftConsoleName
 )
--- a/manifests/00-oauth.yaml
+++ b/manifests/00-oauth.yaml
@@ -1,6 +1,6 @@
 apiVersion: oauth.openshift.io/v1
 kind: OAuthClient
 metadata:
-  name: openshift-console
+  name: console
 grantMethod: auto
 respondWithChallenges: false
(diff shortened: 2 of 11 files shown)


benjaminapetersen:
Expecting `unit` tests to break, and guessing the output will be difficult to read again (large json blobs to manually diff).  @zherman0 fyi.

benjaminapetersen:
Tests fixed, this should pass.

benjaminapetersen:
@jhadvig @zherman0

jhadvig:
`[ERROR] hack/verify-gofmt.sh exited with code 1 after 00h 00m 01s`

benjaminapetersen:
Got it

jhadvig:
```

**Comment by jhadvig** [comment]
Still seeing the same issue:
```

benjaminapetersen:
Thinking on the lock.  That's for the operator deployment, not the console deployment, but seems a good clue.

benjaminapetersen:
Failed on waiting for `console URL`.  But the installer is looking for the `route` named `console`, so renaming the deployment should not be a problem.

https://github.com/openshift/installer/blob/6b4ba7ff0e330c94f72d95c8bd5f64cd173d79a0/cmd/openshift-install/create.go#L340

pkg/api/api.go
```diff
@@ -11,14 +11,13 @@ const (
...
 	OpenShiftConsoleNamespace         = TargetNamespace
 	OpenShiftConsoleOperatorNamespace = "openshift-console-operator"
 	OpenShiftConsoleOperator          = "console-operator"
 	OpenShiftConsoleConfigMapName     = "console-config"
 	OpenShiftConsoleDeploymentName    = OpenShiftConsoleName
-	OpenShiftConsoleServiceName       = OpenShiftConsoleShortName
-	OpenShiftConsoleRouteName         = OpenShiftConsoleShortName
+	OpenShiftConsoleServiceName       = OpenShiftConsoleName
+	OpenShiftConsoleRouteName         = OpenShiftConsoleName
 	OAuthClientName                   = OpenShiftConsoleName
```
spadgett: I don't think we can change the name of the OAuth client. It's not our operator that creates it, right?

spadgett: OK, I see it's in the manifests, but we would need to update RBAC in that case.

benjaminapetersen: Ah, thanks for pointing that out.   Swinging back around and will test this one manually.  Applying all the manifests to my cluster & deploying it now to ensure things line up properly.

benjaminapetersen:
screen shot 2019-02-13 at 11 50 25 am

Validation from the `console` that the only `deployment` running in `openshift-console` is called `console`.  😄

benjaminapetersen:
also the `oauthclient` matches:
```
oc get oauthclient
NAME                           SECRET                                        WWW-CHALLENGE   TOKEN-MAX-AGE   REDIRECT URIS
console                        <nope>                  FALSE           default         https://console-openshift-console.apps.betersen.devcluster.openshift.com
openshift-browser-client       <nope>                  FALSE           default         https://betersen-api.devcluster.openshift.com:6443/oauth/token/display,/oauth/token/display
openshift-challenging-client                                                 TRUE            default         https://betersen-api.devcluster.openshift.com:6443/oauth/token/implicit
```

benjaminapetersen:
Squashed. Let's see some tests pass & put one to rest.

benjaminapetersen:
/retest

flake

PR #139, opened 2019-02-14
Operator status revisions

benjaminapetersen:
Extracted a bunch of operator status updates out into separate named functions.  Was a bit pedantic about naming, but hopefully it helps understand the flow and see if we can tease out any bugs.

@jhadvig

PR #140, opened 2019-02-14
[DO NOT MERGE] Operator status rework

jhadvig:

PR #142, opened 2019-02-15
Operator status revisions squash

benjaminapetersen:
Based on #139 
- Squashed various fixes
- Refactored into cleaner function calls
- Renumbered manifests to ensure the clusteroperator/console is created after the operator.  This should avoid a race that may cause the clusteroperator/console to report failure status simply because the operator does not yet exist.

Some screenshots:
screen shot 2019-02-19 at 3 49 03 pm
screen shot 2019-02-19 at 11 28 34 am
screen shot 2019-02-19 at 11 28 10 am


pkg/console/operator/status.go
```diff
@@ -0,0 +1,257 @@
...
+	"fmt"
+
+	"github.com/sirupsen/logrus"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	operatorsv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
```
benjaminapetersen: Detailing out the purpose behind the status/conditions here to ensure we get it right.

benjaminapetersen:
I think tests are frozen right now on:

```
level=fatal msg="failed to initialize the cluster: Cluster operator machine-config is reporting a failure: Failed to resync 3.11.0-673-gadf12809-dirty because: Get https://172.30.0.1:443/apis/apiextensions.k8s.io/v1beta1/customresourcedefinitions/controllerconfigs.machineconfiguration.openshift.io: dial tcp 172.30.0.1:443: connect: connection refused"
```

benjaminapetersen:
`level=fatal msg="failed to initialize the cluster: Cluster operator machine-config is reporting a failure: Failed to resync 3.11.0-676-g745693cd-dirty because: error syncing: request declared a Content-Length of 483 but only wrote 0 bytes"`

benjaminapetersen:
`level=fatal msg="failed to initialize the cluster: Cluster operator network has not yet reported success"`

`level=fatal msg="failed to initialize the cluster: Cluster operator console has not yet reported success"`

Network operator failed in one test, Console in the other.  Not sure if flakes or related.

benjaminapetersen:
no `console-operator` logs at all that run.

benjaminapetersen:
`error: unable to read image registry.svc.ci.openshift.org/ci-op-8yj29n81/stable@sha256:0d6c76c0d202665f7a16b899f4c62d94c9ac8a14c7540c841a8e802a91775253: received unexpected HTTP status: 504 Gateway Time-out
`

benjaminapetersen:
/assign @zherman0 @jhadvig @spadgett 

I'll take some feedback. I think if tests finally pass at some point I'd rather not have to make changes & restart the whole process 😄

benjaminapetersen:
```
error: unable to read image registry.svc.ci.openshift.org/ci-op-ps43q36m/stable@sha256:a792c98d39e33a119c98300e4179891ffa593036213eda7fedb7625557cc0f8a: received unexpected HTTP status: 504 Gateway Time-out
```

pkg/console/operator/status.go
```diff
@@ -0,0 +1,295 @@
...
+//   group: config.openshift.io
+//   kind:  ClusterOperator
+//   name:  console
+//
+// Status Condition Types
+// Status conditions should not be set to a "default" state when the operator starts up.
+// Instead, set a status explicitly only when the state is known.  This is because
+// the various status conditions may have been set on previous runs of the sync loop, or
+// possibly even by a previous operator container.
+//
```
benjaminapetersen: Read this comment to understand the purpose of these 3 statuses.

pkg/console/operator/status.go
```diff
@@ -0,0 +1,295 @@
...
+//
+// Status Condition Reason & Message
+// Reason:  OperatorSyncLoopError
+// Message: "The operator sync loop was not completed successfully"
+//
+//
+
+// Lets transition to using this, and get the repetition out of all of the above.
+func (c *consoleOperator) SyncStatus(operatorConfig *operatorsv1.Console) (*operatorsv1.Console, error) {
```
benjaminapetersen: This func is the only one that hits the API

pkg/console/operator/status.go
```diff
@@ -0,0 +1,295 @@
...
+			}
+			if hasMessage {
+				buf.WriteString(fmt.Sprintf(" %s", condition.Message))
+			}
+		}
+		logrus.Println(buf.String())
+	}
+}
+
+// setStatusCondition
```
benjaminapetersen: This is slightly less verbose than doing it all inline. I've used funcs below to try to abstract away the noise even more.

manifests/08-clusteroperator.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: config.openshift.io/v1
```
benjaminapetersen: This must exist for CVO to know about `clusteroperator/console`.

pkg/console/operator/sync_v400.go
```diff
@@ -49,46 +50,78 @@ func sync_v400(co *consoleOperator, operatorConfig *operatorv1.Console, consoleC
 
 	rt, rtChanged, rtErr := SyncRoute(co, operatorConfig)
 	if rtErr != nil {
+		co.SyncStatus(co.ConditionResourceSyncFailure(operatorConfig, fmt.Sprintf("%v: %s\n", "route", rtErr)))
```
benjaminapetersen: Here is where all the one-liners are used to try to make this readable.

pkg/console/operator/sync_v400.go
```diff
@@ -49,46 +50,78 @@ func sync_v400(co *consoleOperator, operatorConfig *operatorv1.Console, consoleC
...
+	}
+	// the operand is available if all resources are present & if we have pods
+	if actualDeployment.Status.ReadyReplicas > 0 {
+		co.ConditionDeploymentAvailable(operatorConfig)
+	} else {
+		co.ConditionDeploymentNotAvailable(operatorConfig)
+	}
+
+	// finally write out the set of conditions currently set
+	co.SyncStatus(operatorConfig)
```
benjaminapetersen: On the happy path, this is where we hit the API.  Otherwise, `SyncStatus` is always called right before an error is returned.

benjaminapetersen:
`level=fatal msg="failed to initialize the cluster: Cluster operator network has not yet reported success"
`

/retest

benjaminapetersen:
/retest

`level=fatal msg="failed to initialize the cluster: Cluster operator network has not yet reported success"`

pkg/console/operator/status.go
```diff
@@ -0,0 +1,295 @@
...
+	return operatorConfig
+}
+
+// sets multiple conditions
+func (c *consoleOperator) ConditionsManagementStateRemoved(operatorConfig *operatorsv1.Console) *operatorsv1.Console {
+	reason := "ManagementStateRemoved"
+	message := "The console has been removed."
+	v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorsv1.OperatorCondition{
+		Type:               operatorsv1.OperatorStatusTypeAvailable,
+		Status:             operatorsv1.ConditionFalse,
```
spadgett: We might need to double-check whether this would block an upgrade.

pkg/console/operator/status.go
spadgett: I feel like all the helpers might actually be adding code without really making things simpler, but I won't ask you to change it :)

pkg/console/operator/status.go
```diff
@@ -0,0 +1,295 @@
...
+
+	return operatorConfig
+}
+
+// sets multiple conditions
+func (c *consoleOperator) ConditionsManagementStateUnmanaged(operatorConfig *operatorsv1.Console) *operatorsv1.Console {
+	v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorsv1.OperatorCondition{
+		Type:               operatorsv1.OperatorStatusTypeAvailable,
+		Status:             operatorsv1.ConditionUnknown,
+		Reason:             "ManagementStateUnmanaged",
```
spadgett: worth a const since this is repeated

pkg/console/operatorclient/operatorclient.go
```diff
@@ -0,0 +1,61 @@
...
+type OperatorClient struct {
+	Informers operatorv1informers.SharedInformerFactory
+	Client    operatorv1client.ConsolesGetter
+}
+
+func (c *OperatorClient) Informer() cache.SharedIndexInformer {
+	return c.Informers.Operator().V1().Consoles().Informer()
+}
+
+func (c *OperatorClient) GetOperatorState() (*operatorv1.OperatorSpec, *operatorv1.OperatorStatus, string, error) {
```
spadgett: I might have missed it, but I don't see where these functions are used

pkg/console/operator/status.go
```diff
@@ -0,0 +1,295 @@
...
+// Message: "The operator sync loop was not completed successfully"
+//
+//
+
+// Lets transition to using this, and get the repetition out of all of the above.
+func (c *consoleOperator) SyncStatus(operatorConfig *operatorsv1.Console) (*operatorsv1.Console, error) {
+	logConditions(operatorConfig.Status.Conditions)
+	updatedConfig, err := c.operatorConfigClient.UpdateStatus(operatorConfig)
+	if err != nil {
+		return nil, fmt.Errorf("status update error: %v \n", err)
```
spadgett: It looks like these errors will never get logged if `SyncStatus` fails

pkg/console/operator/status.go
benjaminapetersen: Prefer going back to inline?  See PR #140. We could make the manifest updates, etc there

pkg/console/operator/status.go
benjaminapetersen: I wonder if unmanaged operators ought to block an upgrade...

benjaminapetersen: Oh removed. Hmm.

pkg/console/starter/starter.go
```diff
@@ -136,46 +144,31 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		recorder,
 	)
 
+	clusterOperatorStatus := status.NewClusterOperatorStatusController(
+		"console",
+		[]configv1.ObjectReference{
+			{Group: "operator.openshift.io", Resource: "consoles", Name: api.OpenShiftConsoleName},
+			{Group: "config.openshift.io", Resource: "consoles", Name: api.OpenShiftConsoleName},
```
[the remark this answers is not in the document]
benjaminapetersen: Since PR #143 absolutely.

pkg/console/operatorclient/operatorclient.go
benjaminapetersen: I believe this is a file that is borrowed (ie, copy/paste) from other operators. I think they all impl basically the same methods, with minor tweaks to impl.  This is probably a good candidate for a resource that may be in `client-go` in the future, is my guess, at least the `interface`.

pkg/console/operator/status.go
benjaminapetersen: The first thing this func does is call `logConditions`, which will output all the conditions in the logs. Then it will try to make an API call, and error out if something fails.

spadgett: Right, but that error is never logged. We return an error from this function, but none of the callers handle it. So we'd never know that status isn't updated.

pkg/console/operator/status.go
```diff
@@ -0,0 +1,299 @@
...
+		Message:            "The operator is in an unmanaged state, therefore no operator actions are failing.",
+		LastTransitionTime: metav1.Now(),
+	})
+
+	return operatorConfig
+}
+
+func (c *consoleOperator) ConditionsManagementStateRemoved(operatorConfig *operatorsv1.Console) *operatorsv1.Console {
+	v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorsv1.OperatorCondition{
+		Type: operatorsv1.OperatorStatusTypeAvailable,
```
benjaminapetersen: @spadgett setting `Available:true` here, per your comment about upgrades, in reference to the discussion in `openshift/api`, which I've commented on as it may be in flux.

benjaminapetersen:
/retest

`level=fatal msg="failed to initialize the cluster: Cluster operator network has not yet reported success"`

spadgett:
@benjaminapetersen needs rebase

manifests/08-clusteroperator.yaml
```diff
@@ -0,0 +1,5 @@
...
+kind: ClusterOperator
+metadata:
```
benjaminapetersen: Fyi, the correct name is `console`, though the `console.config.openshift.io` resource is `cluster` as is `console.operator.openshift.io`.

pkg/console/starter/starter.go
```diff
@@ -140,46 +148,30 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		recorder,
 	)
 
+	clusterOperatorStatus := status.NewClusterOperatorStatusController(
+		"console",
+		[]configv1.ObjectReference{
+			{Group: "operator.openshift.io", Resource: "consoles", Name: api.ConfigResourceName},
```
benjaminapetersen: Both configs now named `cluster`.

pkg/console/starter/starter.go
```diff
@@ -140,46 +148,30 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
...
+			{Group: "config.openshift.io", Resource: "consoles", Name: api.ConfigResourceName},
+			{Group: "oauth.openshift.io", Resource: "oauthclients", Name: api.OAuthClientName},
+			{Resource: "namespaces", Name: api.OpenShiftConsoleOperatorNamespace},
```
benjaminapetersen: Referencing `namespaces` implies the resources in those namespaces.

benjaminapetersen:
rebased

pkg/console/operator/status.go
```diff
@@ -0,0 +1,299 @@
...
+	reasonNoPodsAvailable     = "NoPodsAvailable"
+	reasonSyncError           = "SyncError"
+)
+
+// Lets transition to using this, and get the repetition out of all of the above.
+func (c *consoleOperator) SyncStatus(operatorConfig *operatorsv1.Console) (*operatorsv1.Console, error) {
+	logConditions(operatorConfig.Status.Conditions)
+	updatedConfig, err := c.operatorConfigClient.UpdateStatus(operatorConfig)
+	if err != nil {
+		return nil, fmt.Errorf("status update error: %v \n", err)
```
spadgett: I know you marked the conversation as resolved, but I don't see how it's fixed. If the call to `UpdateStatus` on line 63 fails, where does `err` get logged? (`logConditions` logs something different.)

pkg/console/operator/status.go
```diff
@@ -0,0 +1,299 @@
...
+		logrus.Println(buf.String())
+	}
+}
+
+// setStatusCondition
+// A generic helper for setting a status condition.
+// To use when another more specific status function is not sufficient.
+// examples:
+//  setStatusCondition(operatorConfig, Failing, True, "SyncLoopError", "Sync loop failed to complete successfully")
+func (c *consoleOperator) SetStatusCondition(operatorConfig *operatorsv1.Console, conditionType string, conditionStatus operatorsv1.ConditionStatus, conditionReason string, conditionMessage string) *operatorsv1.Console {
```
spadgett: I don't see how this is better than calling `v1helpers.SetOperatorCondition` directly. It actually seems worse to me since it would be easy to mix up the order of the arguments.

spadgett: I also don't see where this is used.

pkg/console/operator/status.go
```diff
@@ -0,0 +1,299 @@
...
+	})
+
+	return operatorConfig
+}
+
+func (c *consoleOperator) ConditionNotFailing(operatorConfig *operatorsv1.Console) *operatorsv1.Console {
+	v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorsv1.OperatorCondition{
+		Type:               operatorsv1.OperatorStatusTypeFailing,
+		Status:             operatorsv1.ConditionFalse,
+		LastTransitionTime: metav1.Now(),
```
spadgett: Ideally we'd add messages to all these, but OK as a follow on.

pkg/console/operator/status.go
```diff
@@ -0,0 +1,299 @@
...
+		Status:             operatorsv1.ConditionTrue,
+		Message:            message,
+		Reason:             reasonSyncError,
+		LastTransitionTime: metav1.Now(),
+	})
+
+	return operatorConfig
+}
+
+func (c *consoleOperator) ConditionResourceSyncSuccess(operatorConfig *operatorsv1.Console) *operatorsv1.Console {
```
spadgett: How is this different than `ConditionNotFailing`?

pkg/console/operatorclient/operatorclient.go
spadgett: If they're not being used, why did we copy them in?

pkg/console/operator/status.go
benjaminapetersen: Oh, my bad, i see what you are saying.  Fixed!

pkg/console/operator/status.go
benjaminapetersen: Its not used at this point, its been factored out.
I'll remove it.

benjaminapetersen: It was a step to initially condense the inline noise:

```golang
if aBadThing {
       logrus.Errorf("Bad things are happening")
       // logic....
	v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorsv1.OperatorCondition{
		Type:               BadType,
		Status:             BadStatus,
		Reason:             "ABadThing",
		Message:            "a bad thing",
		LastTransitionTime: metav1.Now(),
	})
       // oh my if more than one....
       // do other logic
}
// to a one-liner
setStatusCondition(operatorConfig, Failing, True, "SyncLoopError", "Sync loop failed to complete successfully")
```
That said, I agree, it was still too many things to take care of.  Thats when I moved to 

```golang
// one condition or 3, still one line... not 7,14,21, etc.
co.ConditionABadThing(operatorConfig)
```

benjaminapetersen: The right way to do this is probably for:

```golang
// this to return enough info:
structuredThing, err := sync_v400()
// that we could instead handle it all in one place in operator.go
handleConditions(structuredThing, err)
```
Tech debt...

pkg/console/operator/status.go
benjaminapetersen: `Failing:False` is the desired state, along with `Progressing:False` and `Available:True`.  I assumed if in the desired state no other information should be needed, but I'm open to adding info if we feel it is necessary or helpful.

pkg/console/operator/status.go
benjaminapetersen: It is not at this point.  Initially I was considering putting the rest of [ this logic (sync_v400 line 110) ](https://github.com/openshift/console-operator/pull/142/files#diff-283b3e79dd03261002107f949ad39c38R110) in  this function as it may have been necessary to set more than one status. 

I may be willing to eliminate this wrapper as I believe it won't be a "set multiple statuses" kind of function.

pkg/console/operatorclient/operatorclient.go
benjaminapetersen: We copied the file & tweaked it like others:
https://github.com/search?q=org%3Aopenshift+OperatorClient&type=Code
There are about 73 other versions of this.

This is an interface impl, we need these `funcs` whenever we pass the operator client around.

benjaminapetersen: The 4th arg to `clusterOperatorStatus := status.NewClusterOperatorStatusController( )` line `125` of `starter.go` requires [this interface](https://github.com/openshift/library-go/blob/master/pkg/operator/v1helpers/interfaces.go#L8):

```golang
// https://github.com/openshift/library-go/blob/master/pkg/operator/v1helpers/interfaces.go#L8
type OperatorClient interface {
	Informer() cache.SharedIndexInformer
	GetOperatorState() (spec *operatorv1.OperatorSpec, status *operatorv1.OperatorStatus, resourceVersion string, err error)
	UpdateOperatorSpec(string, *operatorv1.OperatorSpec) (spec *operatorv1.OperatorSpec, resourceVersion string, err error)
	UpdateOperatorStatus(string, *operatorv1.OperatorStatus) (status *operatorv1.OperatorStatus, err error)
}
```

spadgett: Got it, thanks. Just making sure this wasn't dead code.

pkg/console/operator/status.go
```diff
@@ -0,0 +1,320 @@
...
+		logrus.Println(buf.String())
+	}
+}
+
+// setStatusCondition
+// A generic helper for setting a status condition.
+// To use when another more specific status function is not sufficient.
+// examples:
+//  setStatusCondition(operatorConfig, Failing, True, "SyncLoopError", "Sync loop failed to complete successfully")
+func (c *consoleOperator) SetStatusCondition(operatorConfig *operatorsv1.Console, conditionType string, conditionStatus operatorsv1.ConditionStatus, conditionReason string, conditionMessage string) *operatorsv1.Console {
```
spadgett: We should remove if unused, but we can do that as a follow on if this passes CI.

pkg/console/operator/sync_v400.go
```diff
@@ -49,46 +51,83 @@ func sync_v400(co *consoleOperator, operatorConfig *operatorv1.Console, consoleC
...
+	logrus.Printf("sync loop 4.0.0 resources updated: %v \n", toUpdate)
+	logrus.Println("-----------------------")
+
+	// if we made it this far, the operator is not failing
+	// but we will handle the state of the operand below
+	co.ConditionResourceSyncSuccess(operatorConfig)
+	// the operand is in a transitional state if any of the above resources changed
+	// or if we have not settled on the desired number of replicas
+	if toUpdate || actualDeployment.Status.ReadyReplicas != deploymentsub.ConsoleReplicas {
+		co.ConditionResourceSyncProgressing(operatorConfig, "Changes made during sync updates, additional sync expected.")
```
spadgett: We might revisit this message, but OK for now.

spadgett:
```
level=warning msg="Found override for ReleaseImage. Please be warned, this is not advised"
level=info msg="Consuming \"Install Config\" from target directory"
level=info msg="Creating cluster..."
level=info msg="Waiting up to 30m0s for the Kubernetes API..."
level=fatal msg="waiting for Kubernetes API: context deadline exceeded"
```

/retest

benjaminapetersen:
/retest

Died before generating any artifacts due to:
`received unexpected HTTP status: 504 Gateway Time-out`

pkg/console/operator/status.go
spadgett: >  I assumed if in the desired state no other information should be needed, but I'm open to adding info if we feel it is necessary or helpful.

I think it would make things more clear if we add a message saying things are good. Particularly because you have to think through a double negative like failing: false. The messages are displayed on the cluster settings page in the UI.

(For a follow on)

pkg/console/operator/sync_v400.go
benjaminapetersen: Agree, its not the best message.  Progressing is quite fast, so its unlikely to be seen, but still, can def revisit.

benjaminapetersen:
woohoo, one set succeeded this time...

benjaminapetersen:
Nice.

spadgett:
```
level=error msg="\t* module.vpc.aws_route_table_association.route_net[3]: 1 error occurred:"
level=error msg="\t* aws_route_table_association.route_net.3: timeout while waiting for state to become 'success' (timeout: 
```

/retest

benjaminapetersen:
Bah, looked like they passed.

spadgett:
> Bah, looked like they passed.

Yeah, they have to run again since another PR merged to master in between

spadgett:
e2e-aws passed, waiting on e2e-aws-operator

spadgett:
all green!

benjaminapetersen:
Fantastic.

PR #144, opened 2019-02-16
List configs by name explicitly

benjaminapetersen:
Lists on the configs should ensure they only grab configs explicitly named `cluster`:

- `console.config.openshift.io - cluster`
- `console.operator.openshift.io - operator`


benjaminapetersen:
/assign @jhadvig @zherman0 

fyi. we need these to conform with all configs, all top level & operator configs are named `cluster`.  At present, our listers work, but we want to ensure they ignore configs created with alternate names.

benjaminapetersen:
/retest

Machine configs issue

PR #145, opened 2019-02-18
Public hostname should be the hostname, not URL

spadgett:
/assign @benjaminapetersen @enj


benjaminapetersen:
Will we want to change this back after https://github.com/openshift/api/pull/212

spadgett:
> Will we want to change this back after [openshift/api#212](https://github.com/openshift/api/pull/212)

Yeah, but I am putting this fix in for now in case there's disagreement on the API or the API changes don't make it in.

PR #147, opened 2019-02-19
Set console-operator replicas to 1

benjaminapetersen:
The director for # of replicas for an operator seems to be:

- Other operators depend on you: `replicas: 2`
- Other operators do not depend on you: `replicas: 1`

If the operator goes down & does not block another operator from recovering, a higher number of replicas is not needed 

- `thing you block, if len() > 0 { replicas = 2 } else { replicas = 1 }`. 
- Also `if have web facing traffic replicas >= 2 else replicas 1`.

I almost updated leader election to:

```
apiVersion: operator.openshift.io/v1alpha1
kind: GenericOperatorConfig
leaderElection:
  namespace: openshift-console
  disabled: true
authentication:
  disabled: true
authorization:
  disabled: true
```
but there may be cases where 2 operators could still be deployed.  Perhaps during an upgrade? In this case, it seems safer to leave leader election enabled to avoid the posibility of ever getting into a state where multiple operators fight over the console operand.

### DIFF
--- a/manifests/07-operator.yaml
+++ b/manifests/07-operator.yaml
@@ -4,7 +4,7 @@ metadata:
   name: console-operator
   namespace: openshift-console-operator
 spec:
-  replicas: 2
+  replicas: 1
   selector:
     matchLabels:
       name: console-operator


benjaminapetersen:
/retest

`level=fatal msg="failed to initialize the cluster: timed out waiting for the condition"`

benjaminapetersen:
/retest

`error: unable to read image registry.svc.ci.openshift.org/ci-op-yyf6bgrr/stable@sha256:c....cc3: received unexpected HTTP status: 504 Gateway Time-out
`

benjaminapetersen:
/retest

same, registry timeout.

benjaminapetersen:
/retry

`level=fatal msg="failed to initialize the cluster: timed out waiting for the condition"`

enj:
> The director for # of replicas for an operator seems to be:
> 
>     * Other operators depend on you: `replicas: 2`
> 
>     * Other operators do not depend on you: `replicas: 1`
> 
> 
> If the operator goes down & does not block another operator from recovering, a higher number of replicas is not needed
> 
>     * `thing you block, if len() > 0 { replicas = 2 } else { replicas = 1 }`.
> 
>     * Also `if have web facing traffic replicas >= 2 else replicas 1`.
> 
> 
> I almost updated leader election to:
> 
> ```
> apiVersion: operator.openshift.io/v1alpha1
> kind: GenericOperatorConfig
> leaderElection:
>   namespace: openshift-console
>   disabled: true
> authentication:
>   disabled: true
> authorization:
>   disabled: true
> ```
> 
> but there may be cases where 2 operators could still be deployed. Perhaps during an upgrade? In this case, it seems safer to leave leader election enabled to avoid the posibility of ever getting into a state where multiple operators fight over the console operand.

Agree with above.  The console itself must have 2+ replicas as it serves public web traffic.

/lgtm

benjaminapetersen:
rebased

PR #148, opened 2019-02-19
Filter route watch to only "console"

benjaminapetersen:
There is no reason to watch any other route.

### DIFF
--- a/pkg/console/starter/starter.go
+++ b/pkg/console/starter/starter.go
@@ -83,6 +83,10 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		options.FieldSelector = fields.OneTermEqualSelector("metadata.name", api.OAuthClientName).String()
 	}
 
+	tweakListOptionsForRoute := func(options *v1.ListOptions) {
+		options.FieldSelector = fields.OneTermEqualSelector("metadata.name", api.OpenShiftConsoleRouteName).String()
+	}
+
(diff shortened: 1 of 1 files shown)


benjaminapetersen:
@enj fyi

benjaminapetersen:
/retest

throttled.

pkg/console/starter/starter.go
```diff
@@ -83,6 +83,10 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		options.FieldSelector = fields.OneTermEqualSelector("metadata.name", api.OAuthClientName).String()
 	}
 
+	tweakListOptionsForRoute := func(options *v1.ListOptions) {
+		options.FieldSelector = fields.OneTermEqualSelector("metadata.name", api.OpenShiftConsoleRouteName).String()
```
enj: Please stop making different constants for the same string.  You do not even use them consistently:

`operator.WithInformer(routes, operator.FilterByNames(api.OpenShiftConsoleName)),`

benjaminapetersen: The `const` for `OpenShiftConsoleRouteName ` is not new.  It's old, from when we didn't have all matching names.  I'd rather use this one to match what is set & do a cleanup later & eliminate all the extra `const`s.  I believe we are down to a single name `console` for everything except the `configmaps` at this point.

PR #149, opened 2019-02-19
[WIP] Reduce noisy leader election logs

benjaminapetersen:


benjaminapetersen:
/hold

Still working on this. Having the leader election is nice, but our logs are bloated.

spadgett:
I don't think the log messages should dictate what values we use. We should set the values that we feel are best for the operator regardless of the logs.

benjaminapetersen:
When I look at log file size in the `/artifacts` dir, I'm usually surprised that the `console-operator` logs are quite large in comparison to other operators.  I don't think our leader election needs to be frequent.  With PR #147, we would only have a single pod under all normal running circumstances anyway.

I plucked the `90`,`45`,`30` from another operator (I think it was CVO) figuring this might be sufficient. The [auth operator PR 76](https://github.com/openshift/cluster-authentication-operator/pull/76) is turning off logs for leader-election entirely.  I'd rather keep them, but reduce the chatter.

spadgett:
> I don't think our leader election needs to be frequent.

That's fair. Just commenting that we shouldn't pick values only to make the log smaller. We should do what makes the most sense for our operator.

benjaminapetersen:
Yup! I agree.  I think we could prob go with even less frequent values honestly.

enj:
I still think you should have a default log level of 2 (like the kube api server operator, the authentication operator, etc).  At log level 2, you see no noise from leader election.

spadgett:
@benjaminapetersen any reason not to go ahead with this change?

benjaminapetersen:
Will close this & just bump our deps when this merges:
https://github.com/openshift/library-go/pull/305

Defaults will be changing.

benjaminapetersen:
Closing, #175 has been updated to pick up the latest library-go with the less noisy leader election defaults.

PR #150, opened 2019-02-19
Update sync loop to set Status.ConsoleURL instead of Status.PublicHostname

benjaminapetersen:
- Depends on [openshift api 212](https://github.com/openshift/api/pull/212)
- Requires a dependency bump


benjaminapetersen:
/assign @spadgett 

@enj fyi

PR #152, opened 2019-02-21
Validate URLs from the CR file and additional unit tests for ocp tag

zherman0:
This PR adds: 
1) Validation in the CRD to test that well formed URLs are supplied
   a) https://www.redhat.com:8080/logout.html is valid
   b) www.redhat.com (no http) is invalid
2) The validation also checks for trailing slash on the DOC link
   a) https://doc.com/     is vaild
   b) https://doc.com/doc    is invalid
3) Validate that the brand is one of the 5 acceptable values: ocp|origin|okd|dedicated|online
4) Validate that managedState is one of the 4 possible values: Managed|Unmanaged|Removed|Forced
5) Unit test for builds with the "ocp" tag set
6) There is no E2E test for URL at the moment because I have not figured out how to load/test values that are being detected at the oc api level
7) I had code in the config map to test URLs but have removed it since it duplicated the validation being done at the oc api level.

### DIFF
--- a/examples/cr-console.operator.config.yaml
+++ b/examples/cr-console.operator.config.yaml
@@ -1,0 +1,16 @@
+apiVersion: operator.openshift.io/v1
+kind: Console
+metadata:
+  name: cluster
+spec:
+  managementState: Managed
+  authentication:
+    # the logout redirect must be a valid URI that starts with http(s)
+    # valid examples: http://localhost:8080/logout.html, https://www.redhat.com/log/out/
+    logoutRedirect: "https://www.redhat.com/i/should/logout"
+  customization:
+    brand: okd
+    # the documentation base must be a valid URI that starts with http(s) AND it must have a trailing slash
+    # the trailing slash is required because additional values will be appended
+    # valid examples: https://localhost:9999/docs/, http://192.168.1.5:3000/doc/location/
+    documentationBaseURL: "https://www.redhat.com/someplace/for/docs/"
--- a/hack/test-unit.sh
+++ b/hack/test-unit.sh
@@ -26,3 +26,20 @@ do
     os::log::info "Testing ${PACKAGE}"
     go test $PACKAGE
 done
+
+OCP_PACKAGES_TO_TEST=(
+    "github.com/openshift/console-operator/pkg/console/subresource/configmap"
+    "github.com/openshift/console-operator/pkg/console/subresource/deployment"
+    "github.com/openshift/console-operator/pkg/console/subresource/oauthclient"
+    "github.com/openshift/console-operator/pkg/console/subresource/route"
+    "github.com/openshift/console-operator/pkg/console/subresource/secret"
+    "github.com/openshift/console-operator/pkg/console/subresource/service"
+    "github.com/openshift/console-operator/pkg/console/subresource/util"
+    "github.com/openshift/console-operator/pkg/console/version"
+)
+
+for PACKAGE in "${OCP_PACKAGES_TO_TEST[@]}"
+do
+    os::log::info "Testing with tag ocp ${PACKAGE}"
+    go test -tags "ocp" $PACKAGE
+done
--- a/manifests/00-crd-operator-config.yaml
+++ b/manifests/00-crd-operator-config.yaml
@@ -13,6 +13,32 @@ spec:
   subresources:
     status: {}
   versions:
-  - name: v1
-    served: true
-    storage: true
+    - name: v1
+      served: true
+      storage: true
+  validation:
+    openAPIV3Schema:
+      properties:
+        spec:
+          properties:
+            managementState:
+              pattern: ^(Managed|Unmanaged|Removed|Forced)$
+              type: string
+              description: managementState indicates whether and how the operator
+                should manage the component
+            authentication:
+              properties:
+                logoutRedirect:
+                  pattern: ^((https):\/\/?)[^\s()<>]+(?:\([\w\d]+\)|([^[:punct:]\s]|\/?))$
+                  type: string
+                  description: Logout Redirect must be a valid URL
+            customization:
+              properties:
+                documentationBaseURL:
+                  pattern: ^((https):\/\/?)[^\s()<>]+(?:\([\w\d]+\)|([^[:punct:]\s]|\/?))\/$
+                  type: string
+                  description: Document base URL must be a valid URL that ends in a trailing slash
+                brand:
+                  pattern: ^(ocp|origin|okd|dedicated|online|azure)$
+                  type: string
+                  description: Brand must be be one of six values - azure|dedicated|ocp|okd|online|origin
--- a/pkg/console/subresource/configmap/configmap_test.go
+++ b/pkg/console/subresource/configmap/configmap_test.go
@@ -27,8 +27,8 @@ clusterInfo:
   consoleBaseAddress: https://` + host + `
   consoleBasePath: ""
 customization:
-  branding: okd
-  documentationBaseURL: https://docs.okd.io/4.0/
+  branding: ` + DEFAULT_BRAND + `
+  documentationBaseURL: ` + DEFAULT_DOC_URL + `
 servingInfo:
   bindAddress: https://0.0.0.0:8443
   certFile: /var/serving-cert/tls.crt
@@ -171,8 +171,8 @@ func TestNewYamlConfig(t *testing.T) {
 			args: args{
 				host:           host,
 				logoutRedirect: "",
-				brand:          "okd",
-				docURL:         "https://docs.okd.io/4.0/",
+				brand:          DEFAULT_BRAND,
+				docURL:         DEFAULT_DOC_URL,
 			},
 			want: exampleYaml,
 		},


zherman0:
@benjaminapetersen @spadgett - take a look.  Still trying to figure out if I can E2E the validation of the URLs at the oc api level within our operator test.

hack/test-unit.sh
```diff
@@ -26,3 +26,20 @@ do
     os::log::info "Testing ${PACKAGE}"
     go test $PACKAGE
 done
+
```
benjaminapetersen: So our `ci` jobs will call `make test-unit`. Does this need a `makefile` update? Thinking.

hack/test-unit.sh
```diff
@@ -26,3 +26,20 @@ do
...
+OCP_PACKAGES_TO_TEST=(
```
benjaminapetersen: I wonder if we can deduplicate somehow.  IE,

```
PACKAGES_TO_TEST=(/* most go here */)
PACKAGES_TO_TEST_FOR_OKD=(/* just one? */)
PACKAGES_TO_TEST_FOR_OCP=(/* just one? */)
// then always run the main set, but conditionally run `_FOR_OKD` & `_FOR_OCP`.  
```
Would this just be a flag to check to run one more loop?

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -27,8 +27,8 @@ clusterInfo:
   consoleBaseAddress: https://` + host + `
   consoleBasePath: ""
 customization:
-  branding: okd
-  documentationBaseURL: https://docs.okd.io/4.0/
+  branding: ` + DEFAULT_BRAND + `
```
benjaminapetersen: I dont mind this, but probably would prefer a separate test as well. The problem with the tests currently is that, when they fail, the debugging is pretty difficult as its a manual comparison between 2 separate large `JSON` blobs.

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -171,8 +171,8 @@ func TestNewYamlConfig(t *testing.T) {
 			args: args{
 				host:           host,
 				logoutRedirect: "",
-				brand:          "okd",
-				docURL:         "https://docs.okd.io/4.0/",
+				brand:          DEFAULT_BRAND,
```
benjaminapetersen: I wonder if we can do something like:

```golang
// configmap.go
func defaultBrand() string {
    return DEFAULT_BRAND
}
func defaultDocURL() sting {
   return DEFAULT_DOC_URL
}
// configmap_test.go
TestDefaultBrand () {} 
TestDefaultDocURL() {}
```

benjaminapetersen: Would that then require:

```golang
func getBrand(operatorConfig *operatorv1.Console) operatorv1.Brand {
	if len(operatorConfig.Spec.Customization.Brand) > 0 {
		return operatorConfig.Spec.Customization.Brand
	}
	return defaultBrand()  // because we can test a func, cant test a string
}
```

benjaminapetersen: It does add an additional wrapper `func`, but the testing would be fairly simple...

manifests/00-crd-operator-config.yaml
```diff
@@ -13,6 +13,25 @@ spec:
   subresources:
     status: {}
   versions:
-  - name: v1
-    served: true
-    storage: true
+    - name: v1
```
benjaminapetersen: It may be good to do this as a separate PR.

hack/test-unit.sh
benjaminapetersen: This is probably not a big deal. Resolving, if we ever had more complexity we can worry about it then.

hack/test-unit.sh
benjaminapetersen: nope, i see your `ocp` tag in the second run. great!

spadgett:
We should also validate management state and look at what other operators are doing for unsupported config overrides.

enj:
Trailing slash on URLs should be disallowed.

https://foo.com is what I expect to see in URL fields.

spadgett:
> Trailing slash on URLs should be disallowed.
> 
> https://foo.com is what I expect to see in URL fields.

It is actually required by console to make sure it is a directory since we append to it.

spadgett:
For documentation URL that is

pkg/console/subresource/configmap/configmap_test.go
zherman0: Well, we could add these functions and do super simple unit test and if there is a problem (and we execute them first) then we might see there is an issue.  However, the other unit tests include the brand and doc link already as part of the deep copy so we would be testing it twice.  I know the output is not perfect, but perhaps we should work on better functionality to show the differences then write other helper functions.

pkg/console/subresource/configmap/configmap_test.go
zherman0: Yeah, it is hard to investigate so we probably should create some better unit test tools to help show what specifically is failing on a deep copy.

manifests/00-crd-operator-config.yaml
zherman0: Can you be more specific here?  Do you want the new CRD in one PR and the unit tests in another?  What about any E2E tests (if they are added)?

examples/customUrls.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: operator.openshift.io/v1
```
benjaminapetersen: I may name this file `/examples/cr-console.operator.config.yaml` or something like that. `customUrls.yaml` may not help the user understand which config we are dealing with (I've had to answer this question numerous times & agree its confusing).

manifests/00-crd-operator-config.yaml
```diff
@@ -13,6 +13,25 @@ spec:
...
+            authentication:
+              properties:
+                logoutRedirect:
+                  pattern: ^((http|https):\/\/?)[^\s()<>]+(?:\([\w\d]+\)|([^[:punct:]\s]|\/?))$
+                  type: string
+            customization:
+              properties:
+                documentationBaseURL:
+                  #pattern: ^(?:http(s)?:\/\/)?[\w.-]+(?:\.[\w\.-]+)+[\w\-\._~:/?#[\]@!\$&'\(\)\*\+,;=.]+\/$
+                  pattern: ^((http|https):\/\/?)[^\s()<>]+(?:\([\w\d]+\)|([^[:punct:]\s]|\/?))\/$
```
benjaminapetersen: @enj has had several objections to the trailing `/` for `documentationBaseURL`.  
@spadgett fyi 
Not sure if we want to chat about that now before we add validation for it.

pkg/console/subresource/configmap/configmap_test.go
benjaminapetersen: There is a object diff output package I've heard mentioned several times to solve this problem.  It might be sufficient.

manifests/00-crd-operator-config.yaml
benjaminapetersen: I tend to be a fan of merging smaller PRs more frequently :)

examples/customUrls.yaml
zherman0: agreed

manifests/00-crd-operator-config.yaml
zherman0: I will break this up a little more.

manifests/00-crd-operator-config.yaml
spadgett: The documentation URL currently is required to end with `/` since we append topics to it. It will prevent users from adding a URL like `https://docs.okd.io/latest/welcome/index.html` which would break our links.

I think we need to keep the trailing slash.

examples/cr-console.operator.config.yaml
```diff
@@ -0,0 +1,15 @@
...
+kind: Console
+metadata:
+  name: cluster
+spec:
+  authentication:
+    # the logout redirect must be a valid URI that starts with http(s)
+    # valid examples: http://localhost:8080/logout.html, https://www.redhat.com/log/out/
+    logoutRedirect: "https://www.redhat.com/i/should/logout"
+  customization:
+    brand: "okd"
```
benjaminapetersen: nit: doesn't need quotes on the `okd` string

pkg/console/subresource/configmap/configmap_test.go
benjaminapetersen: works for me.  Looking @ #159 as a possible solution.

manifests/00-crd-operator-config.yaml
enj: For the top level config `consoleURL` I will require the following validation (because the rules are simple and precise to explain - and they make it easy to consume the string):

1. no error from `url.Parse`
1. scheme must be `https`
1. domain must be valid
1. no path
1. no query
1. no fragment

This means no trailing slash (as that means path = `/`).  It makes sense to make the other URL validation consistent (up to you to decide if it really makes sense to allow `http` scheme on the other URLs) across all configs.  Thus when you append, always do `documentationBaseURL` + `/` + `topic` (I think that is what you are saying you need to be able to do).

manifests/00-crd-operator-config.yaml
```diff
@@ -13,6 +13,25 @@ spec:
...
+        spec:
+          properties:
+            authentication:
+              properties:
+                logoutRedirect:
+                  pattern: ^((http|https):\/\/?)[^\s()<>]+(?:\([\w\d]+\)|([^[:punct:]\s]|\/?))$
+                  type: string
+            customization:
+              properties:
+                documentationBaseURL:
```
benjaminapetersen: We should probably remove the `#pattern` commented out.

manifests/00-crd-operator-config.yaml
spadgett: But then we don't catch the case where the user puts in `/index.html` which becomes `/index.html/<topic>`. The point of requiring the trailing `/` is to make it clear to the admin changing this value that it must be a directory and failing validation if they don't, instead of having broken URLs in the UI.

benjaminapetersen: So I think we can resolve this.

enj: Based on talking on slack with @spadgett and @benjaminapetersen I see the requirements for `documentationBaseURL` to be:

1. no error from `url.Parse`
1. scheme must be `https`
1. domain must be valid
1. **path must be non-empty and must end in slash**
1. no query
1. no fragment

enj: (I realize that you have to use regex hacks and cannot use `url.Parse`)

spadgett: Looks right, thanks

Who are we describing to? If it's the person editing the config, "no error from `url.Parse`" probably isn't meaningful

benjaminapetersen: Right, unfortunately when this fails all it will do is dump the regex, which will not clearly describe any of that. 😢

spadgett: At least we could put something in the doc so that `oc explain` or the help topic describes what it needs to be. Agree the regex error message is not a good user experience.

zherman0: @enj - As for forcing https, I am worried about testing when you have no cert.  For doc link, it might be fine, but logout could easily not have an https option.

enj: The logout url presumably redirects you to a page where you can manage your SSO session.  How would it ever be safe if it was http?

Minting a CA/certs in tests, while annoying, is not hard.

zherman0: I certainly agree that I would not want people to use unsecure methodolgies, but I was more just stating there are occasions were having the option of http would be nice.  I run the operator/console locally on libvirt and I would prefer to not have to setup a cert.  That said, I am just trying to debate the state of the internet and how people operate on whether we can safely assume that no one will use http.  While I would not recommend it, could a customer want to setup an openshift cluster on-prem inside their private network that has no certs for their developers?  I would hate to make things so restrictive as to discourage people from using or trying okd.

zherman0: @spadgett @benjaminapetersen - Do you think I should limit this to only https for now?  I am fine either way.

manifests/00-crd-operator-config.yaml
benjaminapetersen: @zherman0

manifests/00-crd-operator-config.yaml
benjaminapetersen: My observation is that we tend to say we are secure by default, so I'd lean toward keeping the `s`.

spadgett: +1 to https. We're not able to make it more restrictive later, so I'd prefer https for now.

zherman0: Unsecure http has been removed

manifests/00-crd-operator-config.yaml
```diff
@@ -13,6 +13,32 @@ spec:
...
+                  pattern: ^((https):\/\/?)[^\s()<>]+(?:\([\w\d]+\)|([^[:punct:]\s]|\/?))\/$
+                  type: string
+                  description: Document base URL must be a valid URL that ends in a trailing slash
+                brand:
+                  pattern: ^(ocp|origin|okd|dedicated|online|azure)$
+                  type: string
+                  description: Brand must be be one of five values - ocp|origin|okd|dedicated|online
```
benjaminapetersen: `description` for brand also needs to be updated:

`description: Brand must be be one of five values - okd|ocp|origin|dedicated|online|azure`

I don't know if ordering matters... alphabetical vs upstream first vs?

benjaminapetersen: @zherman0 i think the only lingering is getting `azure` in the description here.

zherman0: @benjaminapetersen - fixed description - how does it look?

zherman0:
@spadgett @benjaminapetersen - Can I get an lgtm?

manifests/00-crd-operator-config.yaml
```diff
@@ -13,6 +13,32 @@ spec:
...
+        spec:
+          properties:
+            managementState:
+              pattern: ^(Managed|Unmanaged|Removed|Forced)$
+              type: string
+              description: managementState indicates whether and how the operator
+                should manage the component
+            authentication:
+              properties:
+                logoutRedirect:
```
benjaminapetersen: Late comment, but just realized, `authentication.logoutRedirect` is on `console.config.openshift.io`, not `console.operator.openshift.io` (the top level user-facing console config, not the config for the operator).  This exists here:  https://github.com/openshift/cluster-config-operator/blob/master/manifests/0000_05_config-operator_01_console.crd.yaml

- [ ] We should remove this, PR it there against that file.
- [ ] Also, lets ensure it handles `""` / unset as these are unset by default and optional.

manifests/00-crd-operator-config.yaml
```diff
@@ -13,6 +13,32 @@ spec:
...
+                  pattern: ^((https):\/\/?)[^\s()<>]+(?:\([\w\d]+\)|([^[:punct:]\s]|\/?))$
+                  type: string
+                  description: Logout Redirect must be a valid URL
+            customization:
```
benjaminapetersen: Also, `customization.documentationBaseURL` and `customization.Brand` should handle `""`/unset. I'm betting these do not. Fast-follow!

examples/cr-console.operator.config.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: operator.openshift.io/v1
+kind: Console
+metadata:
+  name: cluster
+spec:
+  managementState: Managed
+  authentication:
```
benjaminapetersen: Same here, @zherman0, `authentication` belongs on the other config, should remove from this example file.

PR #158, opened 2019-02-25
Rename Dockerfile.rhel7 -> Dockerfile.rhel

benjaminapetersen:
@adammhaile per conversation.
@spadgett fyi

coordination:
- [ ] Update https://github.com/openshift/ocp-build-data/blob/openshift-4.0/images/openshift-enterprise-console-operator.yml#L3 at EXACTLY the same time.  If not, builds will break.


adammhaile:
This will need an update to ocp-build-data

spadgett:
@benjaminapetersen @adammhaile I believe we can switch to a single Dockerfile when #180 is done

benjaminapetersen:
This is no longer needed, we will not be building 2 images.
Semi-related to #180

PR #160, opened 2019-02-27
Change unit test to use cleaner error output

zherman0:
Change unit test to use cleaner error output.  Currently any error from unit test failing shows a large blob of data that one must scour through to see what has failed.  This update uses the go-test/deep pkg to only show what has failed in a clean, simple output.
Before:
unit-befoe
After:
unit-after

### DIFF
--- a/pkg/console/subresource/configmap/configmap_test.go
+++ b/pkg/console/subresource/configmap/configmap_test.go
@@ -2,7 +2,7 @@ package configmap
 
 import (
 	"fmt"
-	"reflect"
+	"github.com/go-test/deep"
 	"testing"
 
 	corev1 "k8s.io/api/core/v1"
@@ -103,8 +103,8 @@ func TestDefaultConfigMap(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			if got := DefaultConfigMap(tt.args.operatorConfig, tt.args.consoleConfig, tt.args.rt); !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("DefaultConfigMap() = %v\n ----------- want %v", got, tt.want)
+			if diff := deep.Equal(DefaultConfigMap(tt.args.operatorConfig, tt.args.consoleConfig, tt.args.rt), tt.want); diff != nil {
+				t.Error(diff)
 			}
 		})
 	}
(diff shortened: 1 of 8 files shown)


examples/05-operator-alt-image.yaml
```diff
@@ -16,9 +16,14 @@ spec:
       serviceAccountName: console-operator
       containers:
       - name: console-operator
-        # image: quay.io/benjaminapetersen/console-operator:v4.0.0.6
-        # image: quay.io/benjaminapetersen/console-operator-2:v4.1.0
-        image: quay.io/benjaminapetersen/console-operator-2:v4.1.2
+        # for development
```
benjaminapetersen: Was this intended to be included?  I believe the change to this file has already merged, may need a rebase.

benjaminapetersen:
Generally `lgtm`, looks like it failed to pick up the new package.

spadgett:
hey @zherman0 do you have sample output handy?

i'm sure this is much better than before :)

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -89,12 +89,12 @@ func TestDefaultConfigMap(t *testing.T) {
 					CreationTimestamp:          metav1.Time{},
 					DeletionTimestamp:          nil,
 					DeletionGracePeriodSeconds: nil,
-					Labels:          map[string]string{"app": api.OpenShiftConsoleName},
-					Annotations:     map[string]string{},
-					OwnerReferences: nil,
-					Initializers:    nil,
-					Finalizers:      nil,
-					ClusterName:     "",
+					Labels:                     map[string]string{"app": api.OpenShiftConsoleName},
```
spadgett: you should check that you're using go 1.10

benjaminapetersen:
Agree, a good before / after screenshot would be handy.

examples/05-operator-alt-image.yaml
zherman0: no, removed

pkg/console/subresource/configmap/configmap_test.go
zherman0: resolved

zherman0:
Updated description with images of new output

zherman0:
@spadgett @benjaminapetersen - Can I get an lgtm?

benjaminapetersen:
@zherman0 definitely, mind a quick squash?

PR #162, opened 2019-02-27
Support Spec.UnsupportedConfigOverrides

benjaminapetersen:
`Spec.UnsupportedConfigOverrides` essentially should allow a user to replace our configmap.  This should only be needed as an escape hatch if something terrible has gone wrong.

When used, the `ClusterOperator` for `Console` will record the config overrides under `Status.Conditions.Upgradable`.  This is consistent with what is done by [kube-apiserver-operator](https://github.com/openshift/cluster-kube-apiserver-operator).

ClusterOperator:
`oc get clusteroperator/console -w -o yaml`

screen shot 2019-03-01 at 3 19 18 pm

### DIFF
--- a/pkg/console/operator/status.go
+++ b/pkg/console/operator/status.go
@@ -55,6 +55,7 @@ const (
 	reasonSyncLoopProgressing = "SyncLoopProgressing"
 	reasonNoPodsAvailable     = "NoPodsAvailable"
 	reasonSyncError           = "SyncError"
+	reasonAsExpected          = "AsExpected"
 )
 
 // Lets transition to using this, and get the repetition out of all of the above.
@@ -183,14 +184,14 @@ func (c *consoleOperator) ConditionResourceSyncFailure(operatorConfig *operators
 	v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorsv1.OperatorCondition{
 		Type:               operatorsv1.OperatorStatusTypeAvailable,
 		Status:             operatorsv1.ConditionUnknown,
-		Reason:             reasonUnmanaged,
+		Reason:             reasonSyncError,
 		Message:            message,
 		LastTransitionTime: metav1.Now(),
 	})
 	v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorsv1.OperatorCondition{
 		Type:               operatorsv1.OperatorStatusTypeProgressing,
 		Status:             operatorsv1.ConditionTrue,
-		Reason:             reasonUnmanaged,
+		Reason:             reasonSyncError,
 		Message:            message,
 		LastTransitionTime: metav1.Now(),
 	})
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -235,11 +235,18 @@ func SyncSecret(co *consoleOperator, recorder events.Recorder, operatorConfig *o
 // therefore no additional error handling is needed here.
 func SyncConfigMap(co *consoleOperator, recorder events.Recorder, operatorConfig *operatorv1.Console, consoleConfig *configv1.Console, infrastructureConfig *configv1.Infrastructure, rt *routev1.Route) (*corev1.ConfigMap, bool, error) {
 	logrus.Printf("validating console configmap...")
-	cm, cmChanged, cmErr := resourceapply.ApplyConfigMap(co.configMapClient, recorder, configmapsub.DefaultConfigMap(operatorConfig, consoleConfig, infrastructureConfig, rt))
+
+	defaultConfigmap, _, err := configmapsub.DefaultConfigMap(operatorConfig, consoleConfig, infrastructureConfig, rt)
+	if err != nil {
+		return nil, false, err
+	}
+
+	cm, cmChanged, cmErr := resourceapply.ApplyConfigMap(co.configMapClient, recorder, defaultConfigmap)
 	if cmErr != nil {
 		logrus.Errorf("%q: %v \n", "configmap", cmErr)
 		return nil, false, cmErr
 	}
+
 	logrus.Println("configmap exists and is in the correct state")
 	return cm, cmChanged, cmErr
 }
--- a/pkg/console/starter/starter.go
+++ b/pkg/console/starter/starter.go
@@ -5,6 +5,8 @@ import (
 	"os"
 	"time"
 
+	"github.com/openshift/library-go/pkg/operator/unsupportedconfigoverridescontroller"
+
 	// kube
 
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -175,6 +177,8 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		ctx.EventRecorder,
 	)
 
+	configUpgradeableController := unsupportedconfigoverridescontroller.NewUnsupportedConfigOverridesController(operatorClient, ctx.EventRecorder)
+
 	kubeInformersNamespaced.Start(ctx.Done())
 	operatorConfigInformers.Start(ctx.Done())
 	configInformers.Start(ctx.Done())
@@ -183,7 +187,8 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 
 	go consoleOperator.Run(ctx.Done())
 	go clusterOperatorStatus.Run(1, ctx.Done())
-	<-ctx.Done()
+	go configUpgradeableController.Run(1, ctx.Done())
 
+	<-ctx.Done()
 	return fmt.Errorf("stopped")
 }
--- a/pkg/console/subresource/configmap/configmap.go
+++ b/pkg/console/subresource/configmap/configmap.go
@@ -3,6 +3,9 @@ package configmap
 import (
 	"fmt"
 
+	"github.com/sirupsen/logrus"
+
+	yaml2 "github.com/ghodss/yaml"
 	yaml "gopkg.in/yaml.v2"
 
 	corev1 "k8s.io/api/core/v1"
@@ -12,6 +15,7 @@ import (
 	routev1 "github.com/openshift/api/route/v1"
 	"github.com/openshift/console-operator/pkg/api"
 	"github.com/openshift/console-operator/pkg/console/subresource/util"
+	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
 )
 
 const (
@@ -57,22 +61,49 @@ func getApiUrl(infrastructureConfig *configv1.Infrastructure) string {
 	return ""
 }
 
-func DefaultConfigMap(operatorConfig *operatorv1.Console, consoleConfig *configv1.Console, infrastructureConfig *configv1.Infrastructure, rt *routev1.Route) *corev1.ConfigMap {
-
+// DefaultConfigMap returns
+// - a new configmap,
+// - a bool indicating if config was merged (unsupportedConfigOverrides)
+// - an error
+func DefaultConfigMap(operatorConfig *operatorv1.Console, consoleConfig *configv1.Console, infrastructureConfig *configv1.Infrastructure, rt *routev1.Route) (*corev1.ConfigMap, bool, error) {
 	logoutRedirect := getLogoutRedirect(consoleConfig)
 	brand := getBrand(operatorConfig)
 	docURL := getDocURL(operatorConfig)
 	apiServerURL := getApiUrl(infrastructureConfig)
 
 	host := rt.Spec.Host
-	config := string(NewYamlConfig(host, logoutRedirect, brand, docURL, apiServerURL))
+	config := NewYamlConfig(host, logoutRedirect, brand, docURL, apiServerURL)
+
 	configMap := Stub()
-	configMap.Data = map[string]string{
-		consoleConfigYamlFile: config,
+	configMap.Data = map[string]string{}
+	unsupportedRaw := operatorConfig.Spec.UnsupportedConfigOverrides.Raw
+
+	// merge config overrides, if we have them
+	mergedConfig, err := resourcemerge.MergeProcessConfig(nil, config, unsupportedRaw)
+	if err != nil {
+		logrus.Errorf("failed to generate configmap: %v \n", err)
+		return nil, false, err
 	}
 
+	outConfigYaml, err := yaml2.JSONToYAML(mergedConfig)
+	if err != nil {
+		logrus.Errorf("failed to generate configmap: %v \n", err)
+		return nil, false, err
+	}
+
+	// if we actually merged config overrides, log this information
+	didMerge := len(operatorConfig.Spec.UnsupportedConfigOverrides.Raw) != 0
+	if didMerge {
+		logrus.Println(fmt.Sprintf("With UnsupportedConfigOverrides: %v", string(unsupportedRaw)))
+	}
+
+	configMap.Data[consoleConfigYamlFile] = string(outConfigYaml)
+
+	logrus.Println("generated console config yaml:")
+	logrus.Printf("%s \n", string(outConfigYaml))
 	util.AddOwnerRef(configMap, util.OwnerRefFrom(operatorConfig))
-	return configMap
+
+	return configMap, didMerge, nil
 }
 
 func Stub() *corev1.ConfigMap {
--- a/pkg/console/subresource/configmap/configmap_test.go
+++ b/pkg/console/subresource/configmap/configmap_test.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"testing"
 
+	yaml "gopkg.in/yaml.v2"
+
 	"github.com/go-test/deep"
 
 	corev1 "k8s.io/api/core/v1"
@@ -18,6 +20,7 @@ import (
 const (
 	host          = "localhost"
 	mockAPIServer = "https://api.some.cluster.openshift.com:6443"
+	configKey     = "console-config.yaml"
 	exampleYaml   = `kind: ConsoleConfig
 apiVersion: console.openshift.io/v1beta1
 auth:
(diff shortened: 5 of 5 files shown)


pkg/console/subresource/configmap/configmap.go
```diff
@@ -57,13 +60,25 @@ func DefaultConfigMap(operatorConfig *operatorv1.Console, consoleConfig *configv
 	docURL := getDocURL(operatorConfig)
 
 	host := rt.Spec.Host
-	config := string(NewYamlConfig(host, logoutRedirect, brand, docURL))
+	config := NewYamlConfig(host, logoutRedirect, brand, docURL)
+
+	config, err := resourcemerge.MergeProcessConfig(nil, config, operatorConfig.Spec.UnsupportedConfigOverrides.Raw)
+	if err != nil {
+		// TODO: decide if we should die here.
```
enj: you need to error out - a mistake here should clearly break the operator

pkg/console/subresource/configmap/configmap.go
```diff
@@ -57,13 +60,25 @@ func DefaultConfigMap(operatorConfig *operatorv1.Console, consoleConfig *configv
...
+		// log the error & keep working with the above NewYamlConfig() ?
+		logrus.Errorf("could not create configmap")
+	}
+
+	configString := string(config)
+
 	configMap := Stub()
 	configMap.Data = map[string]string{
-		consoleConfigYamlFile: config,
+		consoleConfigYamlFile: configString,
```
enj: just do `string(config)` here

pkg/console/subresource/configmap/configmap.go
```diff
@@ -3,6 +3,9 @@ package configmap
 import (
 	"fmt"
 
+	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
```
enj: order

benjaminapetersen: bah, really hate that `goland` collapses these by default.

pkg/console/operator/status.go
```diff
@@ -55,6 +55,7 @@ const (
 	reasonSyncLoopProgressing = "SyncLoopProgressing"
 	reasonNoPodsAvailable     = "NoPodsAvailable"
 	reasonSyncError           = "SyncError"
+	reasonAsExpected          = "AsExpected"
```
benjaminapetersen: This `reason` is also used by [kube-apiserver-operator](https://github.com/openshift/cluster-kube-apiserver-operator)

pkg/console/operator/status.go
```diff
@@ -183,14 +184,14 @@ func (c *consoleOperator) ConditionResourceSyncFailure(operatorConfig *operators
 	v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorsv1.OperatorCondition{
 		Type:               operatorsv1.OperatorStatusTypeAvailable,
 		Status:             operatorsv1.ConditionUnknown,
-		Reason:             reasonUnmanaged,
+		Reason:             reasonSyncError,
```
benjaminapetersen: copypasta error fixed

benjaminapetersen:
/hold

`resourcemerge.MergeProcessConfig()` converts our `yaml` config into `json`.  not ideal, `console` pods `crashloop`.  fixing.

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -103,7 +104,18 @@ func TestDefaultConfigMap(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			if diff := deep.Equal(DefaultConfigMap(tt.args.operatorConfig, tt.args.consoleConfig, tt.args.rt), tt.want); diff != nil {
+			cm, _, _, _ := DefaultConfigMap(tt.args.operatorConfig, tt.args.consoleConfig, tt.args.rt)
+
```
benjaminapetersen: @zherman0 @spadgett 
This is a tricky one.  The embedded yaml string really can't be tested well due to unstable ordering (from JSON conversion).  Fortunately, we have the other `TestNewYamlConfig()` func below.

Perhaps this is the way to get these changes unstuck.

pkg/console/operator/sync_v400.go
```diff
@@ -213,7 +213,14 @@ func SyncSecret(co *consoleOperator, recorder events.Recorder, operatorConfig *o
 // therefore no additional error handling is needed here.
 func SyncConfigMap(co *consoleOperator, recorder events.Recorder, operatorConfig *operatorv1.Console, consoleConfig *configv1.Console, rt *routev1.Route) (*corev1.ConfigMap, bool, error) {
 	logrus.Printf("validating console configmap...")
-	cm, cmChanged, cmErr := resourceapply.ApplyConfigMap(co.configMapClient, recorder, configmapsub.DefaultConfigMap(operatorConfig, consoleConfig, rt))
+	defaultConfigmap, didMerge, configMergeMessage, err := configmapsub.DefaultConfigMap(operatorConfig, consoleConfig, rt)
+	if didMerge {
```
benjaminapetersen: I don't love having the `co.UpgradableWithUnsupportedConfigOverrides` down in here, I'd rather be moving these further up the stack into `starter.go`, but I can live with them so long as they stay within `sync_v400.go` until we have a good tech debt time to properly extract this all out to one place.

pkg/console/subresource/configmap/configmap_test.go
benjaminapetersen: The problem here is that its an obvious hack and it mutates the output.
Also, if the func is changed & for some reason the `cm.Data` doesn't return a string, we miss that.  So we could miss a legitimate error.

zherman0:
looks reasonable to me but I'll let @spadgett or @jhadvig look as well

pkg/console/subresource/configmap/configmap.go
```diff
@@ -3,6 +3,10 @@ package configmap
 import (
 	"fmt"
 
+	yaml2 "github.com/ghodss/yaml"
```
jhadvig: This is a wrapper around the `gopkg.in/yaml.v2` we are using below. Had checked the [docs](https://github.com/ghodss/yaml#compatibility) and should be compatible with it, so there shouldn't be need for keeping both **yaml** libs

pkg/console/operator/status.go
```diff
@@ -318,3 +319,15 @@ func (c *consoleOperator) ConditionsManagementStateRemoved(operatorConfig *opera
 
 	return operatorConfig
 }
+
+func (c *consoleOperator) UpgradableWithUnsupportedConfigOverrides(operatorConfig *operatorsv1.Console, message string) *operatorsv1.Console {
```
spadgett: This would be better inlined. I don't see a good reason to wrap a one-liner with a helper, and it obscures what it's setting. We're only calling it in one place.

The helpers make more sense to me if they're setting several related conditions together, but not in the case where it's setting a single condition.

pkg/console/operator/sync_v400.go
```diff
@@ -236,7 +236,16 @@ func SyncSecret(co *consoleOperator, recorder events.Recorder, operatorConfig *o
 // therefore no additional error handling is needed here.
 func SyncConfigMap(co *consoleOperator, recorder events.Recorder, operatorConfig *operatorv1.Console, consoleConfig *configv1.Console, infrastructureConfig *configv1.Infrastructure, rt *routev1.Route) (*corev1.ConfigMap, bool, error) {
 	logrus.Printf("validating console configmap...")
-	cm, cmChanged, cmErr := resourceapply.ApplyConfigMap(co.configMapClient, recorder, configmapsub.DefaultConfigMap(operatorConfig, consoleConfig, infrastructureConfig, rt))
+
+	defaultConfigmap, didMerge, configMergeMessage, err := configmapsub.DefaultConfigMap(operatorConfig, consoleConfig, infrastructureConfig, rt)
+	if didMerge {
+		co.UpgradableWithUnsupportedConfigOverrides(operatorConfig, configMergeMessage)
```
spadgett: So it looks like we never set upgrade anywhere else. Does that mean that the message will be wrong when I disable the unsupported config?

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -105,14 +106,30 @@ func TestDefaultConfigMap(t *testing.T) {
...
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			if diff := deep.Equal(DefaultConfigMap(tt.args.operatorConfig, tt.args.consoleConfig, tt.args.infrastructureConfig, tt.args.rt), tt.want); diff != nil {
+			cm, _, _, _ := DefaultConfigMap(tt.args.operatorConfig, tt.args.consoleConfig, tt.args.infrastructureConfig, tt.args.rt)
+			// Deep.Equal doesn't work well for checking the value of the embedded
+			// console-config.yaml file, as it is just a string.  The MergeProcessConfig()
+			// call within DefaultConfigMap() converts the value to JSON to do the merge,
+			// then converts back, therefore we can't trust the ordering of the strings.
+			//
+			// so here we are going to just check that we have a value
```
spadgett: Why not parse the string as YAML and make sure the result matches what's expected?

pkg/console/subresource/configmap/configmap.go
benjaminapetersen: Ha, my bad.  Didn't look at my imports to realize I imported a second yaml package.  Thanks!

benjaminapetersen: Actually, compatible but not the same interface available.

`yaml2.JSONToYAML()` is something we gain.  I could potentially look for a way to do this manually, but I'm not sure I want to fiddle with that.

pkg/console/subresource/configmap/configmap_test.go
benjaminapetersen: Well, if we are just doing a `string(SomeYAML)` here https://github.com/openshift/console-operator/pull/162/files#diff-ffa541898964e88404798d9f8c923256R121

Then the `TestNewYamlConfig()` test is really already doing the testing we need on this bit.

I imagine if I were going to parse it to check, I'd turn it back into a `go struct` and do more or less the same thing again.

pkg/console/subresource/configmap/configmap.go
jhadvig: Yeah If the interface is different we can maybe just add a "techdebt" comment, that in case we got time we can update, but would really pay attention if its time-consuming.

pkg/console/operator/status.go
jhadvig: refact the status setting in the operator (as we already discussed) and probably use the `v1helpers.SetOperatorCondition` method directly.

pkg/console/operator/sync_v400.go
jhadvig: @spadgett you mean when the `operatorConfig.Spec.UnsupportedConfigOverrides.Raw` ? Cause we are setting the status, when that field is set

pkg/console/operator/status.go
benjaminapetersen: TL;DR its not ideal, but inconsistent seems worse than the correct fix. Lets fix it the right way.

We should not refactor all of the status settings in this PR, that should be done as a separate tech debt PR (as I've said before, I agree, although strongly that it should be factored out into one place -- return enough structures from `sync_v400()` to handle status once, without breaking the fail-early mechanism.

I hesitate to introduce **inconsistencies** before then and just make it messier.  This is really just a simple `factory` function.

pkg/console/subresource/configmap/configmap.go
benjaminapetersen: `yaml` lib has funcs for creating `MapSlice`, `MapItem`, and a bunch of others.  Its a good core YAML lib with a lot of features.  `yaml2` is pretty small (300 lines), and uses `yaml` itself under the hood.  It might just be poorly named for a set of additional helper funcs to the main `yaml` lib.

I don't think we can drop `yaml`, we can't swap `yaml2` for `yaml`, we'd end up just dropping `yaml2` and writing the funcs ourselves.

benjaminapetersen:
Should make use of:

https://github.com/openshift/library-go/blob/c808b4f42728549453f702c5bdc21413f02f09d4/pkg/operator/unsupportedconfigoverridescontroller/unsupportedconfigoverrides_controller.go

pkg/console/operator/status.go
benjaminapetersen: Eliminating this entirely in my next push, will have `NewUnsupportedConfigOverridesController()` handle the status reportin.

pkg/console/operator/sync_v400.go
benjaminapetersen: Same, removing since using `NewUnsupportedConfigOverridesController()`

pkg/console/operator/sync_v400.go
benjaminapetersen: Gone in next push.

pkg/console/subresource/configmap/configmap_test.go
benjaminapetersen: Considering if i should parse it into [the actual struct](https://github.com/openshift/console/blob/master/cmd/bridge/config.go#L13):

```golang
// https://github.com/openshift/console/blob/master/cmd/bridge/config.go#L13
// Config is the top-level console configuration.
type Config struct {
	APIVersion    string `yaml:"apiVersion"`
	Kind          string `yaml:"kind"`
	ServingInfo   `yaml:"servingInfo"`
	ClusterInfo   `yaml:"clusterInfo"`
	Auth          `yaml:"auth"`
	Customization `yaml:"customization"`
}

// ServingInfo holds configuration for serving HTTP.
type ServingInfo struct {
	BindAddress string `yaml:"bindAddress"`
	CertFile    string `yaml:"certFile"`
	KeyFile     string `yaml:"keyFile"`

	// These fields are defined in `HTTPServingInfo`, but are not supported for console. Fail if any are specified.
	// https://github.com/openshift/api/blob/0cb4131a7636e1ada6b2769edc9118f0fe6844c8/config/v1/types.go#L7-L38
	BindNetwork           string        `yaml:"bindNetwork"`
	ClientCA              string        `yaml:"clientCA"`
	NamedCertificates     []interface{} `yaml:"namedCertificates"`
	MinTLSVersion         string        `yaml:"minTLSVersion"`
	CipherSuites          []string      `yaml:"cipherSuites"`
	MaxRequestsInFlight   int64         `yaml:"maxRequestsInFlight"`
	RequestTimeoutSeconds int64         `yaml:"requestTimeoutSeconds"`
}

// ClusterInfo holds information the about the cluster such as master public URL and console public URL.
type ClusterInfo struct {
	ConsoleBaseAddress string `yaml:"consoleBaseAddress"`
	ConsoleBasePath    string `yaml:"consoleBasePath"`
	MasterPublicURL    string `yaml:"masterPublicURL"`
}

// Auth holds configuration for authenticating with OpenShift. The auth method is assumed to be "openshift".
type Auth struct {
	ClientID            string `yaml:"clientID"`
	ClientSecretFile    string `yaml:"clientSecretFile"`
	OAuthEndpointCAFile string `yaml:"oauthEndpointCAFile"`
	LogoutRedirect      string `yaml:"logoutRedirect"`
}

// Customization holds configuration such as what logo to use.
type Customization struct {
	Branding             string `yaml:"branding"`
	DocumentationBaseURL string `yaml:"documentationBaseURL"`
}
```
and compare the objects then.  But that would require pulling the `type Config` from `Console` repo.

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -105,14 +108,49 @@ func TestDefaultConfigMap(t *testing.T) {
...
+			var actualConfig map[string]interface{}
+			// convert the string back into a []byte
+			configBytes := []byte(cm.Data[configKey])
+
+			err = yaml.Unmarshal(configBytes, &actualConfig)
+			if err != nil {
+				t.Error("Problem with consoleConfig.Data[console-config.yaml]", err)
+			}
+
+			// compare the configs
```
benjaminapetersen: Fixed the test to marshall the configs into `map[string]interface{}` for comparison.  There is still the `nil` setting later to `deep.Equal()` test the rest of the struct, but this ensures everything is properly tested.

pkg/console/starter/starter.go
```diff
@@ -175,6 +177,8 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		ctx.EventRecorder,
 	)
 
+	configUpgradeableController := unsupportedconfigoverridescontroller.NewUnsupportedConfigOverridesController(operatorClient, ctx.EventRecorder)
```
benjaminapetersen: This controller handles reporting our status now, specifically `Upgradeable`, example:

```
status:
  conditions:
  - lastTransitionTime: 2019-03-26T19:31:44Z
    reason: AsExpected
    status: "False"
    type: Failing
  - lastTransitionTime: 2019-03-26T19:37:15Z
    message: 'Progressing: Moving to version 4.0.0'
    reason: Progressing
    status: "True"
    type: Progressing
  - lastTransitionTime: 2019-03-26T19:31:54Z
    reason: AsExpected
    status: "True"
    type: Available
  - lastTransitionTime: 2019-03-26T19:45:24Z
    message: 'Upgradeable: With UnsupportedConfigOverrides: {"auth":{"clientID":"samwisegamgee","logoutRedirect":"https://samwisegamgee.com/you-re-nowt-but-a-ninnyhammer-Sam-Gamgee"},"customization":{"branding":"samwisegamgee"}}'
    reason: AsExpected
    status: "True"
    type: Upgradeable
```

pkg/console/subresource/configmap/configmap_test.go
benjaminapetersen: Nope. Solved another way.

benjaminapetersen:
/retest

`98% complete: timed out waiting for the condition"`.  almost.

pkg/console/subresource/configmap/configmap.go
```diff
@@ -57,22 +61,49 @@ func getApiUrl(infrastructureConfig *configv1.Infrastructure) string {
...
+	// if we actually merged config overrides, log this information
+	didMerge := len(operatorConfig.Spec.UnsupportedConfigOverrides.Raw) != 0
+	if didMerge {
+		logrus.Println(fmt.Sprintf("With UnsupportedConfigOverrides: %v", string(unsupportedRaw)))
+	}
+
+	configMap.Data[consoleConfigYamlFile] = string(outConfigYaml)
+
+	logrus.Println("generated console config yaml:")
+	logrus.Printf("%s", string(outConfigYaml))
```
jhadvig: dont we need a newline ?
```suggestion
	logrus.Printf("%s \n", string(outConfigYaml))
```

benjaminapetersen: Yup

pkg/console/subresource/configmap/configmap.go
```diff
@@ -57,22 +61,49 @@ func getApiUrl(infrastructureConfig *configv1.Infrastructure) string {
...
 	configMap := Stub()
-	configMap.Data = map[string]string{
-		consoleConfigYamlFile: config,
+	configMap.Data = map[string]string{}
+	unsupportedRaw := operatorConfig.Spec.UnsupportedConfigOverrides.Raw
+
+	// merge config overrides, if we have them
+	mergedConfig, err := resourcemerge.MergeProcessConfig(nil, config, unsupportedRaw)
+	if err != nil {
+		logrus.Errorf("failed to generate configmap: %v \n", err)
```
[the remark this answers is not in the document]
benjaminapetersen: If we were in a loop doing this many times, I would agree, but there is really not a ton of gain here since we do this once per sync loop at most.  I'm leaning towards not wanting to add additional `if` statements in `golang` `if` statment happy land....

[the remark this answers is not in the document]
benjaminapetersen: `failed to generate configmap` should be correct because we are `return nil, false, err`.

If the `UnsupportedConfigOverrides` are provided, but incorrect, we die, and that will bubble up and a message to fix the overrides.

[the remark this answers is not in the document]
benjaminapetersen: (meaning we won't generate because the function exits early)

benjaminapetersen:
wake up prow :)

pkg/console/operator/status.go
jhadvig: hmm so I don't see this reason is used anywhere :-/

benjaminapetersen: I think since I removed the manual handling of `status` for the override & instead am using `NewUnsupportedConfigOverridesController()`, its now unused.  However, it should be here and we should use it for statuses if all is well.

benjaminapetersen: I'd prefer not to restart the test process over that, might be another 2 weeks till we pass 😄 😭

benjaminapetersen:
/retest 

benjaminapetersen:
/retest

`release "release-latest" failed: pod release-latest was already deleted`  (again)

benjaminapetersen:
yippee CI success 😈

PR #164, opened 2019-02-28
Revert "Validate URLs from the CR file and additional unit tests for ocp tag"

benjaminapetersen:
Reverts openshift/console-operator#152

There are a few items here that I missed on review that should probably be tweaked before we keep.  Reverting to update & re-test manually.


benjaminapetersen:
Cancelling in favor of #165

PR #168, opened 2019-03-07
Report Operand version

jhadvig:
First work based on discussion with @benjaminapetersen and @enj 
Please take a look and let me know whats missing besides the watching the CM (and if we actually need it...)
Thanks

### DIFF
--- a/manifests/05-version-mapping.yaml
+++ b/manifests/05-version-mapping.yaml
@@ -1,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: openshift-console-operator
+  name: version-mapping
+data:
+  "0.0.1-snapshot_openshift": "docker.io/openshift/origin-console:v4.0"
+  "0.0.1-snapshot": "docker.io/openshift/origin-console-operator:v4.0"
--- a/manifests/05-version-mapping.yaml
+++ b/manifests/05-version-mapping.yaml
@@ -4,5 +4,5 @@ metadata:
   namespace: openshift-console-operator
   name: version-mapping
 data:
-  "0.0.1-snapshot_openshift": "docker.io/openshift/origin-console:v4.0"
+  # "0.0.1-snapshot_openshift": "docker.io/openshift/origin-console:v4.0"
   "0.0.1-snapshot": "docker.io/openshift/origin-console-operator:v4.0"
--- a/manifests/05-version-mapping.yaml
+++ b/manifests/05-version-mapping.yaml
@@ -4,5 +4,4 @@ metadata:
   namespace: openshift-console-operator
   name: version-mapping
 data:
-  # "0.0.1-snapshot_openshift": "docker.io/openshift/origin-console:v4.0"
   "0.0.1-snapshot": "docker.io/openshift/origin-console-operator:v4.0"
--- a/manifests/07-operator.yaml
+++ b/manifests/07-operator.yaml
@@ -42,9 +42,7 @@ spec:
         env:
         - name: IMAGE
           value: docker.io/openshift/origin-console:latest
-        - name: OPERATOR_IMAGE
-          value: docker.io/openshift/origin-console-operator:v4.0
-        - name: OPERATOR_IMAGE_VERSION
+        - name: RELEASE_VERSION
           value: "0.0.1-snapshot"
         - name: OPERATOR_NAME
           value: "console-operator"
--- a/manifests/07-operator.yaml
+++ b/manifests/07-operator.yaml
@@ -42,6 +42,10 @@ spec:
         env:
         - name: IMAGE
           value: docker.io/openshift/origin-console:latest
+        - name: OPERATOR_IMAGE
+          value: docker.io/openshift/origin-console-operator:v4.0
+        - name: OPERATOR_IMAGE_VERSION
+          value: "0.0.1-snapshot"
         - name: OPERATOR_NAME
           value: "console-operator"
         - name: POD_NAME
--- a/manifests/08-clusteroperator.yaml
+++ b/manifests/08-clusteroperator.yaml
@@ -7,5 +7,3 @@ status:
   versions:
     - name: operator
       version: "0.0.1-snapshot"
-    - name: openshift-console
-      version: "0.0.1-snapshot_openshift"
--- a/manifests/08-clusteroperator.yaml
+++ b/manifests/08-clusteroperator.yaml
@@ -3,3 +3,9 @@ kind: ClusterOperator
 metadata:
   name: console
 spec: {}
+status:
+  versions:
+    - name: operator
+      version: "0.0.1-snapshot"
+    - name: openshift-console
+      version: "0.0.1-snapshot_openshift"
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -3,6 +3,7 @@ package operator
 import (
 	"errors"
 	"fmt"
+	"os"
 	"reflect"
 
 	"github.com/openshift/console-operator/pkg/console/subresource/util"
@@ -116,6 +117,16 @@ func sync_v400(co *consoleOperator, originalOperatorConfig *operatorv1.Console, 
 	// available is currently defined as "met the users intent"
 	if deploymentsub.IsReady(actualDeployment) && routesub.IsAdmitted(rt) {
 		co.ConditionDeploymentAvailable(operatorConfig)
+
+		if deploymentsub.IsReadyAndUpdated(actualDeployment) {
+			co.ConditionNotProgressing(operatorConfig)
+			version := os.Getenv("RELEASE_VERSION")
+			if co.versionGetter.GetVersions()["operator"] != version {
+				co.versionGetter.SetVersion("operator", version)
+			}
+		} else {
+			co.ConditionProgressing(operatorConfig)
+		}
 	} else {
 		co.ConditionDeploymentNotAvailable(operatorConfig)
 	}
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"strings"
 
 	"github.com/openshift/console-operator/pkg/console/subresource/util"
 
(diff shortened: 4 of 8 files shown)


manifests/05-version-mapping.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: openshift-console-operator
+  name: version-mapping
+data:
+  "0.0.0_version_console": "docker.io/openshift/origin-console:v4.0"
```
benjaminapetersen: Just confirming on your dev cluster these values did get replaced by CVO.

manifests/07-operator.yaml
```diff
@@ -42,6 +42,10 @@ spec:
         env:
         - name: IMAGE
           value: docker.io/openshift/origin-console:latest
+        - name: OPERATOR_IMAGE
+          value: docker.io/openshift/origin-console-operator:v4.0
+        - name: OPERATOR_IMAGE_VERSION
```
benjaminapetersen: Did we know if we also need `OPERAND_IMAGE_VERSION`? I assume not, just asking.

jhadvig: They are not using this envvar anywhere else, so I assume we dont 👍

manifests/05-version-mapping.yaml
benjaminapetersen: So we expect this to be updated to something like `image: quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:<hash>`.

benjaminapetersen: Commented on the side we could write an e2e test to check this, basically just GET the CM & look for something similar to that string, as a validation that its working.  

I'm not sure if thats the best approach or not. But CVO won't know about it if we just `oc apply -f 8  manifests/05-version-mapping.yaml` in a dev cluster, so its not going to get updated, thus makes it hard to test before merging other than using CI to validate.

manifests/07-operator.yaml
benjaminapetersen: neat :)

manifests/05-version-mapping.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: openshift-console-operator
```
benjaminapetersen: When we were discussing the need to `WATCH` this configmap, we contemplated if it needs to be in `openshift-console-operator`, which means we have to setup a bunch of new informers in our `starter.go`.  If we put it in `openshift-console`, we can just add an additional name to `tweakListOptionsForConfigs func`, which would be a lot less of a hassle.    

I'm not sure yet if we need to watch it, looking around.

jhadvig:
@benjaminapetersen  I've updated the PR based on changed in:
- [cluster-openshift-apiserver-operator](https://github.com/openshift/cluster-openshift-apiserver-operator/pull/161/files)
- [cluster-kube-apiserver-operator](https://github.com/openshift/cluster-kube-apiserver-operator/pull/308/files)
...

It appears that the operant key is changed to `x.x.x-snapshot-openshift` and the operator with `x.x.x-snapshot``

Also talked with @mfojtik regarding having the operand versions to be interpolated by pullSpec, and he confirmed that it wont, that only the `x.x.x-snapshot_openshift` string should be displayed.

Also bumped the version to `0.0.1` to align with other operators

manifests/08-clusteroperator.yaml
```diff
@@ -3,3 +3,9 @@ kind: ClusterOperator
 metadata:
   name: console
 spec: {}
+status:
+  versions:
+    - name: operator
```
benjaminapetersen: What does `snapshot` and `snapshot_openshift` mean?
Should it be `snapshot-console` and` snapshot-console-operator`?  Guessing no,just curious the significance.

Looking at [kube-apiserver-operator](https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/manifests/0000_10_kube-apiserver-operator_07_clusteroperator.yaml) quick

manifests/05-version-mapping.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
```
benjaminapetersen: Does this need to be added to `relatedObjects` on our `cluster-operator`?

benjaminapetersen: Nope, we are watching the whole namespace, thats sufficient.

benjaminapetersen:
/lgtm

thanks!

pkg/console/starter/starter.go
```diff
@@ -152,6 +153,9 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		recorder,
 	)
 
+	versionRecorder := status.NewVersionGetter()
+	versionRecorder.SetVersion("operator", os.Getenv("OPERATOR_IMAGE_VERSION"))
```
spadgett: @jhadvig where is the console version itself getting recorded?

spadgett:
> level=fatal msg="failed to initialize the cluster: Cluster operator console is still updating: missing version information for openshift-console"

jhadvig:
Adding the console version to see what jenkins will tell me. Still about to add the pod check.

spadgett:
@jhadvig I just chatted with @derekwaynecarr 

* We don't need to worry about reporting console version yet
* We should read the operator version from the `RELEASE_VERSION` env var
* We should not update the current version until all operator pods are updated to that version

I'd try to model ours after https://github.com/openshift/cluster-image-registry-operator/pull/225

manifests/08-clusteroperator.yaml
```diff
@@ -3,3 +3,9 @@ kind: ClusterOperator
...
+      version: "0.0.1-snapshot"
+    - name: openshift-console
+      version: "0.0.1-snapshot_openshift"
```
spadgett: This would need to come out of the YAML manifest if we're not yet reporting console version

pkg/console/operator/sync_v400.go
```diff
@@ -116,6 +117,16 @@ func sync_v400(co *consoleOperator, originalOperatorConfig *operatorv1.Console,
...
 		co.ConditionDeploymentAvailable(operatorConfig)
+
+		if deploymentsub.IsReadyAndUpdated(actualDeployment) {
+			co.ConditionNotProgressing(operatorConfig)
+			version := os.Getenv("RELEASE_VERSION")
+			if co.versionGetter.GetVersions()["operator"] != version {
+				co.versionGetter.SetVersion("operator", version)
+			}
+		} else {
+			co.ConditionProgressing(operatorConfig)
```
spadgett: It looks like we won't report progressing until at least one replica is ready?

We might need to pull this outside of the `if deploymentsub.IsReady(actualDeployment) && routesub.IsAdmitted(rt) {` block

spadgett: Do we need to remove or update the lines above that deal with progressing? (lines 111-114)

benjaminapetersen: Yeah, I don't think that nesting is quite right.  We can be progressing no matter how many pods we have.  As far as I understand, we are progressing until anything we manage stops changing.

pkg/console/subresource/deployment/deployment.go
```diff
@@ -259,3 +259,9 @@ func IsReady(deployment *appsv1.Deployment) bool {
 	logrus.Printf("Deployment is avalable: %v \n", deployment.Status.ReadyReplicas >= 1)
 	return deployment.Status.ReadyReplicas >= 1
 }
+
```
benjaminapetersen: Test! (but should be easy like `TestIsReady()`)

pkg/console/subresource/deployment/deployment.go
```diff
@@ -259,3 +259,9 @@ func IsReady(deployment *appsv1.Deployment) bool {
...
+func IsReadyAndUpdated(deployment *appsv1.Deployment) bool {
```
benjaminapetersen: Should this be `IsAvailableAndUpdated()`?  Or `IsExpectedGeneration()`?  Naming is hard, I dunno.

`if IsUpdating() { `  vs `if !IsUpdating() {`?

pkg/console/operator/sync_v400.go
jhadvig: True, mainly wanted to see by the additional commit if jenskins will error out, but seems we are on the right path

pkg/console/subresource/deployment/deployment.go
jhadvig: +1 for `IsAvailableAndUpdated()`

pkg/console/operator/sync_v400.go
```diff
@@ -106,12 +108,21 @@ func sync_v400(co *consoleOperator, originalOperatorConfig *operatorv1.Console,
...
-	// or if we have not settled on the desired number of replicas
-	if toUpdate || actualDeployment.Status.ReadyReplicas != deploymentsub.ConsoleReplicas {
+	// or if we have not settled on the desired number of replicas or deployment is not uptodate.
+	if toUpdate {
 		co.ConditionResourceSyncProgressing(operatorConfig, "Changes made during sync updates, additional sync expected.")
 	} else {
-		co.ConditionResourceSyncNotProgressing(operatorConfig)
+		version := os.Getenv("RELEASE_VERSION")
+		if !deploymentsub.IsAvailableAndUpdated(actualDeployment) {
+			co.ConditionResourceSyncProgressing(operatorConfig, fmt.Sprintf("Moving to version %s", strings.Split(version, "-")[0]))
```
jhadvig: Here and below I wasn't sure if we want to use the `ConditionResourceSyncProgressing()` method which sets the *reason* to "SyncLoopProgressing", but it makes sense to set a single *reason* I guess.

jhadvig:
@spadgett @benjaminapetersen I've updated the PR based on your comments + left bunch of comments in changes.
PTAL

pkg/console/operator/sync_v400.go
```diff
@@ -106,12 +108,21 @@ func sync_v400(co *consoleOperator, originalOperatorConfig *operatorv1.Console,
 	// but we will handle the state of the operand below
 	co.ConditionResourceSyncSuccess(operatorConfig)
 	// the operand is in a transitional state if any of the above resources changed
-	// or if we have not settled on the desired number of replicas
-	if toUpdate || actualDeployment.Status.ReadyReplicas != deploymentsub.ConsoleReplicas {
+	// or if we have not settled on the desired number of replicas or deployment is not uptodate.
+	if toUpdate {
```
jhadvig: Here I'm setting and unsetting the Progressing Status.
If any of the resources is syncing set the `ConditionResourceSyncProgressing`, else if requirements  for updating the version are not fulfilled, set the `ConditionResourceSyncProgressing` with the message mentioned in Jessica's mail. Otherwise just set the version.

benjaminapetersen:
Lol we should have squashed this :)

Commits:
```
- Report Operand version
- 1
- 2
```
The `1` and `2` are a little silly in our history.

PR #170, opened 2019-03-08
Add test for route IsAdmitted()

benjaminapetersen:
This adds tests for a variety of possible scenerios for the new `IsAdmitted()` func for `routes`.

This is important for correctly determining `Status.Condition.Available`.

/assign @spadgett @zherman0


benjaminapetersen:
`go fmt` updated

PR #171, opened 2019-03-08
Add unit tests for deployment IsReady() func

benjaminapetersen:
/assign @zherman0 @spadgett 

This is trivial, but I think we want certainty around correct reporting of Ready for `Status.Condition.Available` so it's worth a test.

### DIFF
--- a/pkg/console/subresource/deployment/deployment_test.go
+++ b/pkg/console/subresource/deployment/deployment_test.go
@@ -1,8 +1,9 @@
 package deployment
 
 import (
-	"github.com/go-test/deep"
 	"testing"
+
+	"github.com/go-test/deep"
 
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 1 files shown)


pkg/console/subresource/deployment/deployment_test.go
```diff
@@ -55,12 +56,12 @@ func TestDefaultDeployment(t *testing.T) {
...
-						OwnerReferences: nil,
-						Initializers:    nil,
-						Finalizers:      nil,
-						ClusterName:     "",
+						Labels:                     labels,
+						Annotations:                nil,
+						OwnerReferences:            nil,
+						Initializers:               nil,
+						Finalizers:                 nil,
+						ClusterName:                "",
```
zherman0: Do we want all these new spaces here?

benjaminapetersen:
`go fmt` updated.

PR #173, opened 2019-03-12
Fix missing masterPublicURL on console configmap

benjaminapetersen:
bug 1664574
([link](https://bugzilla.redhat.com/show_bug.cgi?id=1664574))
- Add `infrastructureConfig` to supply console `configmap` with `masterPublicURL`.

This will likely conflict with #162

### DIFF
--- a/manifests/03-rbac-role.yaml
+++ b/manifests/03-rbac-role.yaml
@@ -125,6 +125,14 @@ rules:
 - apiGroups:
   - config.openshift.io
   resources:
+  - infrastructures
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - config.openshift.io
+  resources:
   - consoles
   - consoles/status
   - clusteroperators
--- a/pkg/console/operator/operator.go
+++ b/pkg/console/operator/operator.go
@@ -27,7 +27,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/events"
 
 	// informers
-	configinformerv1 "github.com/openshift/client-go/config/informers/externalversions/config/v1"
+	configinformer "github.com/openshift/client-go/config/informers/externalversions"
 	operatorinformerv1 "github.com/openshift/client-go/operator/informers/externalversions/operator/v1"
 
 	routesinformersv1 "github.com/openshift/client-go/route/informers/externalversions/route/v1"
@@ -57,14 +57,14 @@ type consoleOperator struct {
 	operatorConfigClient operatorclientv1.ConsoleInterface
 	consoleConfigClient  configclientv1.ConsoleInterface
 	// core kube
-
 	secretsClient    coreclientv1.SecretsGetter
 	configMapClient  coreclientv1.ConfigMapsGetter
 	serviceClient    coreclientv1.ServicesGetter
 	deploymentClient appsv1.DeploymentsGetter
 	// openshift
-	routeClient routeclientv1.RoutesGetter
-	oauthClient oauthclientv1.OAuthClientsGetter
+	routeClient          routeclientv1.RoutesGetter
+	oauthClient          oauthclientv1.OAuthClientsGetter
+	infrastructureClient configclientv1.InfrastructureInterface
 	// recorder
 	recorder events.Recorder
 }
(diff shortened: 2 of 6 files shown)


benjaminapetersen:
Testing yet, my cluster fell over.

spadgett:
CI failures look legitimate

benjaminapetersen:
rbac updated

pkg/console/operator/operator.go
```diff
@@ -27,7 +27,8 @@ import (
 	"github.com/openshift/library-go/pkg/operator/events"
 
 	// informers
-	configinformerv1 "github.com/openshift/client-go/config/informers/externalversions/config/v1"
+	configinformer "github.com/openshift/client-go/config/informers/externalversions"
+	// configinformerv1 "github.com/openshift/client-go/config/informers/externalversions/config/v1"
```
spadgett: Remove commented out code

manifests/03-rbac-role.yaml
```diff
@@ -122,6 +122,14 @@ rules:
   - watch
   resourceNames:
   - console
+- apiGroups:
+  - config.openshift.io
+  resources:
+  - infrastructures
```
spadgett: I'd combine this with the existing rule for api group `config.openshift.io`

pkg/console/operator/operator.go
```diff
@@ -148,15 +162,22 @@ func (c *consoleOperator) Sync(obj metav1.Object) error {
 		}
 	}
 
-	if err := c.handleSync(operatorConfig, consoleConfig); err != nil {
+	// we need infrastructure config for apiServerURL.
+	// it is helpful but not essential so we should not die if its not available.
```
spadgett: Is there ever a time we'd expect this not to work? I feel like we should just fail here. Public URL is needed for a few things in console.

benjaminapetersen: Well, I figure we need it only for a few small things, that we can guard in console code. So console can be 99% useful without it. Once it's available, the new pods will roll out anyway.

spadgett: It should always be available right? The drawback is that we're not reporting an error anywhere (except in the operator log). You won't know something is broken.

manifests/03-rbac-role.yaml
benjaminapetersen: We don't want to `update`, `delete`, `create`. 
Just watch rules.

pkg/console/operator/operator.go
benjaminapetersen: Ill update.

pkg/console/operator/operator.go
benjaminapetersen: will do.

manifests/03-rbac-role.yaml
spadgett: Thanks, missed that. Looks good.

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -61,6 +65,11 @@ func TestDefaultConfigMap(t *testing.T) {
 					Spec:   configv1.ConsoleSpec{},
 					Status: configv1.ConsoleStatus{},
 				},
+				infrastructureConfig: &configv1.Infrastructure{
+					Status: configv1.InfrastructureStatus{
+						APIServerURL: "https://api.some.cluster.openshift.com:6443",
```
spadgett: should probably use the const

pkg/console/starter/starter.go
```diff
@@ -127,23 +127,23 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
...
 	}
 
 	// TODO: rearrange these into informer,client pairs, NOT separated.
 	consoleOperator := operator.NewConsoleOperator(
 		// informers
-		consoleOperatorConfigInformers.Operator().V1().Consoles(), // OperatorConfig
-		consoleConfigInformers.Config().V1().Consoles(),           // ConsoleConfig
-
+		operatorConfigInformers.Operator().V1().Consoles(), // OperatorConfig
+		// configInformers.Config().V1().Consoles(),           // ConsoleConfig
```
spadgett: remove commented out code?

pkg/console/subresource/configmap/configmap_test.go
benjaminapetersen: missed that, will follow-up.

PR #177, opened 2019-03-16
manifests: Serve oc from within the cluster

wking:
We need to support the ability to download clients (`oc`) from the cluster when an install is successful.  With this commit, I'm adding a new Deployment for serving the binary from the cli image referenced
from the release image.  That way, the provided `oc` is the appropriate version for the current cluster version.  I've set `replicas` to two to support continued availability during rolling updates.

The Python here-document approach is based on openshift/ci-operator@95a41a39 (openshift/ci-operator#35).  The script creates a scratch directory to serve from, and populates it with the `oc` binary (via a symlink).  It also creates zip and tar versions, so users can do things like:

```console
$ wget https://downloads-openshift-console.apps.wking.devcluster.openshift.com/amd64/linux/oc.zip
$ unzip oc.zip
```

instead of:

```console
$ wget https://downloads-openshift-console.apps.wking.devcluster.openshift.com/amd64/linux/oc
$ chmod +x oc
```

if they prefer.  We don't have to clean up the scratch directory in the script, because it will be reaped as part of container cleanup.

Moving forward, we'll want to replace the cli image with a cli-downloads image or similar that includes builds for other platforms and architectures, but I'm punting on that for now.

[1]: https://github.com/openshift/ci-operator/commit/95a41a395fa778680b3dc93c845407eb8a1a0ae1

### DIFF
--- a/manifests/07-downloads-deployment.yaml
+++ b/manifests/07-downloads-deployment.yaml
@@ -1,0 +1,84 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  namespace: openshift-console
+  name: downloads
+spec:
+  replicas: 2
+  selector:
+    matchLabels:
+      app: console
+      component: downloads
+  template:
+    metadata:
+      name: downloads
+      labels:
+        app: console
+        component: downloads
+    spec:
+      tolerations:
+      - operator: Exists
+      containers:
+      - name: download-server
+        image: docker.io/openshift/origin-cli:latest
+        imagePullPolicy: IfNotPresent
+        ports:
+        - containerPort: 8080
+          name: http
+          protocol: TCP
+        command: ["/bin/sh"]
+        args:
+        - '-c'
+        - |
+          cat <<EOF >>/tmp/serve.py
+          import BaseHTTPServer, os, re, SimpleHTTPServer, socket, tarfile, tempfile, threading, time, zipfile
+
+          # Launch multiple listeners as threads
+          class Thread(threading.Thread):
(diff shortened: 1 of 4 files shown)


benjaminapetersen:
@spadgett fyi

benjaminapetersen:
/lgtm

heredoc will always bother me but I can trust its the right call. 😄

benjaminapetersen:
Contemplating if I should follow and split `/manifests` into 2 subdirs (I don't know if there will ever be a case for additional items in the future.)

wking:
> Contemplating if I should follow and split `/manifests` into 2 subdirs...

Does the CVO walk subdirs?

>  I don't know if there will ever be a case for additional items in the future.

Maybe if we can't find a way to serve all downloads from a single container?

benjaminapetersen:
@wking I'm looking for info about subdirs now. I see at least one operator with subdirs, but I dont know if CVO manages it.  If not, perhaps I'll do a `prefix-`.  

I'm guessing we may also get `extensions` in the future.

wking:
[e2e-aws][1]:

```
level=error msg="\t* aws_vpc.new_vpc: Error creating VPC: VpcLimitExceeded: The maximum number of VPCs has been reached."
```

I'll run my reapers.

[1]: https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_console-operator/177/pull-ci-openshift-console-operator-master-e2e-aws/1081

benjaminapetersen:
🎉

PR #178, opened 2019-03-18
Bug 1689773: Set proper status when console route is not admitted

jhadvig:
Comments in code 👨‍💻 

/assign @benjaminapetersen

### DIFF
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -23,6 +23,7 @@ import (
 
 	// openshift
 	configv1 "github.com/openshift/api/config/v1"
+	operatorsv1 "github.com/openshift/api/operator/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/console-operator/pkg/crypto"
 	"github.com/openshift/library-go/pkg/operator/events"
@@ -123,12 +124,23 @@ func sync_v400(co *consoleOperator, originalOperatorConfig *operatorv1.Console, 
 		}
 	}
 
-	// the operand is available if all resources are present & if we have all the replicas
+	// the operand is available if all resources are:
+	// - present
+	// - if we have at least one ready replica
+	// - route is admitted
 	// available is currently defined as "met the users intent"
-	if deploymentsub.IsReady(actualDeployment) && routesub.IsAdmitted(rt) {
-		co.ConditionDeploymentAvailable(operatorConfig)
-	} else {
+	if !deploymentsub.IsReady(actualDeployment) {
 		co.ConditionDeploymentNotAvailable(operatorConfig)
+	} else if !routesub.IsAdmitted(rt) {
+		co.SetStatusCondition(
+			operatorConfig,
+			operatorsv1.OperatorStatusTypeAvailable,
+			operatorv1.ConditionFalse,
+			"RouteNotAdmitted",
+			"Console route is not admitted",
+		)
+	} else {
+		co.ConditionDeploymentAvailable(operatorConfig)
 	}
 
 	// finally write out the set of conditions currently set if anything has changed


pkg/console/operator/sync_v400.go
```diff
@@ -123,10 +124,23 @@ func sync_v400(co *consoleOperator, originalOperatorConfig *operatorv1.Console,
...
+	// - if we have all the replicas
+	// - route is admitted
 	// available is currently defined as "met the users intent"
-	if deploymentsub.IsReady(actualDeployment) && routesub.IsAdmitted(rt) {
-		co.ConditionDeploymentAvailable(operatorConfig)
+	if deploymentsub.IsReady(actualDeployment) {
+		if routesub.IsAdmitted(rt) {
+			co.ConditionDeploymentAvailable(operatorConfig)
+		} else {
+			co.SetStatusCondition(
```
jhadvig: Here I wasn't sure if we wanna add additional method to the `status.go`, or reuse those that already exist, since we already have quite a lot of them, and don't think that creating a new one for a single usage makes sense.

spadgett: > Here I wasn't sure if we wanna add additional method to the status.go

Yeah, I'd actually prefer removing the existing ones. I think the code is easier to understand with the condition inline like you have here.

pkg/console/operator/sync_v400.go
```diff
@@ -123,10 +124,23 @@ func sync_v400(co *consoleOperator, originalOperatorConfig *operatorv1.Console,
...
 
-	// the operand is available if all resources are present & if we have all the replicas
+	// the operand is available if all resources are:
+	// - present
+	// - if we have all the replicas
+	// - route is admitted
 	// available is currently defined as "met the users intent"
-	if deploymentsub.IsReady(actualDeployment) && routesub.IsAdmitted(rt) {
-		co.ConditionDeploymentAvailable(operatorConfig)
+	if deploymentsub.IsReady(actualDeployment) {
```
spadgett: nit: I have a slight preference to avoid the nesting

```go
if !deploymentsub.IsReady(actualDeployment) {
  co.ConditionDeploymentNotAvailable(operatorConfig)
  co.SetStatusCondition(...)
} else {
  co.ConditionDeploymentAvailable(operatorConfig)
}
```

pkg/console/operator/sync_v400.go
jhadvig: Sounds like a followUp to me 😄

jhadvig:
@spadgett comment addressed

pkg/console/operator/sync_v400.go
```diff
@@ -123,12 +124,23 @@ func sync_v400(co *consoleOperator, originalOperatorConfig *operatorv1.Console,
 		}
 	}
 
-	// the operand is available if all resources are present & if we have all the replicas
+	// the operand is available if all resources are:
+	// - present
+	// - if we have all the replicas
```
spadgett:
```suggestion
	// - if we have at least one ready replica
```

spadgett:
/lgtm cancel

@jhadvig It's an existing problem, but would you mind fixing the comment quickly?

jhadvig:
@spadgett comment updated

spadgett:
/lgtm

thanks @jhadvig !

pkg/console/operator/sync_v400.go
benjaminapetersen: I would like to eliminate them, but also remove all of the inline noise, and instead bubble up a collection of data needed & return from the `sync()` function & create our statuses in one place only.  Long term I don't like these helpers either, but I also don't really want to litter these calls everywhere.

pkg/console/operator/sync_v400.go
```diff
@@ -123,12 +124,23 @@ func sync_v400(co *consoleOperator, originalOperatorConfig *operatorv1.Console,
...
+	// - if we have at least one ready replica
+	// - route is admitted
 	// available is currently defined as "met the users intent"
-	if deploymentsub.IsReady(actualDeployment) && routesub.IsAdmitted(rt) {
-		co.ConditionDeploymentAvailable(operatorConfig)
-	} else {
+	if !deploymentsub.IsReady(actualDeployment) {
 		co.ConditionDeploymentNotAvailable(operatorConfig)
+	} else if !routesub.IsAdmitted(rt) {
+		co.SetStatusCondition(
```
benjaminapetersen: What will get tricky, is we could step on ourselves if we don't aggregate these somehow (but we do want to keep the fail early pattern of the sync loop).   We may have several places that set `OperatorStatusTypeAvailable`, but what if they set `True`, `False`, `True`?  But for different reasons?  Any `False` should result in `False` within a single `Sync` loop run, honestly.  (For `Available`, but not necessarily for the others, `Progressing` would be opposite, if one item is `True`, it should be `True`).

That's where I find the named functions easier to sort out because currently we have to keep these manually maintained.  😦

PR #179, opened 2019-03-19
Handle Console Operator LogLevel

jhadvig:
Comments in code

/assign @benjaminapetersen 

Story: https://jira.coreos.com/browse/CONSOLE-1320

### DIFF
--- a/pkg/console/operator/operator.go
+++ b/pkg/console/operator/operator.go
@@ -250,6 +250,7 @@ func (c *consoleOperator) defaultConsoleOperatorConfig() *operatorsv1.Console {
 		Spec: operatorsv1.ConsoleSpec{
 			OperatorSpec: operatorsv1.OperatorSpec{
 				ManagementState: operatorsv1.Managed,
+				LogLevel:        operatorsv1.Normal,
 			},
 		},
 	}
--- a/pkg/console/subresource/deployment/deployment.go
+++ b/pkg/console/subresource/deployment/deployment.go
@@ -1,6 +1,8 @@
 package deployment
 
 import (
+	"fmt"
+
 	"github.com/sirupsen/logrus"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -196,6 +198,22 @@ func consoleVolumeMounts(vc []volumeConfig) []corev1.VolumeMount {
 
 func consoleContainer(cr *operatorv1.Console) corev1.Container {
 	volumeMounts := consoleVolumeMounts(volumeConfigList)
+	level := ""
+	// Since the console-operator logging has different logging levels then the capnslog,
+	// that we use for console server(bridge) we need to map them to each other
+	switch cr.Spec.LogLevel {
+	case operatorv1.Normal:
+		level = "NOTICE"
+	case operatorv1.Debug:
+		level = "DEBUG"
+	case operatorv1.Trace:
+		fallthrough
+	case operatorv1.TraceAll:
+		fallthrough
+	default:
+		level = "TRACE"
+	}
+	flag := fmt.Sprintf("--log-level=*=%s", level)
 
 	return corev1.Container{
 		Image:           util.GetImageEnv(),
@@ -206,6 +224,7 @@ func consoleContainer(cr *operatorv1.Console) corev1.Container {
 			"--public-dir=/opt/bridge/static",
 			"--config=/var/console-config/console-config.yaml",
 			"--service-ca-file=/var/service-ca/service-ca.crt",
+			flag,
 		},
 		// TODO: can probably remove, this is used for local dev
 		//Env: []corev1.EnvVar{{
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -101,6 +101,9 @@ func sync_v400(co *consoleOperator, originalOperatorConfig *operatorv1.Console, 
 	}
 	toUpdate = toUpdate || depChanged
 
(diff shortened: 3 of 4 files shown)


pkg/console/subresource/deployment/deployment.go
```diff
@@ -206,6 +225,7 @@ func consoleContainer(cr *operatorv1.Console) corev1.Container {
 			"--public-dir=/opt/bridge/static",
 			"--config=/var/console-config/console-config.yaml",
 			"--service-ca-file=/var/service-ca/service-ca.crt",
+			flag,
```
jhadvig: Adding both, flag and argument(above), so the log-level is passed to the `bridge` binary

pkg/console/operator/operator.go
```diff
@@ -250,6 +250,7 @@ func (c *consoleOperator) defaultConsoleOperatorConfig() *operatorsv1.Console {
 		Spec: operatorsv1.ConsoleSpec{
 			OperatorSpec: operatorsv1.OperatorSpec{
 				ManagementState: operatorsv1.Managed,
+				LogLevel:        operatorsv1.Debug,
```
jhadvig: The log-level is set here

benjaminapetersen: So the `- "-v=4"` there is for the `operator` pod.

We want to pipe logging down into the `console` pod.

benjaminapetersen: So we will need to pipe the log level here:

https://github.com/openshift/console-operator/blob/master/pkg/console/subresource/deployment/deployment.go#L204

Which I imagine will require a bit of a translation function from `operatorsv1.Debug` to whatever `capnslog` wants for `Debug`.  They may be the same, maybe not.  But probably will be some kind of `--log-level=*=DEBUG` string creation.

pkg/console/subresource/deployment/deployment.go
```diff
@@ -196,6 +198,20 @@ func consoleVolumeMounts(vc []volumeConfig) []corev1.VolumeMount {
...
+	case operatorv1.Debug:
+		level = "DEBUG"
+	case operatorv1.Trace:
+		fallthrough
+	case operatorv1.TraceAll:
+		fallthrough
+	default:
+		level = "TRACE"
+	}
+	flag := fmt.Sprintf("--log-level=logrus=%s", level)
```
benjaminapetersen: `logrus?`

pkg/console/operator/operator.go
benjaminapetersen: OK this should work, though should we set it to `NORMAL` rather than `DEBUG`?

jhadvig: yep, just wanted to see what it will spit out :)

jhadvig: @benjaminapetersen updated

jhadvig:
@benjaminapetersen updated

pkg/console/subresource/deployment/deployment.go
```diff
@@ -206,6 +224,7 @@ func consoleContainer(cr *operatorv1.Console) corev1.Container {
 			"--public-dir=/opt/bridge/static",
 			"--config=/var/console-config/console-config.yaml",
 			"--service-ca-file=/var/service-ca/service-ca.crt",
+			flag,
```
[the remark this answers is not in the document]
benjaminapetersen: Ok so to ensure the pod template changes enough that `ApplyDeployment()` can notice, lets do this:

```yaml
  template:
    metadata:
      annotations:
        console.openshift.io/console-config-version: "1622686"
        console.openshift.io/image: docker.io/openshift/origin-console:latest
        console.openshift.io/oauth-secret-version: "10542"
        console.openshift.io/service-ca-config-version: "10545"
        # add a log level annotation to kick it
        console.openshift.io/log-level: TRACE
```
That will ensure that `modified=true` which will cause an update to roll out.

benjaminapetersen:
If we don't want to use the `annotation` that way (we are kind of abusing it) we can do something like:

```golang
// in func SyncDeployment()
deployment, _, err := resourceapply.ApplyDeployment(
        c.deployments,
        c.recorder,
        expectedDeployment,
        resourcemerge.ExpectedDeploymentGeneration(expectedDeployment, operatorConfig.Status.Generations),
        // redeploy on operatorConfig.spec changes
        operatorConfig.ObjectMeta.Generation != operatorConfig.Status.ObservedGeneration, 
    )

// so long as somewhere in sync_v400() we add 
operatorConfig.Status.ObservedGeneration = operatorConfig.ObjectMeta.Generation
// at the end of the sync loop before we update status. 
// generation increments on spec changes, so this would probably work better as we dont have to 
// do bookkeeping on the deployment itself to ensure it rolls out, we can just trust that 
// changes to our operator config will flip the boolean
```

That seems better.

pkg/console/operator/operator.go
benjaminapetersen: Sorry meant `Normal` should be the default right?
Even though the `console` `deployment` itself is really sparse on logs...

benjaminapetersen:
/retest

`failed to apply using Terraform`

benjaminapetersen:
lgtm so long as tests pass.

jhadvig:
```
fail [github.com/openshift/origin/test/extended/deployments/deployments.go:399]: Expected
    <string>: --> pre: Running hook pod ...
    --> pre: Success
    --> Scaling up deployment-test-2 from 0 to 1, scaling down deployment-test-1 from 0 to 0 (keep 1 pods available, don't exceed 2 pods)
        Scaling deployment-test-2 up to 1
    --> Success
to contain substring
    <string>: test pre hook executed
```

/retest

PR #180, opened 2019-03-19
Revise branding mechanism for Console Operator

zherman0:
Update the default behavior of branding so that the one and only docker file now builds `ocp` instead of `okd`.
If the user instead wants to manually build openshift with `go build` or `make build`, that process will default to `okd`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,7 @@
 FROM registry.svc.ci.openshift.org/openshift/release:golang-1.10 AS builder
-# create a work dir and copy local files in
 WORKDIR /go/src/github.com/openshift/console-operator
 COPY . .
-RUN make build
+RUN ADDITIONAL_GOTAGS="ocp" make build WHAT="cmd/console"
 
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
 RUN useradd console-operator
@@ -18,9 +17,4 @@ LABEL io.k8s.display-name="OpenShift console-operator" \
       maintainer="Benjamin A. Petersen <bpetersen@redhat.com>"
 
 LABEL io.openshift.release.operator true
-
-# entrypoint specified in 03-operator.yaml as `console-operator`
-# CMD ["/usr/bin/console", "operator", "--kubeconfig", "path/to/config", "--config", "./install/config.yaml", "--v", "4"]
-# CMD ["/usr/bin/console", "operator", "--v", "4"]
-
 
--- a/manifests/03-rbac-role-ns-openshift-config-managed.yaml
+++ b/manifests/03-rbac-role-ns-openshift-config-managed.yaml
@@ -1,0 +1,14 @@
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: console-operator
+  namespace: openshift-config-managed
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - list
+  - watch
--- a/manifests/04-rbac-rolebinding.yaml
+++ b/manifests/04-rbac-rolebinding.yaml
@@ -37,3 +37,17 @@ subjects:
 - kind: ServiceAccount
   name: console-operator
   namespace: openshift-console-operator
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: console-operator
+  namespace: openshift-config-managed
+roleRef:
+  kind: Role
+  name: console-operator
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+  - kind: ServiceAccount
+    name: console-operator
+    namespace: openshift-console-operator
--- a/pkg/console/operator/operator.go
+++ b/pkg/console/operator/operator.go
@@ -77,6 +77,7 @@ func NewConsoleOperator(
 	configInformer configinformer.SharedInformerFactory,
 
 	coreV1 corev1.Interface,
+	managedCoreV1 corev1.Interface,
 	deployments appsinformersv1.DeploymentInformer,
 	routes routesinformersv1.RouteInformer,
 	oauthClients oauthinformersv1.OAuthClientInformer,
@@ -113,6 +114,7 @@ func NewConsoleOperator(
 
 	secretsInformer := coreV1.Secrets()
 	configMapInformer := coreV1.ConfigMaps()
+	managedConfigMapInformer := managedCoreV1.ConfigMaps()
 	serviceInformer := coreV1.Services()
 	configV1Informers := configInformer.Config().V1()
 
@@ -131,6 +133,7 @@ func NewConsoleOperator(
 		operator.WithInformer(oauthClients, targetNameFilter),
 		// special resources with unique names
 		operator.WithInformer(configMapInformer, operator.FilterByNames(configmap.ConsoleConfigMapName, configmap.ServiceCAConfigMapName)),
+		operator.WithInformer(managedConfigMapInformer, operator.FilterByNames(configmap.ConsoleConfigMapName)),
 		operator.WithInformer(secretsInformer, operator.FilterByNames(deployment.ConsoleOauthConfigName)),
 	)
 }
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -7,10 +7,6 @@ import (
 	"reflect"
 	"strings"
 
-	"github.com/openshift/console-operator/pkg/console/subresource/util"
-
-	"github.com/openshift/console-operator/pkg/api"
-
 	// 3rd party
 	"github.com/sirupsen/logrus"
 	// kube
@@ -24,6 +20,8 @@ import (
 	// openshift
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/console-operator/pkg/api"
+	"github.com/openshift/console-operator/pkg/console/subresource/util"
 	"github.com/openshift/console-operator/pkg/crypto"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
@@ -249,11 +247,16 @@ func SyncSecret(co *consoleOperator, recorder events.Recorder, operatorConfig *o
 func SyncConfigMap(co *consoleOperator, recorder events.Recorder, operatorConfig *operatorv1.Console, consoleConfig *configv1.Console, infrastructureConfig *configv1.Infrastructure, rt *routev1.Route) (*corev1.ConfigMap, bool, error) {
 	logrus.Printf("validating console configmap...")
 
-	defaultConfigmap, _, err := configmapsub.DefaultConfigMap(operatorConfig, consoleConfig, infrastructureConfig, rt)
+	managedConfig, mcErr := co.configMapClient.ConfigMaps(api.OpenshiftConfigManagedNamespace).Get(api.OpenShiftConsoleConfigMapName, metav1.GetOptions{})
+	if mcErr != nil && !apierrors.IsNotFound(mcErr) {
+		logrus.Errorf("managed config error: %v \n", mcErr)
+		return nil, false, mcErr
+	}
+
+	defaultConfigmap, _, err := configmapsub.DefaultConfigMap(operatorConfig, consoleConfig, managedConfig, infrastructureConfig, rt)
 	if err != nil {
 		return nil, false, err
 	}
-
 	cm, cmChanged, cmErr := resourceapply.ApplyConfigMap(co.configMapClient, recorder, defaultConfigmap)
 	if cmErr != nil {
 		logrus.Errorf("%q: %v \n", "configmap", cmErr)
--- a/pkg/console/starter/starter.go
+++ b/pkg/console/starter/starter.go
@@ -83,6 +83,11 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		informers.WithNamespace(api.TargetNamespace),
 	)
 
+	kubeInformersManagedNamespaced := informers.NewSharedInformerFactoryWithOptions(
+		kubeClient,
+		resync,
+		informers.WithNamespace(api.OpenshiftConfigManagedNamespace),
+	)
 	// configs are all named "cluster", but our clusteroperator is named "console"
 	configInformers := configinformers.NewSharedInformerFactoryWithOptions(
 		configClient,
@@ -108,12 +113,12 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		oauthinformers.WithTweakListOptions(tweakListOptionsForOAuth),
 	)
 
-	recorder := ctx.EventRecorder
-
 	operatorClient := &operatorclient.OperatorClient{
 		Informers: operatorConfigInformers,
 		Client:    operatorConfigClient.OperatorV1(),
 	}
+
+	recorder := ctx.EventRecorder
 
 	versionGetter := status.NewVersionGetter()
 
@@ -123,6 +128,7 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		operatorConfigInformers.Operator().V1().Consoles(), // OperatorConfig
 		configInformers,                                    // ConsoleConfig
 		kubeInformersNamespaced.Core().V1(),                // Secrets, ConfigMaps, Service
+		kubeInformersManagedNamespaced.Core().V1(),         // Managed ConfigMaps
 		kubeInformersNamespaced.Apps().V1().Deployments(),  // Deployments
 		routesInformersNamespaced.Route().V1().Routes(),    // Route
 		oauthInformers.Oauth().V1().OAuthClients(),         // OAuth clients
@@ -163,11 +169,18 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 
 	configUpgradeableController := unsupportedconfigoverridescontroller.NewUnsupportedConfigOverridesController(operatorClient, ctx.EventRecorder)
 
-	kubeInformersNamespaced.Start(ctx.Done())
-	operatorConfigInformers.Start(ctx.Done())
-	configInformers.Start(ctx.Done())
-	routesInformersNamespaced.Start(ctx.Done())
-	oauthInformers.Start(ctx.Done())
+	for _, informer := range []interface {
+		Start(stopCh <-chan struct{})
+	}{
+		kubeInformersNamespaced,
+		kubeInformersManagedNamespaced,
+		operatorConfigInformers,
+		configInformers,
+		routesInformersNamespaced,
+		oauthInformers,
+	} {
+		informer.Start(ctx.Done())
+	}
 
 	go consoleOperator.Run(ctx.Done())
 	go clusterOperatorStatus.Run(1, ctx.Done())
--- a/pkg/console/subresource/configmap/configmap.go
+++ b/pkg/console/subresource/configmap/configmap.go
@@ -2,11 +2,10 @@ package configmap
 
 import (
 	"fmt"
-
 	"github.com/sirupsen/logrus"
 
 	yaml2 "github.com/ghodss/yaml"
-	yaml "gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v2"
 
 	corev1 "k8s.io/api/core/v1"
 
@@ -65,7 +64,7 @@ func getApiUrl(infrastructureConfig *configv1.Infrastructure) string {
 // - a new configmap,
 // - a bool indicating if config was merged (unsupportedConfigOverrides)
 // - an error
-func DefaultConfigMap(operatorConfig *operatorv1.Console, consoleConfig *configv1.Console, infrastructureConfig *configv1.Infrastructure, rt *routev1.Route) (*corev1.ConfigMap, bool, error) {
+func DefaultConfigMap(operatorConfig *operatorv1.Console, consoleConfig *configv1.Console, managedConfig *corev1.ConfigMap, infrastructureConfig *configv1.Infrastructure, rt *routev1.Route) (*corev1.ConfigMap, bool, error) {
 	logoutRedirect := getLogoutRedirect(consoleConfig)
 	brand := getBrand(operatorConfig)
 	docURL := getDocURL(operatorConfig)
@@ -77,11 +76,12 @@ func DefaultConfigMap(operatorConfig *operatorv1.Console, consoleConfig *configv
 	configMap := Stub()
 	configMap.Data = map[string]string{}
 	unsupportedRaw := operatorConfig.Spec.UnsupportedConfigOverrides.Raw
+	newConfig := extractYAML(managedConfig)
 
 	// merge config overrides, if we have them
-	mergedConfig, err := resourcemerge.MergeProcessConfig(nil, config, unsupportedRaw)
+	mergedConfig, err := resourcemerge.MergeProcessConfig(nil, config, newConfig, unsupportedRaw)
 	if err != nil {
-		logrus.Errorf("failed to generate configmap: %v \n", err)
+		logrus.Errorf("failed to merge configmap: %v \n", err)
 		return nil, false, err
 	}
 
@@ -192,4 +192,14 @@ func authServerYaml(logoutRedirect string) yaml.MapSlice {
 
 func consoleBaseAddr(host string) string {
 	return util.HTTPS(host)
+}
+
+// Helper function that pulls the yaml struct out of the data section of a configmap yaml
+func extractYAML(managedConfig *corev1.ConfigMap) []byte {
+	data := managedConfig.Data
+	for _, v := range data {
+		return []byte(v)
+	}
+
+	return []byte{}
 }
--- a/pkg/console/subresource/configmap/configmap_test.go
+++ b/pkg/console/subresource/configmap/configmap_test.go
@@ -40,6 +40,31 @@ servingInfo:
   certFile: /var/serving-cert/tls.crt
   keyFile: /var/serving-cert/tls.key
 `
+	exampleManagedConfigMapData = `kind: ConsoleConfig
+apiVersion: console.openshift.io/v1beta1
+customization:
+  branding: online
+  documentationBaseURL: https://docs.okd.io/4.0/
+`
+	exampleYamlWithManagedConfig = `kind: ConsoleConfig
+apiVersion: console.openshift.io/v1beta1
+auth:
+  clientID: console
+  clientSecretFile: /var/oauth-config/clientSecret
+  logoutRedirect: ""
+  oauthEndpointCAFile: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+clusterInfo:
+  consoleBaseAddress: https://` + host + `
+  consoleBasePath: ""
+  masterPublicURL: ` + mockAPIServer + `
+customization:
+  branding: online 
+  documentationBaseURL: https://docs.okd.io/4.0/
+servingInfo:
+  bindAddress: https://0.0.0.0:8443
+  certFile: /var/serving-cert/tls.crt
+  keyFile: /var/serving-cert/tls.key
+`
 )
 
 // To manually run these tests: go test -v ./pkg/console/subresource/configmap/...
@@ -47,6 +72,7 @@ func TestDefaultConfigMap(t *testing.T) {
 	type args struct {
 		operatorConfig       *operatorv1.Console
 		consoleConfig        *configv1.Console
+		managedConfig        *corev1.ConfigMap
 		infrastructureConfig *configv1.Infrastructure
 		rt                   *routev1.Route
 	}
@@ -56,7 +82,7 @@ func TestDefaultConfigMap(t *testing.T) {
 		want *corev1.ConfigMap
 	}{
 		{
-			name: "Test Default Config Map",
+			name: "Test generating default configmap without managed config override",
 			args: args{
 				operatorConfig: &operatorv1.Console{
 					TypeMeta:   metav1.TypeMeta{},
@@ -68,6 +94,7 @@ func TestDefaultConfigMap(t *testing.T) {
 					Spec:   configv1.ConsoleSpec{},
 					Status: configv1.ConsoleStatus{},
 				},
+				managedConfig: &corev1.ConfigMap{},
 				infrastructureConfig: &configv1.Infrastructure{
 					Status: configv1.InfrastructureStatus{
 						APIServerURL: mockAPIServer,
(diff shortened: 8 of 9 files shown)


benjaminapetersen:
Thinking now, just to make sure we don't need to worry about `docker build` as well.

Dockerfile
```diff
@@ -1,13 +1,12 @@
-FROM registry.svc.ci.openshift.org/openshift/release:golang-1.10 AS builder
-# create a work dir and copy local files in
+FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.10 AS builder
 WORKDIR /go/src/github.com/openshift/console-operator
 COPY . .
-RUN make build
+RUN ADDITIONAL_GOTAGS="ocp" make build WHAT="cmd/console"
 
-FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
+FROM registry.svc.ci.openshift.org/ocp/4.0:base
```
spadgett: Not sure if we should change the base image. Let's check on that.

spadgett:
I'd hold off on this change until we are also reading the configmap. Otherwise branding will be wrong for OKD.

spadgett:
We should be able remove Dockerfile.rhel7 after this, but probably need to coordinate with ART

benjaminapetersen:
@zherman0 are you doing the configmap as part of this PR, or separate?I imagine that part might need to come first.

spadgett:
/hold

Adding a hold to make sure we read the console-config config map in openshift-config-managed before going ahead with the Dockerfile changes.

zherman0:
I was breaking things into smaller parts for my PRs, but since the configmap and this change are tightly coupled, I will put everything into this PR. For now I will set this as WIP.

Dockerfile
benjaminapetersen: Also be mindful of #187 which is going to merge soon with some tweaks to the Dockerfile.

manifests/03-rbac-role-ns-openshift-config-managed.yaml
```diff
@@ -0,0 +1,21 @@
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: read-openshift-config-managed
```
benjaminapetersen: Probably `name: console-operator` to match the rest.

manifests/04-rbac-rolebinding.yaml
```diff
@@ -37,3 +37,30 @@ subjects:
 - kind: ServiceAccount
   name: console-operator
   namespace: openshift-console-operator
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: openshift-config-reader
```
benjaminapetersen: `name: console-operator` 
` oc get rolebinding -n openshift-console` should probably list `console-operator` as a rolebinding.

manifests/03-rbac-role-ns-openshift-config-managed.yaml
benjaminapetersen: ` oc get rolebinding -n openshift-console` should probably list `console-operator` as a rolebinding.

benjaminapetersen: example:
```
oc get rolebinding -n openshift-console
NAME                    AGE
console-operator        35h
system:deployers        35h
system:image-builders   35h
system:image-pullers    35h
```

pkg/console/operator/operator.go
```diff
@@ -3,6 +3,7 @@ package operator
 import (
 	// standard lib
 	"fmt"
+	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
```
benjaminapetersen: You'll want to move this import down with the other `openshift` imports.

pkg/console/operator/sync_v400.go
```diff
@@ -240,12 +240,11 @@ func SyncSecret(co *consoleOperator, recorder events.Recorder, operatorConfig *o
 // therefore no additional error handling is needed here.
 func SyncConfigMap(co *consoleOperator, recorder events.Recorder, operatorConfig *operatorv1.Console, consoleConfig *configv1.Console, infrastructureConfig *configv1.Infrastructure, rt *routev1.Route) (*corev1.ConfigMap, bool, error) {
 	logrus.Printf("validating console configmap...")
-
-	defaultConfigmap, _, err := configmapsub.DefaultConfigMap(operatorConfig, consoleConfig, infrastructureConfig, rt)
+	managedConfig, err := co.configMapClient.ConfigMaps(api.OpenShiftConsoleOperatorNamespace).Get(api.OpenshiftManagedConfigMap, metav1.GetOptions{})
```
benjaminapetersen: Wondering if we should do the `.Get()` where we get the other configs:
https://github.com/openshift/console-operator/blob/master/pkg/console/operator/operator.go#L151

Its a lot of configs to plumb through... but prob reveals that some tech debt to clean-up later would be to collect them and pass them 

```golang
// in pkg/console/operator/operator.go#L151
// since we have to pass them all so much?
type ConfigSet struct {
   consoleConfig
   operatorConfig
   infrastructureConfig
   ingressConfig
   brandingConfig
}
```

benjaminapetersen: Definitely need to handle the error here, however.  Add an extra 
```golang
    // we dont want to totally ignore the error here:
    managedConfig, err := co.configMapClient.ConfigMaps(api.OpenShiftConsoleOperatorNamespace).Get(api.OpenshiftManagedConfigMap, metav1.GetOptions{})
    // but we are ok with 404.  any others that matter?
    if !apierrors.IsNotFound(err) {
	return nil, false, err 
    }
```

pkg/console/starter/starter.go
```diff
@@ -2,6 +2,9 @@ package starter
 
 import (
 	"fmt"
+	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
```
benjaminapetersen: import ordering

pkg/console/starter/starter.go
```diff
@@ -108,13 +111,37 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		oauthinformers.WithTweakListOptions(tweakListOptionsForOAuth),
 	)
 
-	recorder := ctx.EventRecorder
+	resourceSyncerInformers := v1helpers.NewKubeInformersForNamespaces(
+		kubeClient,
+		api.TargetNamespace,
```
benjaminapetersen: Does the `syncer` need to watch all of these namespaces?  Or just `openshift-config-managed`?

pkg/console/starter/starter.go
```diff
@@ -108,13 +111,37 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
...
 
 	operatorClient := &operatorclient.OperatorClient{
 		Informers: operatorConfigInformers,
 		Client:    operatorConfigClient.OperatorV1(),
 	}
 
+	resourceSyncer := resourcesynccontroller.NewResourceSyncController(
+		operatorClient,
+		resourceSyncerInformers,
+		v1helpers.CachedSecretGetter(kubeClient.CoreV1(), resourceSyncerInformers),
```
benjaminapetersen: Did we need a secret as well?

manifests/03-rbac-role-ns-operator.yaml
```diff
@@ -9,6 +9,7 @@ rules:
   resources:
   - configmaps
   - events
+  - secrets
```
benjaminapetersen: Are we creating a new secret?

pkg/console/starter/starter.go
```diff
@@ -108,13 +111,37 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
...
+		v1helpers.CachedConfigMapGetter(kubeClient.CoreV1(), resourceSyncerInformers),
+		ctx.EventRecorder,
+	)
+
+	// add syncing for the managed ConfigMap
```
benjaminapetersen: NIT: I may change the comment from `what` to `why`:
`add syncing for the managed ConfigMap` 
vs 
`We need to sync the console-config configmap from openshift-config-managed because it may contain necessary branding configuration.`

pkg/console/starter/starter.go
```diff
@@ -145,11 +173,13 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		"console",
 		[]configv1.ObjectReference{
 			{Group: "operator.openshift.io", Resource: "consoles", Name: api.ConfigResourceName},
-			{Group: "config.openshift.io", Resource: "consoles", Name: api.ConfigResourceName},
-			{Group: "config.openshift.io", Resource: "infrastructures", Name: api.ConfigResourceName},
+			{Group: configv1.GroupName, Resource: "consoles", Name: api.ConfigResourceName},
+			{Group: configv1.GroupName, Resource: "infrastructures", Name: api.ConfigResourceName},
 			{Group: "oauth.openshift.io", Resource: "oauthclients", Name: api.OAuthClientName},
+			{Group: configv1.GroupName, Resource: "oauths", Name: api.ConfigResourceName},
```
benjaminapetersen: Do we need to list ` Resource: "oauths"`?

pkg/console/starter/starter.go
```diff
@@ -145,11 +173,13 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
...
 			{Resource: "namespaces", Name: api.OpenShiftConsoleOperatorNamespace},
 			{Resource: "namespaces", Name: api.OpenShiftConsoleNamespace},
+			{Resource: "namespaces", Name: api.OpenshiftConfigManagedNamespace},
```
benjaminapetersen: I believe since we are using the `Syncer` to copy the configmap into our namespace, we don't have to list the `openshift-config-managed` namespace here.  The implication is that we care about ALL the things in that namespace.  Since we copy over the configmap, we don't really care about anything else in the namespace.

pkg/console/starter/starter.go
```diff
@@ -159,13 +189,21 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 
 	configUpgradeableController := unsupportedconfigoverridescontroller.NewUnsupportedConfigOverridesController(operatorClient, ctx.EventRecorder)
 
-	kubeInformersNamespaced.Start(ctx.Done())
-	operatorConfigInformers.Start(ctx.Done())
-	configInformers.Start(ctx.Done())
-	routesInformersNamespaced.Start(ctx.Done())
-	oauthInformers.Start(ctx.Done())
+	for _, informer := range []interface {
```
benjaminapetersen: This is a clever refactor. Nice.

pkg/console/subresource/configmap/configmap.go
```diff
@@ -193,3 +193,13 @@ func authServerYaml(logoutRedirect string) yaml.MapSlice {
 func consoleBaseAddr(host string) string {
 	return util.HTTPS(host)
 }
+
+// Helper function that pulls the yaml struct out of the data section of a configmap yaml
```
benjaminapetersen: Does this solve the `|` issue?

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -68,6 +76,7 @@ func TestDefaultConfigMap(t *testing.T) {
 					Spec:   configv1.ConsoleSpec{},
 					Status: configv1.ConsoleStatus{},
 				},
+				newConfig: &corev1.ConfigMap{},
```
benjaminapetersen: Probably we now have a case for more than one test.  

```
tests := []struct {
    name string 
    args args
    want *corev1.ConfigMap
}{ 
   name: "Generate default configmap"
}, {
   name: "Generate configmap when extra branding config from openshift-config-managed is provided"
}, {
   name: "Ben should have written an extra test for operatorConfig.UnsupportedConfigOverrides"
}
```
I shall have to go write the test I forgot to write now :)

pkg/console/subresource/configmap/configmap.go
```diff
@@ -77,11 +76,12 @@ func DefaultConfigMap(operatorConfig *operatorv1.Console, consoleConfig *configv
 	configMap := Stub()
 	configMap.Data = map[string]string{}
 	unsupportedRaw := operatorConfig.Spec.UnsupportedConfigOverrides.Raw
+	newConfig := getDataFromConfigMapAsYaml(managedConfig)
```
benjaminapetersen: If the managed config `404`'d, what happens here?

pkg/console/subresource/configmap/configmap.go
benjaminapetersen: Curious if you `oc delete configmap console-config -n openshift-config-managed` does this error?

benjaminapetersen:
Related to:
- https://github.com/openshift/origin-branding/pull/4

pkg/console/starter/starter.go
benjaminapetersen: Ah, looks like this is not optional, the `secretsGetter` is a required parameter.  Even if we aren't syncing secrets.

manifests/03-rbac-role-ns-operator.yaml
benjaminapetersen: Ah, I see, `v1helpers.CachedSecretGetter(kubeClient.CoreV1(), resourceSyncerInformers),`
We have to give the syncer a secretGetter, even though we aren't copying a secret.

pkg/console/operator/sync_v400.go
zherman0: I originally put it at the very top (func sync) and passed it all the way through, but 1) it was not needed in other functions and 2) it was accessible through the consoleOperator.
My follow up question would be do the other items need to be passed?

pkg/console/subresource/configmap/configmap.go
zherman0: Yes, by looping through the map[]string it forces the "value" of the key "console-config.yaml" into a properly formatted string which removes the block ('|') character.

pkg/console/subresource/configmap/configmap.go
zherman0: I did test it; the theory would be an empty byte array is returned and then gets ignored by the mergeProcessConfig

pkg/console/subresource/configmap/configmap.go
zherman0: Let me retest this.  I did something a little different by removing the data in the console-config.yaml vs not having Data at all. I will get back to you.

pkg/console/subresource/configmap/configmap.go
zherman0: Tested this again with 1) regular values, 2) cm but with no data, 3) no cm in the openshift-config-managed ns.  All passed and I did not see errors in the log.

pkg/console/subresource/configmap/configmap.go
zherman0: Tested this again with 1) regular values, 2) cm but with no data, 3) no cm in the openshift-config-managed ns.  All passed and I did not see errors in the log.

pkg/console/subresource/configmap/configmap_test.go
zherman0: Great idea.  You should probably wait until this PR merges but I will add another unit test for managedConfigs.

benjaminapetersen: Yeah I'll wait on mine to avoid a conflict.

pkg/console/subresource/configmap/configmap.go
benjaminapetersen: Nice.

Dockerfile
```diff
@@ -1,8 +1,7 @@
 FROM registry.svc.ci.openshift.org/openshift/release:golang-1.10 AS builder
-# create a work dir and copy local files in
 WORKDIR /go/src/github.com/openshift/console-operator
 COPY . .
-RUN make build
+RUN ADDITIONAL_GOTAGS="ocp" make build WHAT="cmd/console"
```
benjaminapetersen: `ocp` by default, `okd` if this configmap exists (which the operator will merge):
https://github.com/openshift/origin-branding/pull/4/files#diff-26921f007ed806c079e9bb8160490990R12

just for review confirmation.

manifests/03-rbac-role-ns-openshift-config-managed.yaml
```diff
@@ -0,0 +1,21 @@
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: console-operator
```
benjaminapetersen: We are assuming the creation of the `namespace` `openshift-config-managed` before this file is applied.  I think its a safe assumption, but just pointing it out.  I'm not sure off the top of my head on the ordering.

benjaminapetersen: @spadgett do you know the ordering by any chance?

manifests/04-rbac-rolebinding.yaml
```diff
@@ -37,3 +37,30 @@ subjects:
...
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: console-operator
+  namespace: openshift-console
+roleRef:
+  kind: Role
+  name: console-operator
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+  - kind: User
```
benjaminapetersen: Up to this point, all of our `subjects` `kind` have been `kind: ServiceAccount`.  Why is this one `User`?

pkg/console/starter/starter.go
```diff
@@ -16,8 +16,11 @@ import (
 	"github.com/openshift/console-operator/pkg/api"
 	operatorclient "github.com/openshift/console-operator/pkg/console/operatorclient"
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
+	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
 	"github.com/openshift/library-go/pkg/operator/status"
 	"github.com/openshift/library-go/pkg/operator/unsupportedconfigoverridescontroller"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	"github.com/sirupsen/logrus"
```
benjaminapetersen: `logrus` should go up @ top above `kube` imports.

pkg/console/starter/starter.go
```diff
@@ -103,13 +106,35 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
...
+		v1helpers.CachedConfigMapGetter(kubeClient.CoreV1(), resourceSyncerInformers),
+		ctx.EventRecorder,
+	)
+
+	// We need to sync the console-config configmap from openshift-config-managed because it may contain necessary branding configuration.
+	if err := resourceSyncer.SyncConfigMap(
+		resourcesynccontroller.ResourceLocation{Namespace: api.OpenShiftConsoleOperatorNamespace, Name: api.OpenshiftManagedConfigMap},
+		resourcesynccontroller.ResourceLocation{Namespace: api.OpenshiftConfigManagedNamespace, Name: api.OpenShiftConsoleConfigMapName},
+	); err != nil {
+		logrus.Errorf("failed to copy managed configmap: %v \n", err)
```
benjaminapetersen: If it's a `404`, will this error always display?   Because it may not actually be an error, correct?  We are ok with a `not found`.

manifests/03-rbac-role-ns-openshift-config-managed.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - configmaps
+  - secrets
+  - pods
+  verbs:
+  - get
+  - list
+  - watch
+  - delete
+  - update
+  - create
```
spadgett: We're asking for a lot of unnecessary permissions here. We only should need permission to read config maps. (I definitely don't want the operator to be able to read secrets in this namespace if it's not needed.)

manifests/04-rbac-rolebinding.yaml
```diff
@@ -37,3 +37,30 @@ subjects:
...
+metadata:
+  name: console-operator
+  namespace: openshift-config-managed
+roleRef:
+  kind: Role
+  name: console-operator
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+  - kind: User
+    name: system:serviceaccount:openshift-console-operator:console-operator
```
spadgett:
```yaml
- kind: ServiceAccount
  name: console-operator
```

manifests/04-rbac-rolebinding.yaml
```diff
@@ -37,3 +37,30 @@ subjects:
...
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: console-operator
+  namespace: openshift-console
```
spadgett: Don't we already have this role binding above?

pkg/console/subresource/configmap/configmap.go
```diff
@@ -193,3 +193,13 @@ func authServerYaml(logoutRedirect string) yaml.MapSlice {
 func consoleBaseAddr(host string) string {
 	return util.HTTPS(host)
 }
+
+// Helper function that pulls the yaml struct out of the data section of a configmap yaml
+func getDataFromConfigMapAsYaml(managedConfig *corev1.ConfigMap) []byte {
```
benjaminapetersen: Could just call this `extractYAML` if you want to go shorter but no big deal.  Probs should caps `YAML` either way.

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -39,6 +39,13 @@ servingInfo:
   bindAddress: https://0.0.0.0:8443
   certFile: /var/serving-cert/tls.crt
   keyFile: /var/serving-cert/tls.key
+`
```
benjaminapetersen: `exampleConfigMapData` vs `exampleManagedConfigData`, just to clarify its `managed`.

pkg/console/operator/operator.go
```diff
@@ -68,7 +68,8 @@ type consoleOperator struct {
 	infrastructureClient configclientv1.InfrastructureInterface
 	versionGetter        status.VersionGetter
 	// recorder
-	recorder events.Recorder
+	recorder       events.Recorder
+	resourceSyncer resourcesynccontroller.ResourceSyncer
```
spadgett: It's unclear to me why we need to sync the config map. I'd think we would only need to do that if we were mounting it into a pod. In our case, we can read it in place in the operator.

manifests/03-rbac-role-ns-openshift-config-managed.yaml
```diff
@@ -0,0 +1,21 @@
...
+  namespace: openshift-config-managed
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - services
+  - events
+  - configmaps
+  - secrets
+  - pods
```
benjaminapetersen: Def should not need `pods`, guessing not `events`.  And we are just `watch`ing, so full CRUD is too much.

manifests/03-rbac-role-ns-openshift-config-managed.yaml
zherman0: I get this error if I do not give read access for secrets:
```Failed to list *v1.Secret: secrets is forbidden: User "system:serviceaccount:openshift-console-operator:console-operator" cannot list resource "secrets" in API group "" in the namespace "openshift-config-managed"
```

zherman0: I have remove the full CRUD for that ns

benjaminapetersen: You may also be able to use:

```
resourceNames:
- openshift-config
```
to only read that resource specifically, though the `lister` may complain (should test this).

pkg/console/starter/starter.go
zherman0: The SyncConfigMap function there is only checking for the namespaces at that point and is not checking if there is a specific file.  An error will only be thrown if the namespaces do not exist.  We check later in the code when we are looking for the specific file for a 404.

manifests/04-rbac-rolebinding.yaml
zherman0: I was using a different name prior to the last updates. Now this is a repeat and has been removed.

manifests/03-rbac-role-ns-openshift-config-managed.yaml
zherman0: done

pkg/console/starter/starter.go
```diff
@@ -103,13 +106,35 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
...
 
 	operatorClient := &operatorclient.OperatorClient{
 		Informers: operatorConfigInformers,
 		Client:    operatorConfigClient.OperatorV1(),
 	}
 
+	resourceSyncer := resourcesynccontroller.NewResourceSyncController(
+		operatorClient,
+		resourceSyncerInformers,
+		v1helpers.CachedSecretGetter(kubeClient.CoreV1(), resourceSyncerInformers),
```
benjaminapetersen: It would be nice if the `ConfigMapGetter` and `SecretGetter` were optional, it would eliminate our extra RBAC rules.

manifests/03-rbac-role-ns-operator.yaml
```diff
@@ -21,6 +21,7 @@ rules:
   - ""
   resources:
   - pods
+  - secrets
```
zherman0: Without this, get error:
```
Failed to list *v1.Secret: secrets is forbidden: User "system:serviceaccount:openshift-console-operator:console-operator" cannot list resource "secrets" in API group "" in the namespace "openshift-console-operator"
```

pkg/console/starter/starter.go
```diff
@@ -108,13 +112,36 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
...
+		api.OpenshiftConfigManagedNamespace,
+		api.OpenShiftConsoleOperatorNamespace,
+	)
 
 	operatorClient := &operatorclient.OperatorClient{
 		Informers: operatorConfigInformers,
 		Client:    operatorConfigClient.OperatorV1(),
 	}
 
+	resourceSyncer := resourcesynccontroller.NewResourceSyncController(
```
spadgett: I don't think we need the resource sync at all. We're not mounting the config map into a pod. We should be able to read it in place.

zherman0: Refactored code to not use the resourcesync.  Added new informer for the ConfigMaps in the `openshift-config-managed` namesapce and trigger off that to update the console config.

manifests/03-rbac-role-ns-openshift-config-managed.yaml
spadgett: > We are assuming the creation of the `namespace` `openshift-config-managed` before this file is applied. I think its a safe assumption, but just pointing it out. I'm not sure off the top of my head on the ordering.

Looking at the manifests in the release payload here, it shouldn't be an issue:

https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-ci-test/pr-logs/pull/openshift_console-operator/180/pull-ci-openshift-console-operator-master-e2e-aws/1229/artifacts/release-latest/release-payload-latest/

spadgett:
The test flakes we're seeing should be fixed by https://github.com/openshift/origin/pull/22543

pkg/console/starter/starter.go
zherman0: No longer using resource sync.

zherman0:
@spadgett  @benjaminapetersen - New unit tests added and commits squashed

manifests/03-rbac-role-ns-openshift-config-managed.yaml
```diff
@@ -0,0 +1,14 @@
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: console-operator
+  namespace: openshift-config-managed
+rules:
```
benjaminapetersen: Want to specify just the `console-config` specifically? I don't think we need `rbac` for all `configmaps` in the namespace.

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -39,6 +39,31 @@ servingInfo:
...
+	exampleManagedConfigMapData = `kind: ConsoleConfig
+apiVersion: console.openshift.io/v1beta1
+customization:
+  branding: online
+  documentationBaseURL: https://docs.okd.io/4.0/
+`
+	exampleYamlWithNewConfig = `kind: ConsoleConfig
```
benjaminapetersen: Any chance we have a clearer name than `NewConfig`?

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -112,19 +139,88 @@ func TestDefaultConfigMap(t *testing.T) {
 				BinaryData: nil,
 			},
 		},
+		{
+			name: "Test new Config Map that would come from managed namespace",
```
benjaminapetersen: - `Test generating default configmap without managed config override`
- `Test generating configmap with managed config to override branding`

Lets not use `New`.  Its new today, but won't be new forever.

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -112,19 +139,88 @@ func TestDefaultConfigMap(t *testing.T) {
...
+					Data:       map[string]string{configKey: exampleManagedConfigMapData},
+					BinaryData: nil,
+				},
+				infrastructureConfig: &configv1.Infrastructure{
+					Status: configv1.InfrastructureStatus{
+						APIServerURL: mockAPIServer,
+					},
+				},
+				rt: &routev1.Route{
+					TypeMeta:   metav1.TypeMeta{},
```
benjaminapetersen: Do we need all of the noise on the `route`?  Consider adding only the relevant fields.

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -112,19 +139,88 @@ func TestDefaultConfigMap(t *testing.T) {
...
+						AlternateBackends: nil,
+						Port:              nil,
+						TLS:               nil,
+						WildcardPolicy:    "",
+					},
+					Status: routev1.RouteStatus{},
+				},
+			},
+			want: &corev1.ConfigMap{
+				TypeMeta: metav1.TypeMeta{},
```
benjaminapetersen: Same here, is it necessary to add all of the boilerplate to the `configmap`, or can we trim it down to only the fields in use?

manifests/03-rbac-role-ns-openshift-config-managed.yaml
zherman0: I tried using resourceNames: "console-config" but that was not working, then I read:
```
Notably, if resourceNames are set, then the verb must not be list, watch, create, or deletecollection. Because resource names are not present in the URL for create, list, watch, and deletecollection API requests, those verbs would not be allowed by a rule with resourceNames set, since the resourceNames portion of the rule would not match the request
```
Any other suggestions?

zherman0: @benjaminapetersen @spadgett - Any suggestions? ^^^

benjaminapetersen: ok should be fine then.

benjaminapetersen:
/lgtm

We should get an e2e test for this as I expect we will not manually test it often.  Created https://jira.coreos.com/browse/CONSOLE-1385 to track.

PR #183, opened 2019-03-20
Select host from admitted ingress, else fall back to route.spec.host

benjaminapetersen:
Selects `host` from the `admitted` `ingress` on the route, if it exists. Else, fallback to `spec.host` as we have been doing in the past.

[CONSOLE-1336](https://jira.coreos.com/browse/CONSOLE-1336)

/assign @spadgett @jhadvig @zherman0

### DIFF
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -173,8 +173,12 @@ func sync_v400(co *consoleOperator, originalOperatorConfig *operatorv1.Console, 
 }
 
 func SyncConsoleConfig(co *consoleOperator, consoleConfig *configv1.Console, route *routev1.Route) (*configv1.Console, error) {
-	logrus.Printf("Updating console.config.openshift.io with hostname: %v \n", route.Spec.Host)
-	consoleConfig.Status.ConsoleURL = util.HTTPS(route.Spec.Host)
+	host, err := routesub.GetCanonicalHost(route)
+	if err != nil {
+		return nil, err
+	}
+	logrus.Printf("Updating console.config.openshift.io with hostname: %v \n", host)
+	consoleConfig.Status.ConsoleURL = util.HTTPS(host)
 	return co.consoleConfigClient.UpdateStatus(consoleConfig)
 }
 
@@ -203,6 +207,10 @@ func SyncDeployment(co *consoleOperator, recorder events.Recorder, operatorConfi
 // should not be called until route & secret dependencies are verified
 func SyncOAuthClient(co *consoleOperator, operatorConfig *operatorv1.Console, sec *corev1.Secret, rt *routev1.Route) (*oauthv1.OAuthClient, bool, error) {
 	logrus.Printf("validating oauthclient...")
+	host, err := routesub.GetCanonicalHost(rt)
+	if err != nil {
+		return nil, false, err
+	}
 	oauthClient, err := co.oauthClient.OAuthClients().Get(oauthsub.Stub().Name, metav1.GetOptions{})
 	if err != nil {
 		logrus.Errorf("%q: %v \n", "oauth", err)
@@ -210,7 +218,7 @@ func SyncOAuthClient(co *consoleOperator, operatorConfig *operatorv1.Console, se
 		return nil, false, errors.New("oauth client for console does not exist.")
 	}
 	// this should take care of all of our syncronization
-	oauthsub.RegisterConsoleToOAuthClient(oauthClient, rt, secretsub.GetSecretString(sec))
+	oauthsub.RegisterConsoleToOAuthClient(oauthClient, host, secretsub.GetSecretString(sec))
 	oauthClient, oauthChanged, oauthErr := oauthsub.ApplyOAuth(co.oauthClient, oauthClient)
 	if oauthErr != nil {
 		logrus.Errorf("%q: %v \n", "oauth", oauthErr)
--- a/pkg/console/starter/starter.go
+++ b/pkg/console/starter/starter.go
@@ -2,10 +2,14 @@ package starter
 
 import (
 	"fmt"
+
+	"github.com/openshift/api/oauth"
+
 	"os"
 	"time"
 
 	// kube
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/client-go/informers"
@@ -13,6 +17,7 @@ import (
 
 	// openshift
 	configv1 "github.com/openshift/api/config/v1"
+	operatorv1 "github.com/openshift/api/operator"
 	"github.com/openshift/console-operator/pkg/api"
 	operatorclient "github.com/openshift/console-operator/pkg/console/operatorclient"
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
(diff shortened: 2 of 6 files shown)


spadgett:
In what cases would these not match? We do check specifically if `spec.host` is the admitted ingress when determining if we're ready.

If `spec.host` is edited, what happens? (I don't think our operator can guard against that since it's a generated host, so we don't know what `spec.host` should be.)

benjaminapetersen:
/hold

There is more to this.  A `route` may have multiple `ingress`, and multiple `ingress` may be `admitted`.  If so, we need to pick the best one. We probably need to get the `config.openshift.io` `ingress` and use the `spec.domain`:

```bash 
oc get ingress.config.openshift.io -o yaml
apiVersion: v1
items:
- apiVersion: config.openshift.io/v1
  kind: Ingress
  metadata:
    creationTimestamp: 2019-03-20T18:26:23Z
    generation: 1
    name: cluster
    resourceVersion: "326"
    selfLink: /apis/config.openshift.io/v1/ingresses/cluster
    uid: aa855444-4b3d-11e9-a524-02302e09da8e
  spec:
    domain: apps.bpetersen-mar-20.devcluster.openshift.com
  status: {}
```

benjaminapetersen:
/hold

This needs additional work to get the `ingress` config from `config.openshift.io` and then use that to improve our matching.

pkg/console/subresource/route/route.go
```diff
@@ -162,13 +162,36 @@ func wildcard() routev1.WildcardPolicyType {
 	return routev1.WildcardPolicyNone
 }
 
+// Use the best host option for the route.
+// Initially this will be Route.Spec.Host, but once an ingress has
+// been admitted, it should be the host value from the ingress.
+func GetBestHost(route *routev1.Route) string {
```
jhadvig: I think we should have tests these functions

benjaminapetersen: agree

benjaminapetersen: I'll write tests once we are happy with impl.  I just revised all of these funcs per some discussions to ensure we got it right.  I think the new impl is better.

jhadvig:
@benjaminapetersen need to rebase cause of the https://github.com/openshift/console-operator/commit/c94c31b56d1b2fe2513794bc27c9dd8f238a6698

benjaminapetersen:
@jhadvig I expect #162 will also cause a conflict.  I'm ok with that, will make updates from comments & rebase after that merges.

pkg/console/starter/starter.go
```diff
@@ -163,11 +167,12 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
...
 		"console",
 		[]configv1.ObjectReference{
-			{Group: "operator.openshift.io", Resource: "consoles", Name: api.ConfigResourceName},
-			{Group: "config.openshift.io", Resource: "consoles", Name: api.ConfigResourceName},
-			{Group: "oauth.openshift.io", Resource: "oauthclients", Name: api.OAuthClientName},
-			{Resource: "namespaces", Name: api.OpenShiftConsoleOperatorNamespace},
-			{Resource: "namespaces", Name: api.OpenShiftConsoleNamespace},
+			{Group: operatorv1.GroupName, Resource: "consoles", Name: api.ConfigResourceName},
+			{Group: configv1.GroupName, Resource: "consoles", Name: api.ConfigResourceName},
+			{Group: configv1.GroupName, Resource: "ingress", Name: api.ConfigResourceName},
```
jhadvig: shouldn't thus be plural ?
```suggestion
			{Group: configv1.GroupName, Resource: "ingresses", Name: api.ConfigResourceName},
```

pkg/console/subresource/route/route.go
```diff
@@ -162,31 +166,42 @@ func wildcard() routev1.WildcardPolicyType {
 	return routev1.WildcardPolicyNone
 }
 
+// The canonical host is the ingress on the route with a host that matches the
+// ingressConfig.Spec.Domain.  When we have an ingress that matches, and when it
+// is admitted, the console will be fully accessible.
+func GetCanonicalHost(route *routev1.Route, ingressConfig *configv1.Ingress) (string, error) {
```
jhadvig: Let's add tests for this one :)

pkg/console/starter/starter.go
benjaminapetersen: Looks like you are correct:

https://github.com/openshift/cluster-config-operator/blob/master/manifests/0000_05_config-operator_01_ingress.crd.yaml#L10

I don't say `ingress` or `ingresses` often 😄

pkg/console/subresource/route/route.go
benjaminapetersen: Absolutely, if this seems the right approach, I'll def add tests for them.

benjaminapetersen:
Rebased & updated.
Tests will come.

jhadvig:
@benjaminapetersen we have a conflict in `03-rbac-role.yaml`, cause of the recaft you did last week.

pkg/console/subresource/route/route.go
```diff
@@ -162,31 +166,42 @@ func wildcard() routev1.WildcardPolicyType {
...
+	// only checking canonical host against admitted ingress
+	for _, ingress := range route.Status.Ingress {
+		if !isIngressAdmitted(ingress) {
+			continue
+		}
+		if ingress.RouterCanonicalHostname == ingressConfig.Spec.Domain {
```
benjaminapetersen: Considering as well:

```golang
route.status.ingress[].routerName==default
```

benjaminapetersen:
```golang
status:
  ingress:
  - conditions:
    - lastTransitionTime: 2019-04-01T15:07:00Z
      status: "True"
      type: Admitted
    host: console-openshift-console.apps.bpetersen-apr-1.devcluster.openshift.com
    routerCanonicalHostname: apps.bpetersen-apr-1.devcluster.openshift.com
    // this is probably the one we want, and means we may not need the ingressConfig at all.
    routerName: default
    wildcardPolicy: None```
```

jhadvig: So here Im not sure:
1. If the `routerName` will be always `default`,  cause I haven't found any docs or code that would suggest that `default` is the value that we should be checking.

2. Not sure if we shouldn't check both, 
```go
if ingress.RouterName == default  && ingress.RouterCanonicalHostname == ingressConfig.Spec.Domain {
...
```

benjaminapetersen: Right, is the name `default` intended to be API.

ironcladlou: @smarterclayton is it okay for other components to rely on the `default` ingresscontroller name? I'm not sure what alternative we could suggest (a selector?).

ironcladlou: Decision is to trust that the instance named `default` is the OOTB ingresscontroller. So the name `default` is indeed implicit stable API.

benjaminapetersen: Great we will work with this then.

jhadvig:
@benjaminapetersen let's rebase this one so we can merge it :)

benjaminapetersen:
/retest

`level=fatal msg="failed to initialize the cluster: Cluster operator console has not yet reported success: timed out waiting for the condition"`

benjaminapetersen:
Curious about that one.  May sync up with a bugzilla.

spadgett:
> `level=fatal msg="failed to initialize the cluster: Cluster operator console has not yet reported success: timed out waiting for the condition"`

We need to understand what happened there before merging the PR

/hold

spadgett:
> E0403 16:59:09.960740       1 reflector.go:134] github.com/openshift/client-go/config/informers/externalversions/factory.go:101: Failed to list *v1.Ingress: ingresses.config.openshift.io is forbidden: User "system:serviceaccount:openshift-console-operator:console-operator" cannot list resource "ingresses" in API group "config.openshift.io" at the cluster scope

jhadvig:
@benjaminapetersen had add the `ingresses` to the `03-rbac-role-cluster.yaml` and formated the starter.go... tests are passing now :)
Check https://github.com/openshift/console-operator/pull/194

Let me know what you think, so I can close this one :)

benjaminapetersen:
@jhadvig actually I was eliminating the `ingress` config, I just missed dropping it from the related objects, thus the RBAC fail.  

I just pushed the change & am testing locally.

pkg/console/subresource/route/route.go
```diff
@@ -162,31 +170,41 @@ func wildcard() routev1.WildcardPolicyType {
 	return routev1.WildcardPolicyNone
 }
 
+func GetCanonicalHost(route *routev1.Route) (string, error) {
```
ironcladlou: Log spam warning: if you're not caching this result and call this on every sync, consider logging throughout at a different verbosity level

benjaminapetersen: Thanks!  Fortunately, our sync loop idles out & is pretty quiet after the initial flurry, so we haven't worried about log levels yet.

benjaminapetersen:
`Evicted Pod The node had condition: [DiskPressure].`

Will `/retest` in a bit.

benjaminapetersen:
/retest

`level=fatal msg="failed to initialize the cluster: Cluster operator ingress has not yet reported success: timed out waiting for the condition"`

benjaminapetersen:
/retest

`fail [github.com/openshift/origin/test/extended/operators/operators.go:164]: Apr  5 15:02:08.684: Some cluster operators never became available <nil>/authentication, <nil>/openshift-apiserver`

benjaminapetersen:
/retest

`"Pod kube-system/etcd-member-ip-10-0-136-96.ec2.internal is not healthy: container certs has restarted more than 5 times",`

pkg/console/operator/sync_v400.go
```diff
@@ -173,8 +173,12 @@ func sync_v400(co *consoleOperator, originalOperatorConfig *operatorv1.Console,
 }
 
 func SyncConsoleConfig(co *consoleOperator, consoleConfig *configv1.Console, route *routev1.Route) (*configv1.Console, error) {
-	logrus.Printf("Updating console.config.openshift.io with hostname: %v \n", route.Spec.Host)
-	consoleConfig.Status.ConsoleURL = util.HTTPS(route.Spec.Host)
+	host, err := routesub.GetCanonicalHost(route)
+	if err != nil {
+		return nil, err
+	}
+	logrus.Printf("Updating console.config.openshift.io with hostname: %v \n", host)
```
jhadvig: r/Updating/updating

PR #187, opened 2019-03-27
Remove hardcoded amd64 from Dockerfile for MultiArch builds

jeffdyoung:
Remove hardcoded amd64 from Dockerfile for MultiArch builds.

### DIFF
--- a/Dockerfile.rhel7
+++ b/Dockerfile.rhel7
@@ -1,12 +1,14 @@
 FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.10 AS builder
 WORKDIR /go/src/github.com/openshift/console-operator
 COPY . .
-RUN ADDITIONAL_GOTAGS="ocp" make build WHAT="cmd/console"
+RUN ADDITIONAL_GOTAGS="ocp" make build WHAT="cmd/console"; \
+    mkdir -p /tmp/build; \
+    cp /go/src/github.com/openshift/console-operator/_output/local/bin/linux/$(go env GOARCH)/console /tmp/build/console
 
 FROM registry.svc.ci.openshift.org/ocp/4.0:base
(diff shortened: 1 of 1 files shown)


Dockerfile.rhel7
```diff
@@ -1,12 +1,14 @@
 FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.10 AS builder
 WORKDIR /go/src/github.com/openshift/console-operator
 COPY . .
-RUN ADDITIONAL_GOTAGS="ocp" make build WHAT="cmd/console"
+RUN ADDITIONAL_GOTAGS="ocp" make build WHAT="cmd/console"; \
```
benjaminapetersen: I'm fine with removing `amd64`, let's make sure we are consistent with other operators, however.

benjaminapetersen: (I imagine this step should be pretty standardized)

jeffdyoung: @benjaminapetersen  There seems to be 2 ways to approach this:

Build output to w/o arch in the output path:
https://github.com/openshift/cluster-autoscaler-operator/blob/master/Dockerfile.rhel7
https://github.com/openshift/cluster-samples-operator/blob/master/Dockerfile.rhel7

Build output w/arch in the output path:
https://github.com/openshift/origin/pull/22410
https://github.com/openshift/prom-label-proxy/pull/7
https://github.com/openshift/cluster-capacity/pull/20/files

^ for the ones that do use arch-specific paths, this is the fix we've been using.

Having gone through 30ish of these repos, there doesn't seems to be a right/wrong/preferred way.. just repo owner's preference.

benjaminapetersen: Great, works for me.

benjaminapetersen:
/retest

PR #175, opened 2019-03-14
Bump deps to pick up leader election tweaks & Azure branding option

benjaminapetersen:
- Replacing #174 
- Picks up update to leader election defaults (quiets them down)

PR #188, opened 2019-03-28
Split roles into 3 files for clarity & to avoid mistakes

benjaminapetersen:
From #175, all of the roles in 1 files makes it easy to make mistakes.  This just splits the file up cleanly so it is more obvious what role you are changing.  #175 will need a rebase soon, it includes 1 `rbac` change, but its hard to see due to the files being split apart.  I'd like to make this change first as a separate PR so that the real rbac change is an obvious single line change.

/assign @jhadvig


benjaminapetersen:
/retest

benjaminapetersen:
/retest 

PR #189, opened 2019-04-01
Simplify SyncSecret()

benjaminapetersen:
It is incorrect to `error` when we create our secret on a 404.  This is blocking, so it is fine to immediately go from `not found` to `create` and after that return errors.

/assign @jhadvig


benjaminapetersen:
added label as the robot missed the comment.

PR #195, opened 2019-04-05
Improve Status:Available & improve logging

benjaminapetersen:
- Removes log spam & adds specificity when logging happens to help ensure all logs are meaningful
- Ensures Status Available settings are accurate and reduces the possibility of status flutter.

### DIFF
--- a/pkg/console/errors/sync_errors.go
+++ b/pkg/console/errors/sync_errors.go
@@ -1,0 +1,31 @@
+package errors
+
+// a sync progressing error captures the idea that the operand is in an incomplete state
+// and needs to be reconciled.  It should be used when the sync loop should abort, but
+// it implies the idea of "progressing", not "failure".  A built-in error type can be
+// used and returned when a true failure is encountered.
+type SyncProgressingError struct {
+	message string
+}
+
+// implement the error interface
+func (e *SyncProgressingError) Error() string {
+	return e.message
+}
+
+// NewSyncError("Sync failed on xyz")
+// A builder func, should we choose to add additional metadata on the custom err
+func NewSyncError(msg string) *SyncProgressingError {
+	err := &SyncProgressingError{
+		message: msg,
+	}
+	return err
+}
+
+func IsSyncError(err error) bool {
+	if err == nil {
+		return false
+	}
+	_, ok := err.(*SyncProgressingError)
+	return ok
+}
--- a/pkg/console/operator/operator.go
+++ b/pkg/console/operator/operator.go
@@ -3,12 +3,12 @@ package operator
 import (
 	// standard lib
 	"fmt"
+	"reflect"
 	"time"
 
 	"github.com/sirupsen/logrus"
 
 	// kube
-
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -39,6 +39,7 @@ import (
 	operatorclientv1 "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1"
 
 	// operator
+	customerrors "github.com/openshift/console-operator/pkg/console/errors"
 	"github.com/openshift/console-operator/pkg/console/subresource/configmap"
 	"github.com/openshift/console-operator/pkg/console/subresource/deployment"
 	"github.com/openshift/console-operator/pkg/console/subresource/oauthclient"
@@ -48,8 +49,7 @@ import (
 )
 
 const (
-	controllerName           = "Console"
-	workloadFailingCondition = "WorkloadFailing"
+	controllerName = "Console"
 )
 
 var CreateDefaultConsoleFlag bool
@@ -176,38 +176,58 @@ func (c *consoleOperator) Sync(obj metav1.Object) error {
 		return err
 	}
 
-	// all configs needed to do a sync
 	if err := c.handleSync(operatorConfig, consoleConfig, infrastructureConfig); err != nil {
-		c.SyncStatus(c.ConditionDegraded(operatorConfig, "SyncLoopError", "Operator sync loop failed to complete."))
 		return err
 	}
-	c.SyncStatus(c.ConditionNotDegraded(operatorConfig))
+
 	return nil
 }
 
-func (c *consoleOperator) handleSync(originalOperatorConfig *operatorsv1.Console, consoleConfig *configv1.Console, infrastructureConfig *configv1.Infrastructure) error {
-	operatorConfig := originalOperatorConfig.DeepCopy()
+func (c *consoleOperator) handleSync(operatorConfig *operatorsv1.Console, consoleConfig *configv1.Console, infrastructureConfig *configv1.Infrastructure) error {
+	operatorConfigCopy := operatorConfig.DeepCopy()
 
-	switch operatorConfig.Spec.ManagementState {
+	switch operatorConfigCopy.Spec.ManagementState {
 	case operatorsv1.Managed:
 		logrus.Println("console is in a managed state.")
 		// handled below
 	case operatorsv1.Unmanaged:
 		logrus.Println("console is in an unmanaged state.")
-		c.SyncStatus(c.ConditionsManagementStateUnmanaged(operatorConfig))
+		c.ConditionsManagementStateUnmanaged(operatorConfigCopy)
+		if !reflect.DeepEqual(operatorConfigCopy, operatorConfig) {
+			c.SyncStatus(operatorConfigCopy)
+		}
 		return nil
 	case operatorsv1.Removed:
 		logrus.Println("console has been removed.")
-		c.SyncStatus(c.ConditionsManagementStateRemoved(operatorConfig))
-		return c.deleteAllResources(operatorConfig)
+		c.ConditionsManagementStateRemoved(operatorConfigCopy)
+		if !reflect.DeepEqual(operatorConfigCopy, operatorConfig) {
+			c.SyncStatus(operatorConfigCopy)
+		}
+		return c.deleteAllResources(operatorConfigCopy)
 	default:
-		// TODO should update status
-		return fmt.Errorf("unknown state: %v", operatorConfig.Spec.ManagementState)
+		c.ConditionsManagementStateInvalid(operatorConfigCopy)
+		if !reflect.DeepEqual(operatorConfigCopy, operatorConfig) {
+			c.SyncStatus(operatorConfigCopy)
+		}
+		return fmt.Errorf("console is in an unknown state: %v", operatorConfigCopy.Spec.ManagementState)
 	}
 
-	_, _, _, err := sync_v400(c, operatorConfig, consoleConfig, infrastructureConfig)
+	// we can default to not failing, and wait to see if sync returns an error
+	c.ConditionNotDegraded(operatorConfigCopy)
+	err := sync_v400(c, operatorConfigCopy, consoleConfig, infrastructureConfig)
 	if err != nil {
-		return err
+		if !customerrors.IsSyncError(err) {
+			c.SyncStatus(c.ConditionResourceSyncDegraded(operatorConfigCopy, err.Error()))
+			return err
+		} else {
+			c.SyncStatus(operatorConfigCopy)
+			return nil
+		}
+	}
+	// finally write out the set of conditions currently set if anything has changed
+	// to avoid a hot loop
+	if !reflect.DeepEqual(operatorConfigCopy, operatorConfig) {
+		c.SyncStatus(operatorConfigCopy)
 	}
 	return nil
 }
--- a/pkg/console/operator/status.go
+++ b/pkg/console/operator/status.go
@@ -50,17 +50,28 @@ import (
 //
 
 const (
+	reasonAsExpected          = "AsExpected"
+	reasonWorkloadFailing     = "WorkloadFailing"
 	reasonUnmanaged           = "ManagementStateUnmanaged"
 	reasonRemoved             = "ManagementStateRemoved"
+	reasonInvalid             = "ManagementStateInvalid"
 	reasonSyncLoopProgressing = "SyncLoopProgressing"
 	reasonSyncError           = "SynchronizationError"
 	reasonNoPodsAvailable     = "NoPodsAvailable"
-	reasonAsExpected          = "AsExpected"
+	reasonSyncLoopError       = "SyncLoopError"
 )
+
+// TODO:
+// a single builder helper would be sufficiently easy to read and
+// reason about. Consider migrating to this structure as part of aggregating status
+// AddStatus(ConditionFail)
+//			.To(ConditionTrue)
+//			.Reason("FooBar")
+//			.Message("This truly broke FooBar"))
 
 // Lets transition to using this, and get the repetition out of all of the above.
 func (c *consoleOperator) SyncStatus(operatorConfig *operatorsv1.Console) (*operatorsv1.Console, error) {
-	logConditions(operatorConfig.Status.Conditions)
+	c.logConditions(operatorConfig.Status.Conditions)
 	updatedConfig, err := c.operatorConfigClient.UpdateStatus(operatorConfig)
 	if err != nil {
 		errMsg := fmt.Errorf("status update error: %v \n", err)
@@ -75,8 +86,9 @@ func (c *consoleOperator) SyncStatus(operatorConfig *operatorsv1.Console) (*oper
 //   Status.Condition.<Condition>: <Bool> (<Reason>)
 //   Status.Condition.<Condition>: <Bool> (<Reason>) <Message>
 //   Status.Condition.<Condition>: <Bool> <Message>
-func logConditions(conditions []operatorsv1.OperatorCondition) {
+func (c *consoleOperator) logConditions(conditions []operatorsv1.OperatorCondition) {
 	logrus.Println("Operator.Status.Conditions")
+
 	for _, condition := range conditions {
 		buf := bytes.Buffer{}
 		buf.WriteString(fmt.Sprintf("Status.Condition.%s: %s", condition.Type, condition.Status))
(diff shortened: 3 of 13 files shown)


benjaminapetersen:
added logging for when resource versions change:

```
time="2019-04-05T20:14:47Z" level=info msg="validating console deployment..."
time="2019-04-05T20:14:47Z" level=info msg="deployment resource version [console.openshift.io/console-config-version] has changed from: 10471 to 768455 \n"
time="2019-04-05T20:14:47Z" level=info msg="deployment resource version [console.openshift.io/oauth-secret-version] has changed from: 10477 to 768461 \n"
I0405 20:14:47.850074       1 apps.go:59] Deployment "openshift-console/console" changes: {"metadata":{"annotations":{"console.openshift.io/console-config-version":"768455","console.openshift.io/oauth-secret-version":"768461"}},"spec":{"progressDeadlineSeconds":null,"revisionHistoryLimit":null,"strategy":{"rollingUpdate":null,"type":null},"template":{"metadata":{"annotations":{"console.openshift.io/console-config-version":"768455","console.openshift.io/oauth-secret-version":"768461"}},"spec":{"dnsPolicy":null,"volumes":[{"name":"console-serving-cert","secret":{"secretName":"console-serving-cert"}},{"name":"console-oauth-config","secret":{"secretName":"console-oauth-config"}},{"configMap":{"name":"console-config"},"name":"console-config"},{"configMap":{"name":"service-ca"},"name":"service-ca"}]}}}}
I0405 20:14:47.855287       1 option.go:62] Console: handling update openshift-console/console: /apis/apps/v1/namespaces/openshift-console/deployments/console
time="2019-04-05T20:14:47Z" level=info msg="deployment exists and is in the correct state"
```
This helps identify what causes pods to roll.

benjaminapetersen:
Now, if I:

```
oc delete secret console-oauth-config -n openshift-console
# secret "console-oauth-config" deleted
```
I can see:
```
oc get -w deployment console -n openshift-console
NAME      READY   UP-TO-DATE   AVAILABLE   AGE
console   2/2     2            2           5m10s
console   2/2   2     2     5m25s
console   2/2   2     2     5m25s
console   2/2   0     2     5m25s
console   2/2   1     2     5m25s
console   3/2   1     3     5m41s
console   2/2   1     2     5m41s
console   2/2   2     2     5m41s
```
There is some jitter there, but always `2` `READY` pods.  `UP-TO-DATE` can drop to `0`, then climbs back to `2`.  This validates we are correctly doing a rolling deployment, there should be no downtime.

benjaminapetersen:
/retest

benjaminapetersen:
/retest

`e2e-aws` does not flap on `available`.  Need to see `e2e-aws-operator`.

pkg/console/operator/operator.go
```diff
@@ -175,10 +174,8 @@ func (c *consoleOperator) Sync(obj metav1.Object) error {
 
 	// all configs needed to do a sync
 	if err := c.handleSync(operatorConfig, consoleConfig, infrastructureConfig); err != nil {
```
benjaminapetersen: Status is handled fine within `handleSync()` & below. There is no need to handle here.

pkg/console/operator/status.go
```diff
@@ -50,17 +50,35 @@ import (
 //
 
 const (
-	reasonUnmanaged           = "ManagementStateUnmanaged"
-	reasonRemoved             = "ManagementStateRemoved"
+	// standard reasons used by other operators for consistency
```
benjaminapetersen: This is just organizational.

pkg/console/operator/status.go
```diff
@@ -50,17 +50,35 @@ import (
...
+	reasonRemoved   = "ManagementStateRemoved"
+	// these are most meaningful and specific to our operator
 	reasonSyncLoopProgressing = "SyncLoopProgressing"
 	reasonNoPodsAvailable     = "NoPodsAvailable"
-	reasonSyncError           = "SyncError"
-	reasonAsExpected          = "AsExpected"
+	reasonSyncIncomplete      = "ResourceSyncIncomplete"
 )
 
+// TODO: we should have the concept of an error budget, track various types of failures,
```
benjaminapetersen: Created a JIRA ticket in the backlog to follow-up on the idea of an error/failure budget if we ever think its actionable.

pkg/console/operator/status.go
```diff
@@ -136,70 +141,27 @@ func (c *consoleOperator) ConditionNotFailing(operatorConfig *operatorsv1.Consol
...
-
 // When a sync failure happens,
-// we dont know if the operand is available
+// we dont know if the operand is available, so it is best not to set it.
+//  on install, an incomplete sync will mean the operator is unavailable.
+//  however, on a later sync when a change is encountered, this is not true.
 // we do know we are progressing because we are trying to change something about the operand
 // we do know we failed to make the update
-func (c *consoleOperator) ConditionResourceSyncFailure(operatorConfig *operatorsv1.Console, message string) *operatorsv1.Console {
-	// message := "The operator failed to update a resource of the operand."
```
benjaminapetersen: `SyncFailure` is not exactly accurate, this is used to imply we deliberately aborted the sync loop with an early return, with the intent that the next sync loop will pick up and progress our state further.  Therefore, the name change to simply `SyncIncomplete` and removal of a statement on `Availability`.

pkg/console/operator/status.go
```diff
@@ -226,12 +188,12 @@ func (c *consoleOperator) ConditionDeploymentAvailable(operatorConfig *operators
 	return operatorConfig
 }
 
-func (c *consoleOperator) ConditionDeploymentNotAvailable(operatorConfig *operatorsv1.Console) *operatorsv1.Console {
+func (c *consoleOperator) ConditionDeploymentNotAvailable(operatorConfig *operatorsv1.Console, message string) *operatorsv1.Console {
```
benjaminapetersen: Changed this to take a message string so we can pass in `pods: 0,1,2` which is more informative.

pkg/console/operator/sync_v400.go
```diff
@@ -53,49 +53,63 @@ func sync_v400(co *consoleOperator, originalOperatorConfig *operatorv1.Console,
...
+		fmt.Printf("incomplete sync: %v \n", msg)
+		co.SyncStatus(co.ConditionResourceSyncIncomplete(operatorConfig, msg))
 		return operatorConfig, consoleConfig, toUpdate, svcErr
 	}
 	toUpdate = toUpdate || svcChanged
 
 	cm, cmChanged, cmErr := SyncConfigMap(co, recorder, operatorConfig, consoleConfig, infrastructureConfig, rt)
 	if cmErr != nil {
-		co.SyncStatus(co.ConditionResourceSyncFailure(operatorConfig, fmt.Sprintf("%q: %v\n", "configmap", cmErr)))
+		msg := fmt.Sprintf("%q: %v\n", "configmap", cmErr)
```
benjaminapetersen: These are a bit cumbersome, but `logging` + `status` seems the best way to ensure state is clearly known to someone attempting to debug.

pkg/console/operator/sync_v400.go
```diff
@@ -160,13 +180,28 @@ func sync_v400(co *consoleOperator, originalOperatorConfig *operatorv1.Console,
...
 
 	defer func() {
-		logrus.Printf("sync loop 4.0.0 complete:")
-		logrus.Printf("\t service changed: %v", svcChanged)
-		logrus.Printf("\t route changed: %v", rtChanged)
-		logrus.Printf("\t configMap changed: %v", cmChanged)
-		logrus.Printf("\t secret changed: %v", secChanged)
-		logrus.Printf("\t oauth changed: %v", oauthChanged)
-		logrus.Printf("\t deployment changed: %v", depChanged)
+		logrus.Printf("sync loop 4.0.0 complete")
```
benjaminapetersen: Reduce log spam.  There is nothing useful in just stating `true/false` a resource changed.  But if it changed, logging the version may help us know that something fluttered for some reason (and if the rest of the logs are unclear, we will know we need to add additional logging).

pkg/console/subresource/deployment/deployment.go
```diff
@@ -154,6 +171,27 @@ func Stub() *appsv1.Deployment {
 	return dep
 }
 
+func LogDeploymentAnnotationChanges(client appsclientv1.DeploymentsGetter, updated *appsv1.Deployment) {
```
benjaminapetersen: This is intended to ensure we can trace what object may have caused the deployment to roll pods.

pkg/console/operator/sync_v400.go
```diff
@@ -177,36 +212,44 @@ func SyncConsoleConfig(co *consoleOperator, consoleConfig *configv1.Console, rou
...
 		return nil, false, applyDepErr
 	}
-	logrus.Println("deployment exists and is in the correct state")
 	return deployment, deploymentChanged, nil
 }
 
 // applies changes to the oauthclient
 // should not be called until route & secret dependencies are verified
 func SyncOAuthClient(co *consoleOperator, operatorConfig *operatorv1.Console, sec *corev1.Secret, rt *routev1.Route) (*oauthv1.OAuthClient, bool, error) {
-	logrus.Printf("validating oauthclient...")
```
benjaminapetersen: For the most part, these lines seem to be spam.  We should really only need a line to signal kicking the sync, not a ton of repetitive noise.

pkg/console/operator/sync_v400.go
```diff
@@ -177,36 +212,44 @@ func SyncConsoleConfig(co *consoleOperator, consoleConfig *configv1.Console, rou
 	if err != nil {
 		return nil, err
 	}
-	logrus.Printf("Updating console.config.openshift.io with hostname: %v \n", host)
-	consoleConfig.Status.ConsoleURL = util.HTTPS(host)
+	httpsHost := util.HTTPS(host)
+	if consoleConfig.Status.ConsoleURL != httpsHost {
```
benjaminapetersen: Reduce spam, only log when change happens.

benjaminapetersen:
/assign @spadgett @jhadvig @zherman0 

Feedback welcome!

pkg/console/operator/status.go
```diff
@@ -50,17 +50,21 @@ import (
...
+	reasonWorkloadFailing = "WorkloadFailing"
+	// the simplest resolutions would likely come from management state
+	reasonUnmanaged = "ManagementStateUnmanaged"
+	reasonRemoved   = "ManagementStateRemoved"
+	// these are most meaningful and specific to our operator
 	reasonSyncLoopProgressing = "SyncLoopProgressing"
 	reasonNoPodsAvailable     = "NoPodsAvailable"
-	reasonSyncError           = "SyncError"
-	reasonAsExpected          = "AsExpected"
+	reasonSyncIncomplete      = "ResourceSyncIncomplete"
```
spadgett: Should this be something like `SyncInProgress` or `Reconciling`? Incomplete sounds like there's a problem.

pkg/console/operator/sync_v400.go
```diff
@@ -53,49 +53,63 @@ func sync_v400(co *consoleOperator, originalOperatorConfig *operatorv1.Console,
 
 	rt, rtChanged, rtErr := SyncRoute(co, operatorConfig)
 	if rtErr != nil {
-		co.SyncStatus(co.ConditionResourceSyncFailure(operatorConfig, fmt.Sprintf("%v: %s\n", "route", rtErr)))
+		msg := fmt.Sprintf("%v: %s\n", "route", rtErr)
+		fmt.Printf("incomplete sync: %v \n", msg)
```
spadgett: logrus? (here and below)

spadgett: I'm confused: is this an error or isn't it? If it's an error from the API, it seems like we should be failing. (Ideally, we'd have an error budget, though.)

pkg/console/operator/sync_v400.go
spadgett: +1 to reducing log spam

pkg/console/subresource/deployment/deployment.go
```diff
@@ -154,6 +171,27 @@ func Stub() *appsv1.Deployment {
...
+	existing, err := client.Deployments(updated.Namespace).Get(updated.Name, metav1.GetOptions{})
+	if err != nil {
+		return
```
spadgett: Shouldn't we log the error?

benjaminapetersen: I can. The caller has a good bit of logging around it (`SyncDeployment`) but no point in coupling.

pkg/console/operator/sync_v400.go
benjaminapetersen: Without an error budget, its wrong to fail.  That has been explicitly stated.  But the idea of an error budget has not been fleshed out well at this point.

This may be an error (404, etc), but we should be able to tolerate it.  Its possible that the last stanza (example):

```golang
	// any error should be returned & kill the sync loop
        // if we hit this one, we may be truly failing
	if err != nil {
		logrus.Errorf("%q: %v \n", "secret", err)
		return nil, false, err
	}
	return secret, false, nil
```
Would be the correct place to truly `Failing:True`, if we had a concrete understanding of an acceptible error budget.

pkg/console/operator/status.go
benjaminapetersen: Sounds good, I will update.

benjaminapetersen: I will eliminate this entirely, the `reasonSyncLoopProgressing = "SyncLoopProgressing"` and `func ConditionResourceSyncProgressing()` is sufficient.

pkg/console/operator/sync_v400.go
spadgett: Does this mean we never transition to Failing: True? What if things are completely broken and console never reconciles?

pkg/console/operator/sync_v400.go
```diff
@@ -132,14 +146,17 @@ func sync_v400(co *consoleOperator, originalOperatorConfig *operatorv1.Console,
 	// - route is admitted
 	// available is currently defined as "met the users intent"
```
benjaminapetersen: - [ ] If we get to this point, `Failing:False` is the correct status.

benjaminapetersen:
/hold 

Rebased.
Update for `Failed` coming

pkg/console/operator/operator.go
```diff
@@ -175,10 +174,8 @@ func (c *consoleOperator) Sync(obj metav1.Object) error {
...
-		c.SyncStatus(c.ConditionFailing(operatorConfig, "SyncLoopError", "Operator sync loop failed to completele."))
```
spadgett: It looks like we never set `Failing: True` after this PR?

spadgett: Or is that what "Update for Failed coming" means? :)

benjaminapetersen: Yeah the `/hold` means more is coming :)

spadgett: sorry i thought it was about the failed -> degraded changes. carry on :)

pkg/console/operator/status.go
benjaminapetersen: Unnecessary, the `sync` client will already wait a minute before reporting a real failure to account for an error budget.

pkg/console/subresource/deployment/deployment.go
```diff
@@ -79,6 +95,7 @@ func DefaultDeployment(operatorConfig *operatorv1.Console, cm *corev1.ConfigMap,
 	meta.Annotations[configMapResourceVersionAnnotation] = cm.GetResourceVersion()
 	meta.Annotations[serviceCAConfigMapResourceVersionAnnotation] = serviceCAConfigMap.GetResourceVersion()
 	meta.Annotations[secretResourceVersionAnnotation] = sec.GetResourceVersion()
+	meta.Annotations[routeResourceVersionAnnotation] = rt.GetResourceVersion()
```
spadgett: Is this needed? The public URL is in the config map. Is there another time we need to force a rollout when the route changes?

benjaminapetersen: If our availability flickers, I'd like to be able to see clearly what resource changed, via its resource version, on the deployment.  This is just to give us a clue should routing ever flutter.

spadgett: But it will force unnecessary rollouts whenever the route status changes. This could force several unnecessary rollouts on initial install. It would be better to simply log the route status.

pkg/console/subresource/oauthclient/oauthclient.go
```diff
@@ -18,7 +18,7 @@ import (
 // - should look like resourceapply.ApplyService and the other Apply funcs
 //   once its in a trustworthy state, PR to library-go so it can live with
 //   the other Apply funcs
-func ApplyOAuth(client oauthclient.OAuthClientsGetter, required *oauthv1.OAuthClient) (*oauthv1.OAuthClient, bool, error) {
+func CustomApplyOAuth(client oauthclient.OAuthClientsGetter, required *oauthv1.OAuthClient) (*oauthv1.OAuthClient, bool, error) {
```
benjaminapetersen: Just renaming this to make it clear, this is not a `library-go` apply func, its a custom impl emulating a `library-go` apply func.

pkg/console/errors/sync_errors.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+// NewSyncError("Sync failed on xyz")
+func NewSyncError(msg string) *SyncProgressingError {
+	err := &SyncProgressingError{
+		message: msg,
+	}
+	return err
+}
+
```
benjaminapetersen: This helper + the custom err is what will clarify our psuedo-failing condition which signals `progressing` rather than `degrated`.

benjaminapetersen:
/retest

benjaminapetersen:
squashed.  couple rebase discrepancies to work out, will push again.

pkg/console/operator/operator.go
```diff
@@ -176,38 +176,53 @@ func (c *consoleOperator) Sync(obj metav1.Object) error {
...
 
-	switch operatorConfig.Spec.ManagementState {
+	switch operatorConfigCopy.Spec.ManagementState {
 	case operatorsv1.Managed:
 		logrus.Println("console is in a managed state.")
 		// handled below
 	case operatorsv1.Unmanaged:
 		logrus.Println("console is in an unmanaged state.")
-		c.SyncStatus(c.ConditionsManagementStateUnmanaged(operatorConfig))
+		c.SyncStatus(c.ConditionsManagementStateUnmanaged(operatorConfigCopy))
```
spadgett: Won't this hot-loop for unmanaged/removed state? I think we need an equals check like we do below for managed state.

pkg/console/operator/operator.go
```diff
@@ -176,38 +176,53 @@ func (c *consoleOperator) Sync(obj metav1.Object) error {
...
-		// TODO should update status
-		return fmt.Errorf("unknown state: %v", operatorConfig.Spec.ManagementState)
+		c.SyncStatus(c.ConditionsManagementStateInvalid(operatorConfigCopy))
+		return fmt.Errorf("console is in an unknown state: %v", operatorConfigCopy.Spec.ManagementState)
 	}
 
-	_, _, _, err := sync_v400(c, operatorConfig, consoleConfig, infrastructureConfig)
+	// we can default to not failing, and wait to see if sync returns an error
+	c.ConditionNotFailing(operatorConfigCopy)
+	_, _, _, err := sync_v400(c, operatorConfigCopy, consoleConfig, infrastructureConfig)
```
spadgett: Not new, but do we need the extra return values in `sync_v400` if we just ignore them?

pkg/console/operator/operator.go
```diff
@@ -176,38 +176,53 @@ func (c *consoleOperator) Sync(obj metav1.Object) error {
...
 	if err != nil {
-		return err
+		// if we didn't get a custom sync error, then we are failing this does
+		// not yet handle the idea of an error budget.
```
spadgett: Is this comment still accurate since library-go will only move to degraded status after a certain amount of time?

pkg/console/operator/sync_v400.go
```diff
@@ -130,25 +143,22 @@ func sync_v400(co *consoleOperator, originalOperatorConfig *operatorv1.Console,
...
 	if !deploymentsub.IsReady(actualDeployment) {
-		co.ConditionDeploymentNotAvailable(operatorConfig)
+		msg := fmt.Sprintf("%v pods available for console deployment", actualDeployment.Status.ReadyReplicas)
+		logrus.Println(msg)
+		co.ConditionDeploymentNotAvailable(operatorConfig, msg)
 	} else if !routesub.IsAdmitted(rt) {
+		fmt.Println("console route is not admitted")
```
spadgett: logrus or remove print statement

pkg/console/operator/operator.go
benjaminapetersen: We don't, I'll remove them.  These values assumed we would do operator status here (before we knew what that would look like).

pkg/console/operator/operator.go
benjaminapetersen: agree, removed.

pkg/console/subresource/deployment/deployment.go
benjaminapetersen: Good point. 😅 
I'll eliminate this and just log the changes.

benjaminapetersen:
`could not wait for build: the build console-operator failed after 1m17s with reason DockerBuildFailed: Docker build strategy has failed.`

pkg/console/operator/operator.go
benjaminapetersen: It doesn't hot loop (has been tested).  
I believe because `UpdateStatus` does a `client.Put()`, so the resource isn't changed if the value is the same thus no loop.

spadgett: I just tried, and it seems to hot loop:

screencast 2019-04-23 09-59-36

spadgett:
```
=== RUN   TestUnmanaged
--- FAIL: TestUnmanaged (6.46s)
	console.go:94: changing console operator state to 'Unmanaged'...
	framework.go:33: deleting console Deployment...
	framework.go:33: deleting console Service...
	framework.go:33: deleting console Route...
	framework.go:33: deleting console ConfigMap...
	unmanaged_test.go:20: waiting to check if the operator has not recreate deleted resources...
	unmanaged_test.go:29: error: deleted console Deployment was recreated
	console.go:70: changing console operator state to 'Managed'...
```

benjaminapetersen:
Kinda hate to keep bumping the poll, `30 seconds` is a long time, but it is what it is.

benjaminapetersen:
`e2e-aws` `Available` is `False,False,False,True...........` the way we want it.

pkg/console/operator/operator.go
benjaminapetersen: It just did for me as well.
Puzzled on why it didn't before. I'll add the guard. Best be safe.

spadgett: > It just did for me as well.
> Puzzled on why it didn't before.

I don't think it's new :/

PR #202, opened 2019-04-16
Change default api value to v1

zherman0:
Change the auto generated default api version to use the correct v1 instead of v1beta1


benjaminapetersen:
/lgtm

Thanks!

spadgett:
@zherman0 unit test failure

PR #203, opened 2019-04-17
Update SyncRoute() to look for host on ingress, not route.spec.host

benjaminapetersen:
Everywhere else we use `GetCanonicalHost()`, no longer `Route.Spec.Host`.  However, I am looking to see if there is a significant timing difference between these values being set.  That said, I don't believe `Route.Spec.Host` is reliable.

### DIFF
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -171,9 +171,9 @@ func sync_v400(co *consoleOperator, originalOperatorConfig *operatorv1.Console, 
 }
 
 func SyncConsoleConfig(co *consoleOperator, consoleConfig *configv1.Console, route *routev1.Route) (*configv1.Console, error) {
-	host, err := routesub.GetCanonicalHost(route)
-	if err != nil {
-		return nil, err
+	host := routesub.GetCanonicalHost(route)
+	if host == "" {
+		return nil, errors.New("waiting on host")
 	}
 	logrus.Printf("updating console.config.openshift.io with hostname: %v \n", host)
 	consoleConfig.Status.ConsoleURL = util.HTTPS(host)
@@ -205,9 +205,9 @@ func SyncDeployment(co *consoleOperator, recorder events.Recorder, operatorConfi
 // should not be called until route & secret dependencies are verified
 func SyncOAuthClient(co *consoleOperator, operatorConfig *operatorv1.Console, sec *corev1.Secret, rt *routev1.Route) (*oauthv1.OAuthClient, bool, error) {
 	logrus.Printf("validating oauthclient...")
-	host, err := routesub.GetCanonicalHost(rt)
-	if err != nil {
-		return nil, false, err
+	host := routesub.GetCanonicalHost(rt)
+	if host == "" {
+		return nil, false, errors.New("waiting on host")
 	}
 	oauthClient, err := co.oauthClient.OAuthClients().Get(oauthsub.Stub().Name, metav1.GetOptions{})
 	if err != nil {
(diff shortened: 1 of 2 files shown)


benjaminapetersen:
/assign @spadgett @jhadvig 

I believe this is correct.  Even if it is a bit slower, we won't report `Available` until `GetCanonicalHost()` returns a host.  I don't think we should trust two sources on the `route`.

pkg/console/operator/sync_v400.go
```diff
@@ -332,10 +332,10 @@ func SyncRoute(co *consoleOperator, operatorConfig *operatorv1.Console) (*routev
 
 	// we will not proceed until the route is valid. this eliminates complexity with the
 	// configmap, secret & oauth client as they can be certain they have a host if we pass this point.
-	if len(rt.Spec.Host) == 0 {
-		// TODO STATUS
+	_, err := routesub.GetCanonicalHost(rt)
+	if err != nil {
```
spadgett: Not new, but I think `GetCanonicalHost` should just return `nil` when there's no host set yet other than an error. I would try to distinguish between errors from the API and something just not being set yet.

benjaminapetersen: agree, i'll update.

pkg/console/subresource/route/route.go
```diff
@@ -123,7 +121,7 @@ func wildcard() routev1.WildcardPolicyType {
 	return routev1.WildcardPolicyNone
 }
 
-func GetCanonicalHost(route *routev1.Route) (string, error) {
+func GetCanonicalHost(route *routev1.Route) string {
```
benjaminapetersen: @spadgett updated here to just return a string.

pkg/console/operator/sync_v400.go
```diff
@@ -332,10 +332,10 @@ func SyncRoute(co *consoleOperator, operatorConfig *operatorv1.Console) (*routev
 
 	// we will not proceed until the route is valid. this eliminates complexity with the
 	// configmap, secret & oauth client as they can be certain they have a host if we pass this point.
-	if len(rt.Spec.Host) == 0 {
-		// TODO STATUS
+	host := routesub.GetCanonicalHost(rt)
```
benjaminapetersen: This is still the first attempt in the sync loop to get the `host`, so the above 2 instances of `GetCanonicalHost()` will not be called if this one does not pass.

pkg/console/operator/sync_v400.go
```diff
@@ -332,10 +332,10 @@ func SyncRoute(co *consoleOperator, operatorConfig *operatorv1.Console) (*routev
...
+	if host == "" {
 		logrus.Errorf("%q: %v \n", "route", rtErr)
-		return nil, false, errors.New("waiting on route.spec.host")
+		return nil, false, errors.New("waiting on host")
```
spadgett: I'd make this message more specific like, "waiting on admitted host for default router"

pkg/console/operator/sync_v400.go
```diff
@@ -171,9 +171,9 @@ func sync_v400(co *consoleOperator, originalOperatorConfig *operatorv1.Console,
 }
 
 func SyncConsoleConfig(co *consoleOperator, consoleConfig *configv1.Console, route *routev1.Route) (*configv1.Console, error) {
-	host, err := routesub.GetCanonicalHost(route)
-	if err != nil {
-		return nil, err
+	host := routesub.GetCanonicalHost(route)
+	if host == "" {
+		return nil, errors.New("waiting on host")
```
benjaminapetersen: These errors will be transformed into custom `SyncError` in my next PR that will be used to report `Progressing` vs `Failing`.

PR #206, opened 2019-04-18
e2e tests for LogLevel

jhadvig:
/assign @benjaminapetersen 

Check the tests comments

Story: https://jira.coreos.com/browse/CONSOLE-1354

### DIFF
--- a/pkg/console/subresource/deployment/deployment.go
+++ b/pkg/console/subresource/deployment/deployment.go
@@ -229,20 +229,24 @@ func consoleVolumeMounts(vc []volumeConfig) []corev1.VolumeMount {
 	return volMountList
 }
 
-type LogMapping map[operatorv1.LogLevel]string
-
-var LogMap = LogMapping{
-	operatorv1.Normal:   "--log-level=*=NOTICE",
-	operatorv1.Debug:    "--log-level=*=DEBUG",
-	operatorv1.Trace:    "--log-level=*=TRACE",
-	operatorv1.TraceAll: "--log-level=*=TRACE",
+func GetLogLevelFlag(logLevel operatorv1.LogLevel) string {
+	flag := ""
+	switch logLevel {
+	case operatorv1.Normal:
+		flag = "--log-level=*=NOTICE"
+	case operatorv1.Debug:
+		flag = "--log-level=*=DEBUG"
+	case operatorv1.Trace, operatorv1.TraceAll:
+		flag = "--log-level=*=TRACE"
+	}
+	return flag
 }
 
 func consoleContainer(cr *operatorv1.Console) corev1.Container {
 	volumeMounts := consoleVolumeMounts(volumeConfigList)
 	// Since the console-operator logging has different logging levels then the capnslog,
 	// that we use for console server(bridge) we need to map them to each other
-	flag := LogMap[cr.Spec.LogLevel]
+	flag := GetLogLevelFlag(cr.Spec.LogLevel)
 
 	return corev1.Container{
 		Image:           util.GetImageEnv(),
--- a/pkg/console/subresource/deployment/deployment.go
+++ b/pkg/console/subresource/deployment/deployment.go
@@ -229,24 +229,20 @@ func consoleVolumeMounts(vc []volumeConfig) []corev1.VolumeMount {
 	return volMountList
 }
 
+type LogMapping map[operatorv1.LogLevel]string
+
+var LogMap = LogMapping{
+	operatorv1.Normal:   "--log-level=*=NOTICE",
+	operatorv1.Debug:    "--log-level=*=DEBUG",
+	operatorv1.Trace:    "--log-level=*=TRACE",
+	operatorv1.TraceAll: "--log-level=*=TRACE",
+}
+
 func consoleContainer(cr *operatorv1.Console) corev1.Container {
 	volumeMounts := consoleVolumeMounts(volumeConfigList)
-	level := ""
 	// Since the console-operator logging has different logging levels then the capnslog,
 	// that we use for console server(bridge) we need to map them to each other
-	switch cr.Spec.LogLevel {
-	case operatorv1.Normal:
-		level = "NOTICE"
-	case operatorv1.Debug:
-		level = "DEBUG"
-	case operatorv1.Trace:
-		fallthrough
-	case operatorv1.TraceAll:
-		fallthrough
-	default:
-		level = "TRACE"
-	}
-	flag := fmt.Sprintf("--log-level=*=%s", level)
+	flag := LogMap[cr.Spec.LogLevel]
 
 	return corev1.Container{
 		Image:           util.GetImageEnv(),
--- a/pkg/testframework/console-operator.go
+++ b/pkg/testframework/console-operator.go
@@ -57,12 +57,12 @@ func ensureConsoleIsInDesiredState(t *testing.T, client *Clientset, state operat
 }
 
 func ManageConsole(t *testing.T, client *Clientset) error {
-	cr, err := client.Consoles().Get(consoleapi.ConfigResourceName, metav1.GetOptions{})
+	operatorConfig, err := client.Consoles().Get(consoleapi.ConfigResourceName, metav1.GetOptions{})
 	if err != nil {
 		return err
 	}
 
-	if isOperatorManaged(cr) {
+	if isOperatorManaged(operatorConfig) {
 		t.Logf("console operator already in 'Managed' state")
 		return nil
 	}
@@ -81,12 +81,12 @@ func ManageConsole(t *testing.T, client *Clientset) error {
 }
 
 func UnmanageConsole(t *testing.T, client *Clientset) error {
-	cr, err := client.Consoles().Get(consoleapi.ConfigResourceName, metav1.GetOptions{})
+	operatorConfig, err := client.Consoles().Get(consoleapi.ConfigResourceName, metav1.GetOptions{})
 	if err != nil {
 		return err
 	}
 
-	if isOperatorUnmanaged(cr) {
+	if isOperatorUnmanaged(operatorConfig) {
 		t.Logf("console operator already in 'Unmanaged' state")
 		return nil
 	}
@@ -105,12 +105,12 @@ func UnmanageConsole(t *testing.T, client *Clientset) error {
 }
 
 func RemoveConsole(t *testing.T, client *Clientset) error {
-	cr, err := client.Consoles().Get(consoleapi.ConfigResourceName, metav1.GetOptions{})
+	operatorConfig, err := client.Consoles().Get(consoleapi.ConfigResourceName, metav1.GetOptions{})
 	if err != nil {
 		return err
 	}
 
-	if isOperatorRemoved(cr) {
+	if isOperatorRemoved(operatorConfig) {
 		t.Logf("console operator already in 'Removed' state")
 		return nil
 	}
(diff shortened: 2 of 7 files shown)


test/e2e/logging_test.go
```diff
@@ -0,0 +1,102 @@
...
+func isFlagInCommand(t *testing.T, command []string, loggingFlag string) bool {
+	t.Logf("checking if '%s' flag is set on the console deployment container command...", loggingFlag)
+	for _, flag := range command {
+		if flag == loggingFlag {
+			return true
+		}
+	}
+	return false
+}
+
```
benjaminapetersen: Perhaps we should use a map to clarify the `1:1` relationship between the key & value here: 

```golang
type LogMapping map[string]string

var logMap = LogMapping{
   operatorsv1.Normal: "--log-level=*=NOTICE",
   operatorsv1.Debug: "--log-level=*=DEBUG",
   operatorsv1.Trace: "--log-level=*=TRACE",
   operatorsv1.TraceAll: "--log-level=*=TRACE",
}
```
Might even be cleaner to use in the `deployment.go` file & shared here.  Thoughts?

pkg/testframework/console.go
```diff
@@ -147,3 +147,65 @@ func MustRemoveConsole(t *testing.T, client *Clientset) error {
 	}
 	return nil
 }
+
```
benjaminapetersen: So at this point `framework/console.go` is for `console` `operator` functions.  I think we should keep it that way (and perhaps even rename it to `console-operator.go` as its not for top level config stuff.  We should probably put these `funcs` in a `deployments.go` file.  I'm worried that if we get into testing the `console operator` config, then the `console` top level config`, then the `clusteroperator` `console`, then the various resources for the `console` itself, and everything is kept in one file.... we will cry. 😂

pkg/testframework/console.go
```diff
@@ -147,3 +147,65 @@ func MustRemoveConsole(t *testing.T, client *Clientset) error {
...
+	t.Logf("setting console operator to '%s' LogLevel ...", logLevel)
+	_, err = client.Consoles().Patch(consoleapi.ConfigResourceName, types.MergePatchType, []byte(fmt.Sprintf(`{"spec": {"logLevel": "%s"}}`, logLevel)))
+	if err != nil {
+		return err
+	}
+
+	err = wait.PollImmediate(1*time.Second, 1*time.Minute, func() (bool, error) {
+		newCr, err := client.Consoles().Get(consoleapi.ConfigResourceName, metav1.GetOptions{})
+		deployment, err := GetConsoleDeployment(client)
```
benjaminapetersen: This makes me wonder if it would be handy to just have a separate `GenerationChanged(resourceGeneration)` or `CompareGenerations(oldGen, newGen)` function.  I imagine we could use this a good bit in tests, though I don't think we can make it generic to all resources (strongly typed).

test/e2e/logging_test.go
```diff
@@ -0,0 +1,102 @@
...
+		t.Fatalf("error: flag not found in command %v \n", deployment.Spec.Template.Spec.Containers[0].Command)
+	}
+}
+
+// TestTraceLogLevel sets 'TraceAll' LogLevel on the console operator and tests
+// if '--log-level=*=TRACE' flag is set on the console deployment
+func TestTraceAllLogLevel(t *testing.T) {
+	client := testframework.MustNewClientset(t, nil)
+	defer testframework.SetLogLevel(t, client, operatorsv1.Normal)
+	testframework.MustManageConsole(t, client)
```
benjaminapetersen: Do we need the `MustManageConsole`?
Thinking about this.  Ideally all of our tests would be black boxed.  Or perhaps share a `Setup()` func in this suite?

pkg/testframework/console.go
jhadvig: > We should probably put these funcs in a deployments.go file

I would rather keep them in the `testframework` pkg, since all the functions are e2e test specific.

pkg/testframework/console-operator.go
```diff
@@ -57,12 +57,12 @@ func ensureConsoleIsInDesiredState(t *testing.T, client *Clientset, state operat
 }
 
 func ManageConsole(t *testing.T, client *Clientset) error {
-	cr, err := client.Consoles().Get(consoleapi.ConfigResourceName, metav1.GetOptions{})
+	operatorConfig, err := client.Consoles().Get(consoleapi.ConfigResourceName, metav1.GetOptions{})
```
benjaminapetersen: Thanks for the cleanup!

pkg/testframework/console.go
benjaminapetersen: Agree (just meant inside `/pkg/testframework/` keeping things tidy).

benjaminapetersen:
/lgtm

thx!

benjaminapetersen:
```
hack/test-unit.sh
[INFO] Testing github.com/openshift/console-operator/pkg/cmd/operator
pkg/console/subresource/deployment/deployment.go:308:3: undefined: fmt
[ERROR] PID 8: hack/test-unit.sh:27: `go test $PACKAGE` exited with status 2.
```

jhadvig:
```
# github.com/openshift/console-operator/pkg/console/subresource/deployment
pkg/console/subresource/deployment/deployment.go:308:3: undefined: fmt
```
weirdest thing is that after the change there is no line 308 ...
@benjaminapetersen

jhadvig:
@benjaminapetersen rebased the PR and should be green now 🍏

benjaminapetersen:
```
or state to 'Managed'...
=== RUN   TestUnmanaged
--- FAIL: TestUnmanaged (6.34s)
	console-operator.go:94: changing console operator state to 'Unmanaged'...
	framework.go:33: deleting console Deployment...
	framework.go:33: deleting console Service...
	framework.go:33: deleting console Route...
	framework.go:33: deleting console ConfigMap...
	unmanaged_test.go:17: validating that the operator does not recreate deleted resources when ManagementState:Unmanaged...
	unmanaged_test.go:26: error: deleted console Deployment was recreated
```
Hmm.  Wonder why the `Deployment` comes back?

pkg/console/subresource/deployment/deployment.go
```diff
@@ -229,24 +229,20 @@ func consoleVolumeMounts(vc []volumeConfig) []corev1.VolumeMount {
 	return volMountList
 }
 
+type LogMapping map[operatorv1.LogLevel]string
```
benjaminapetersen: I think I realized my suggestion here may also be incorrect, the more indiomatic way to do this is prob like:
https://github.com/openshift/console-operator/pull/212/files#r278728198

benjaminapetersen: a `case` can match several items and is substantially faster.

jhadvig:
@benjaminapetersen comment addressed.
Also added additional commit that should fix the flakes

PTAL

pkg/console/subresource/deployment/deployment.go
```diff
@@ -229,20 +229,26 @@ func consoleVolumeMounts(vc []volumeConfig) []corev1.VolumeMount {
...
-	operatorv1.Trace:    "--log-level=*=TRACE",
-	operatorv1.TraceAll: "--log-level=*=TRACE",
+func GetLogLevelFlag(logLevel operatorv1.LogLevel) string {
+	flag := ""
+	switch logLevel {
+	case operatorv1.Normal:
+		flag = "--log-level=*=NOTICE"
+	case operatorv1.Debug:
+		flag = "--log-level=*=DEBUG"
+	case operatorv1.Trace:
```
benjaminapetersen: Would it be helpful to reuturn both `(string, ok)`?  
Instead of `fallthrough`, use double matching case (this seems to be the pattern I see more often in our codebases):

```golang
// do we validate loglevel on admission? if so, `ok` isn't needed. otherwise could be helpful 
func GetLogLevelFlag(brandstr string) (logLevel string, ok bool) {
   flag := ""
   switch logLevel {
   case operatorv1.Normal:
      flag = "--log-level=*=NOTICE"
   case operatorv1.Debug:
      flag = "--log-level=*=DEBUG"
   // i think this is the preferred idiom
   case operatorv1.Trace, operatorv1.TraceAll:
      flag = "--log-level=*=TRACE"
   }
   return flag
}
```
Though I do not remember offhand if we validate `LogLevel` at the config level. If so, its not possible to get an invalid value, in which case `found` is unnecessary.

pkg/testframework/framework.go
```diff
@@ -137,6 +139,20 @@ func DeleteCompletely(getObject func() (runtime.Object, error), deleteObject fun
 	})
 }
 
+func AreConsoleResourcesAvailable(client *Clientset) error {
```
benjaminapetersen: nit: I think the `golang` idiom is to drop the `is` and `are`, so maybe:

```golang
func AllResourcesAvailable()
func AllResourcesUnavailable()
```

pkg/testframework/framework.go
```diff
@@ -159,6 +175,20 @@ func IsResourceAvailable(errChan chan error, client *Clientset, resource string)
...
 
+func AreConsoleResourcesUnavailable(client *Clientset) error {
+	errChan := make(chan error)
+	go IsResourceUnavailable(errChan, client, "ConfigMap")
+	go IsResourceUnavailable(errChan, client, "Route")
+	go IsResourceUnavailable(errChan, client, "Service")
+	go IsResourceUnavailable(errChan, client, "Deployment")
+	checkErr := <-errChan
+
+	if checkErr != nil {
```
benjaminapetersen: Is the `if` check necessary?  `if` we check `nil`, but return `nil` otherwise, can we just return `checkErr`?

pkg/testframework/console-operator.go
```diff
@@ -77,6 +77,11 @@ func ManageConsole(t *testing.T, client *Clientset) error {
 		return fmt.Errorf("unable to change console operator state to 'Managed': %s", err)
 	}
 
+	err = AreConsoleResourcesAvailable(client)
```
benjaminapetersen: Is this the change that catches the flake?

test/e2e/managed_test.go
```diff
@@ -16,15 +16,9 @@ func TestManaged(t *testing.T) {
...
-	errChan := make(chan error)
-	go testframework.IsResourceAvailable(errChan, client, "ConfigMap")
-	go testframework.IsResourceAvailable(errChan, client, "Route")
-	go testframework.IsResourceAvailable(errChan, client, "Service")
-	go testframework.IsResourceAvailable(errChan, client, "Deployment")
-	checkErr := <-errChan
-
-	if checkErr != nil {
-		t.Fatal(checkErr)
+	err := testframework.AreConsoleResourcesAvailable(client)
```
benjaminapetersen: Cleaner.

pkg/testframework/console-operator.go
jhadvig: Yes ! Should be :)

benjaminapetersen: awesome.

pkg/testframework/framework.go
```diff
@@ -159,15 +172,27 @@ func IsResourceAvailable(errChan chan error, client *Clientset, resource string)
 	errChan <- err
 }
 
+func ConsoleResourcesUnavailable(client *Clientset) error {
```
benjaminapetersen: Ah, yeah we already have `IS` in `IsResourceUnavailable` lol.  Oh well, naming is hard.

pkg/testframework/framework.go
```diff
@@ -137,6 +139,17 @@ func DeleteCompletely(getObject func() (runtime.Object, error), deleteObject fun
 	})
 }
 
+func ConsoleResourcesAvailable(client *Clientset) error {
```
benjaminapetersen: Cool.  Technically we do have `2` `configmap`s, and then a whole additional `Deployment,Service,Route` for downloads, so we may at some point want to work that into our helpers, but not important now.

benjaminapetersen:
/lgtm

nice, happy to see that flake disappear.

benjaminapetersen:
terraform fails

benjaminapetersen:
almost there...

benjaminapetersen:
still waiting....

benjaminapetersen:
/retest

terraform

PR #208, opened 2019-04-18
Reduce default log level

benjaminapetersen:
Default logs should not be `debug` (developer centric), should be `info`.  I hesitated to want to reduce this earlier on, but for development we should all be deploying our own `deployment` of a custom `operator` image based on feature branches anyway, so it doesn't make sense to leave our log levels so high.

This doesn't affect our explicit `logrus` based logging.  We will have to handle that when we swap to `klog`, and ensure the update uses `debug`,`info`, etc.

### DIFF
--- a/manifests/07-operator.yaml
+++ b/manifests/07-operator.yaml
@@ -32,9 +32,8 @@ spec:
         - console
         - operator
         args:
-        - "-v=4"
+        - "-v=2"
         - "--create-default-console=true"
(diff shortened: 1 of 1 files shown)


spadgett:
Do we want to wait until we are sure we're no longer flapping?

jhadvig:
@spadgett weird, before I've tagged the PR for merge all the tests have been green :-/

PR #211, opened 2019-04-23
Use CVO to create operator config

benjaminapetersen:
There is an email from @damemi titled `[aos-devel] Notice: Operator configs should be managed by the CVO` instructing we should make this change.

- Includes the `create-only` annotation
- `CRD` is lexicographically ordered `00-`, `CR` is `01-`.  This is because the CR should be created after the CRD.  The CR being at the same level as the `oauthclient` should be irrelevant.
- Removes our `defaultConsoleOperatorConfig()`, which was used at the start of the `sync` loop if the `GET` on our `config` was a `404`.  The CVO will now be our single source of truth providing the config, we will simply err until it gets recreated.

### DIFF
--- a/manifests/01-operator-config.yaml
+++ b/manifests/01-operator-config.yaml
@@ -1,0 +1,8 @@
+apiVersion: operator.openshift.io/v1
+kind: Console
+metadata:
+  name: cluster
+  annotations:
+    release.openshift.io/create-only: "true"
+spec:
+  managementState: Managed
--- a/pkg/cmd/operator/cmd.go
+++ b/pkg/cmd/operator/cmd.go
@@ -1,7 +1,6 @@
 package operator
 
 import (
-	"github.com/openshift/console-operator/pkg/console/operator"
 	// 3rd party
 	"github.com/spf13/cobra"
 	// kube / openshift
@@ -26,15 +25,5 @@ func NewOperator() *cobra.Command {
 	// https://github.com/spf13/cobra#create-rootcmd
 	cmd.Long = `An Operator for a web console for OpenShift.
 				`
-	cmd.Flags().BoolVarP(
-		&operator.CreateDefaultConsoleFlag,
-		"create-default-console",
-		"d",
-		false,
-		`Instructs the operator to create a console
-        custom resource on startup if one does not exist. 
-        `,
-	)
-
 	return cmd
 }
--- a/pkg/console/operator/operator.go
+++ b/pkg/console/operator/operator.go
@@ -52,8 +52,6 @@ const (
 	controllerName = "Console"
 )
 
-var CreateDefaultConsoleFlag bool
-
 type consoleOperator struct {
 	operatorConfigClient operatorclientv1.ConsoleInterface
 	consoleConfigClient  configclientv1.ConsoleInterface
@@ -140,15 +138,7 @@ func NewConsoleOperator(
 
 // key is actually the pivot point for the operator, which is our Console custom resource
 func (c *consoleOperator) Key() (metav1.Object, error) {
-	operatorConfig, err := c.operatorConfigClient.Get(api.ConfigResourceName, metav1.GetOptions{})
-	if errors.IsNotFound(err) && CreateDefaultConsoleFlag {
-		if _, err := c.operatorConfigClient.Create(c.defaultConsoleOperatorConfig()); err != nil {
-			logrus.Errorf("no console operator config found. Creating. %v \n", err)
-			return nil, err
-		}
-	}
-
-	return operatorConfig, err
+	return c.operatorConfigClient.Get(api.ConfigResourceName, metav1.GetOptions{})
 }
 
 func (c *consoleOperator) Sync(obj metav1.Object) error {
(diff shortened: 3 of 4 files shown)


pkg/console/operator/operator.go
```diff
@@ -140,15 +140,7 @@ func NewConsoleOperator(
 
 // key is actually the pivot point for the operator, which is our Console custom resource
 func (c *consoleOperator) Key() (metav1.Object, error) {
-	operatorConfig, err := c.operatorConfigClient.Get(api.ConfigResourceName, metav1.GetOptions{})
-	if errors.IsNotFound(err) && CreateDefaultConsoleFlag {
```
spadgett: Can we remove the create default console flag?

manifests/01-operator-config.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: operator.openshift.io/v1
+kind: Console
+metadata:
+  name: cluster
+  annotations:
```
benjaminapetersen: For those who missed it, this annotation means "create only, if it does not exist, but leave it alone if it gets updated".

pkg/console/operator/operator.go
benjaminapetersen: We are still creating out top level `console` config.
Just not the `operator` config.

benjaminapetersen: But thats a good Q.  Should both be created by CVO now instead of the operator? If so, we could eliminate the flag.

spadgett: Looks like it's already there?

https://github.com/openshift/cluster-config-operator/blob/master/empty-resources/0000_05_config-operator_02_console.cr.yaml

benjaminapetersen: `/empty-resources` is new to me.  Great. I think we can delete the other `func` as well, and the flag.

benjaminapetersen: Both removed.  We now trust someone else will recreate all of our configs.

spadgett: `CreateDefaultConsoleFlag` is still referenced in a few other places

benjaminapetersen: Yup. Should be eliminated now.

pkg/cmd/operator/cmd.go
```diff
@@ -26,15 +25,5 @@ func NewOperator() *cobra.Command {
 	// https://github.com/spf13/cobra#create-rootcmd
 	cmd.Long = `An Operator for a web console for OpenShift.
 				`
-	cmd.Flags().BoolVarP(
```
benjaminapetersen: Eliminated:
```
console operator -h
An Operator for a web console for OpenShift.

Usage:
  console operator [flags]

Flags:
      --config string            Location of the master configuration file to run from.
  -d, --create-default-console   Instructs the operator to create a console
                                         custom resource on startup if one does not exist.

  -h, --help                     help for operator
      --kubeconfig string        Location of the master configuration file to run from.
```
now:
```
console operator -h
An Operator for a web console for OpenShift.

Usage:
  console operator [flags]

Flags:
      --config string       Location of the master configuration file to run from.
  -h, --help                help for operator
      --kubeconfig string   Location of the master configuration file to run from.
```

pkg/console/operator/operator.go
benjaminapetersen: no longer in the help output:

```
console operator -h
An Operator for a web console for OpenShift.

Usage:
  console operator [flags]

Flags:
      --config string       Location of the master configuration file to run from.
  -h, --help                help for operator
      --kubeconfig string   Location of the master configuration file to run from.
```

spadgett:
You need to remove the flag from the manifest as well:

https://github.com/openshift/console-operator/blob/master/manifests/07-operator.yaml#L36

benjaminapetersen:
Ah, that I did miss (the one in the manifest).

benjaminapetersen:
#206 contains flake fix.

spadgett:
> Also not sure whats the purpose of the console config in the openshift-config-managed namespace? when it's created? why it's not present on the cluster by default ?

That's for OKD clusters to set the default branding. Otherwise we would default to OCP branding on OKD clusters.

spadgett:
```
error: unable to read image registry.svc.ci.openshift.org/ci-op-vb7zcd8b/stable-initial@sha256:51a4b2fcd1b3f2127d6e1c9be8d1fd44d193bf8085f6e71afefb5213dd9215a9: received unexpected HTTP status: 504 Gateway Time-out
```

/retest

spadgett:
```
info: Loading sha256:2818ae6b66af928d478db15ac4d1a5a91f85b91d401a09bbb539d19662c53eda service-ca-operator
info: Included 86 images from 30 input operators into the release
error: received unexpected HTTP status: 504 Gateway Time-out
```

/retest

smarterclayton:
As we get closer to release, please ensure code changes have a bug and the bug is associated in the PR title - follow the conventions described in previous emails about how to associate bugs with PRs.  The PR title must be `Bug XXXX: <description>`.

PR #212, opened 2019-04-23
E2E test specific to branding

zherman0:
E2E tests specific to branding.
1) Check that we can set brand via operator config
2) Check that setting the console config in the openshift-config-managed namespace is honored when a specific brand is chosen.

### DIFF
--- a/test/e2e/branding_test.go
+++ b/test/e2e/branding_test.go
@@ -1,0 +1,198 @@
+package e2e
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	operatorsv1 "github.com/openshift/api/operator/v1"
+	consoleapi "github.com/openshift/console-operator/pkg/api"
+	"github.com/openshift/console-operator/pkg/testframework"
+)
+
+var validOperatorConfigBrands = []operatorsv1.Brand{operatorsv1.BrandOKD, operatorsv1.BrandOCP, operatorsv1.BrandOnline, operatorsv1.BrandDedicated, operatorsv1.BrandAzure}
+var validManagedConfigMapBrands = []operatorsv1.Brand{operatorsv1.BrandOKD, operatorsv1.BrandOCP}
+
+// Test prep - setup the client used by each test
+func setupBrandingTestCase(t *testing.T) (*testframework.Clientset, operatorsv1.Brand, map[string]string) {
+	client := testframework.MustNewClientset(t, nil)
+	// Get the original operator config
+	originalConfig, err := client.Consoles().Get(consoleapi.ConfigResourceName, metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("could not get operator config, %v", err)
+	}
+	originalConfigBrand := originalConfig.Spec.Customization.Brand
+	// Get the original Managed Config Map
+	originalManagedConfigMap, err := client.ConfigMaps(consoleapi.OpenShiftConfigManagedNamespace).Get(consoleapi.OpenShiftConsoleConfigMapName, metav1.GetOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		t.Fatalf("could not get console-config configmap, %v", err)
+	}
+
+	return client, originalConfigBrand, originalManagedConfigMap.Data
+}
+
+func cleanupBrandingTestCase(t *testing.T, client *testframework.Clientset, originalConfigBrand operatorsv1.Brand, originalManagedConfigMapData map[string]string) {
+
+	setOperatorConfigBrand(t, client, originalConfigBrand)
+
+	managedConfigMap := generateTestConfigMap(operatorsv1.BrandOKD)
+	if originalManagedConfigMapData != nil {
+		managedConfigMap.Data = originalManagedConfigMapData
+	}
+	_, err := client.ConfigMaps(consoleapi.OpenShiftConfigManagedNamespace).Update(managedConfigMap)
+	if err != nil {
+		t.Fatalf("could not reset managed config map  %v", err)
+	}
+
+}
+
+// TestOperatorConfigBranding() tests that changing the brand value on the operator-config
+// will result in the brand being set on the console-config in openshift-console.
+// Implicitly it ensures that the operator-config brand overrides brand set on
+// console-config in openshift-config-managed, if the managed configmap exists.
+func TestOperatorConfigBranding(t *testing.T) {
+	client, originalConfigBrand, originalManagedConfigMapData := setupBrandingTestCase(t)
+	defer cleanupBrandingTestCase(t, client, originalConfigBrand, originalManagedConfigMapData)
+	// Set a temporary managed config to test it does not override the operator config values
+	_, err := updateOrCreateConsoleConfigMap(client, generateTestConfigMap(operatorsv1.BrandOKD))
+	if err != nil {
+		t.Fatalf("error: could not apply managed config map %v", err)
+	}
+
+	for _, expectedBrand := range validOperatorConfigBrands {
+		t.Logf("update operator with %v", expectedBrand)
+		// helper to update the operator config
+		setOperatorConfigBrand(t, client, expectedBrand)
+		// now check if it has set the brand
+		err = wait.Poll(1*time.Second, pollTimeout, func() (stop bool, err error) {
+			gotBrand := getConsoleBrand(t, client)
+			return (gotBrand == expectedBrand), nil
+		})
+		if err != nil {
+			t.Fatalf("error: brand was never updated, %v", err)
+		}
+	}
+}
+
+// Test setting brand via the config map in the openshift-config-managed namespace, this requires the operator config not be set
+func TestBrandingFromManagedConfigMap(t *testing.T) {
+	client, originalConfigBrand, originalManagedConfigMapData := setupBrandingTestCase(t)
+	defer cleanupBrandingTestCase(t, client, originalConfigBrand, originalManagedConfigMapData)
+	// Set operator config to empty so it does not override the managed config map values
+	setOperatorConfigBrand(t, client, "")
+
+	for _, expectedBrand := range validManagedConfigMapBrands {
+		t.Logf("update data for the config map in openshift-config-managed namespace with %v", expectedBrand)
+		// Create new configmap for test
+		_, err := updateOrCreateConsoleConfigMap(client, generateTestConfigMap(expectedBrand))
+		if err != nil {
+			t.Fatalf("error: could not apply managed config map %v", err)
+		}
+
+		err = wait.Poll(1*time.Second, pollTimeout, func() (stop bool, err error) {
+			gotBrand := getConsoleBrand(t, client)
+			return (gotBrand == expectedBrand), nil
+		})
+		if err != nil {
+			t.Fatalf("error: brand was never updated, %v", err)
+		}
+	}
+}
+
+func generateTestConfigMap(brand operatorsv1.Brand) *v1.ConfigMap {
+	return &v1.ConfigMap{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "ConfigMap",
+			APIVersion: "v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "console-config",
+			Namespace: "openshift-config-managed",
+		},
+		Data: map[string]string{
+			"console-config.yaml": `kind: ConsoleConfig
+apiVersion: console.openshift.io/v1
+customization:
+  branding: ` + string(brand) + `
+  documentationBaseURL: https://docs.okd.io/4.0/
+`,
+		},
+		BinaryData: nil,
+	}
+}
+
+// Set Brand on the operator config
+func setOperatorConfigBrand(t *testing.T, client *testframework.Clientset, brand operatorsv1.Brand) {
+	operatorConfig, err := client.Consoles().Get(consoleapi.ConfigResourceName, metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("could not get operator config, %v", err)
+	}
+	spec := operatorsv1.ConsoleSpec{
+		OperatorSpec: operatorsv1.OperatorSpec{
+			ManagementState: "Managed",
+		},
+		Customization: operatorsv1.ConsoleCustomization{
+			Brand: brand,
+		},
+	}
+	operatorConfig.Spec = spec
+	_, err = client.Consoles().Update(operatorConfig)
+	if err != nil {
+		t.Fatalf("could not update operator config with brand=%v, %v", brand, err)
+	}
+}
+
+// Get the brand from the console-config in the data of the console CM
+func getConsoleBrand(t *testing.T, client *testframework.Clientset) operatorsv1.Brand {
+	cm, err := testframework.GetConsoleConfigMap(client)
+	if err != nil {
+		t.Fatalf("error: %s", err)
+	}
+
+	data := cm.Data["console-config.yaml"]
+	brandingValue := ""
+	temp := strings.Split(data, "\n")
+	for _, item := range temp {
+		if strings.Contains(item, "branding") {
+			brandingValue = strings.Split(strings.TrimSpace(item), ":")[1]
+		}
+	}
+	brand, _ := stringToBrand(strings.TrimSpace(brandingValue))
+	return brand
+}
+
+// Helper function to convert string version of valid brands to their corresponding constant name
+func stringToBrand(brandstr string) (b operatorsv1.Brand, ok bool) {
+	switch found := operatorsv1.Brand(brandstr); found {
+	case
+		operatorsv1.BrandDedicated,
+		operatorsv1.BrandOnline,
+		operatorsv1.BrandOKD,
+		operatorsv1.BrandOCP,
+		operatorsv1.BrandOpenShift,
+		operatorsv1.BrandAzure:
+		return found, true
+	default:
+		fmt.Printf("invalid brand (%s), defaulting to %s \n", brandstr, operatorsv1.BrandOKD)
+		return operatorsv1.BrandOKD, false
+	}
+}
+
+// Helper function that decides whether to update a config map (if it exists) or create a new one
+func updateOrCreateConsoleConfigMap(client *testframework.Clientset, cm *v1.ConfigMap) (*v1.ConfigMap, error) {
+	// Check if configMap exist so we know whether to update or create
+	_, err := client.ConfigMaps(cm.ObjectMeta.Namespace).Get(cm.Name, metav1.GetOptions{})
+	if errors.IsNotFound(err) {
+		return client.ConfigMaps(cm.ObjectMeta.Namespace).Create(cm)
+	} else if err != nil {
+		return nil, err
+	}
+	return client.ConfigMaps(cm.ObjectMeta.Namespace).Update(cm)
+
+}
--- a/test/e2e/logging_test.go
+++ b/test/e2e/logging_test.go
@@ -32,7 +32,7 @@ func TestDebugLogLevel(t *testing.T) {
 	}
 	flagToTest := deploymentsub.GetLogLevelFlag(operatorsv1.Debug)
 	if !isFlagInCommand(t, deployment.Spec.Template.Spec.Containers[0].Command, flagToTest) {
-		t.Fatalf("error: flag not found in command %v \n", deployment.Spec.Template.Spec.Containers[0].Command)
+		t.Fatalf("error: flag (%s) not found in command %v \n", flagToTest, deployment.Spec.Template.Spec.Containers[0].Command)
 	}
 }
 
@@ -53,7 +53,7 @@ func TestTraceLogLevel(t *testing.T) {
 	}
 	flagToTest := deploymentsub.GetLogLevelFlag(operatorsv1.Trace)
 	if !isFlagInCommand(t, deployment.Spec.Template.Spec.Containers[0].Command, flagToTest) {
-		t.Fatalf("error: flag not found in command %v \n", deployment.Spec.Template.Spec.Containers[0].Command)
+		t.Fatalf("error: flag (%s) not found in command %v \n", flagToTest, deployment.Spec.Template.Spec.Containers[0].Command)
 	}
 }
 
@@ -74,7 +74,7 @@ func TestTraceAllLogLevel(t *testing.T) {
 	}
 	flagToTest := deploymentsub.GetLogLevelFlag(operatorsv1.TraceAll)
 	if !isFlagInCommand(t, deployment.Spec.Template.Spec.Containers[0].Command, flagToTest) {
-		t.Fatalf("error: flag not found in command %v \n", deployment.Spec.Template.Spec.Containers[0].Command)
+		t.Fatalf("error: flag (%s) not found in command %v \n", flagToTest, deployment.Spec.Template.Spec.Containers[0].Command)
 	}
 }
 


test/e2e/managed_test.go
```diff
@@ -61,3 +61,14 @@ func TestEditManagedRoute(t *testing.T) {
 		t.Fatalf("error: %s", err)
 	}
 }
+
```
benjaminapetersen: This test should be in a new file.
The existing ones could prob be better named 

```bash
managementstate_managed_test.go
managementstate_unmanaged_test.go
managementstate_removed_test.go
```
I may suggest creating:

```bash
branding_operator_config_test.go
branding_managed_config_test.go
```
or even just:

```bash 
branding_test.go
```
Either way is fine, just so long as its clear these features do not overlap.

benjaminapetersen: If I were to suggest an approach for the simpler test, I'd go with something like:

```golang
// branding_test.go
// the simpler test first would be branding set via the operator config
func TestOperatorConfigBranding(t *testing.T) {
  online := "online"
  client := testframework.MustNewClientset(t, nil)
  // defer the cleanup
  // we should prob have a Reset() func, to just put the operator config back to defaults
  defer testframework.Reset(t, client, operatorsv1.Normal)
  // helper to update the operator config 
  err := testframework.SetBrand(t, client, online)
   if err != nil {
      t.Fatalf("error: %s", err)
   }
   // now check the console-config in openshift-console & verify it has brand set
  cm, err := testframework.GetConsoleConfigmap(client)
   if err != nil {
      t.Fatalf("error: %s", err)
   }
   if !brandIs(online) {
      t.Fatalf("error: brand was not set to %v", online)
   }
}

// after the above, can do the more complex branding change via the 
// configmap in openshift-config-managed
TestBrandingViaManagedConfigmap() {
   // TODO: later, after the first test works
}
```
`pkg/testframework/console.go` (note that #206 will rename this file to `console-operator.go`)

```golang
// helper funcs
func SetBrand(t *testing.T, client *Clientset, brand string) error {
   // need the config	
   operatorConfig, err := client.Consoles().Get(consoleapi.ConfigResourceName, metav1.GetOptions{})
   if err != nil {
      return err
   }
   // set it
   operatorConfig.Spec.Customization.Brand = brand
   err := client.Consoles().Update(operatorConfig)
   if err != nil {
      return err
   }
   return nil   
}
```
Untested as I wrote this in github, but its a direction.

test/e2e/managed_test.go
```diff
@@ -61,3 +61,14 @@ func TestEditManagedRoute(t *testing.T) {
...
+func TestEditBrandingFromManagedConfigMap(t *testing.T) {
+	client := testframework.MustNewClientset(t, nil)
+	defer testframework.MustManageConsole(t, client)
+	testframework.MustManageConsole(t, client)
+
+	err := mergeAndCheckConfigMap(t, client, true)
```
benjaminapetersen: So, your e2e test shouldn't really know about the merging, that happens in the `operator`.  This func really just needs to:

- `GET` `configmap` `console-config` in `openshift-config-managed`,  set `brand=something`, and `UPDATE/PUT` it back. 
- Wait a second (but the `UPDATE/PUT` is blocking, so should be fine)
- `GET` the `configmap` `console-config` in `openshift-console` and see if it has `brand=something`

test/e2e/util.go
```diff
@@ -90,3 +93,62 @@ func patchAndCheckRoute(t *testing.T, client *testframework.Clientset, isOperato
 	})
 	return err
 }
+
+func mergeAndCheckConfigMap(t *testing.T, client *testframework.Clientset, isOperatorManaged bool) error {
```
benjaminapetersen: Why merging in an e2e test...?

test/e2e/util.go
```diff
@@ -90,3 +93,62 @@ func patchAndCheckRoute(t *testing.T, client *testframework.Clientset, isOperato
...
+	}
+
+	t.Logf("polling for new branding value on the console ConfigMap")
+	err = wait.Poll(5*time.Second, pollTimeout, func() (stop bool, err error) {
+		configMap, err := testframework.GetConsoleConfigMap(client)
+		if err != nil {
+			return true, err
+		}
+		brandingValue := getBrandingValue(originalConfigMap)
+		brandingValue = getBrandingValue(configMap)
```
jhadvig: Not sure whats the point of initialising  the variable and then overwriting it

test/e2e/util.go
```diff
@@ -90,3 +93,62 @@ func patchAndCheckRoute(t *testing.T, client *testframework.Clientset, isOperato
...
+		t.Logf("Branding value in console: %s, in managed CM: %s", brandingValue, expectedBranding)
```
jhadvig: Should start with lowerCase

test/e2e/managed_test.go
jhadvig: +1 for `branding_test.go`

test/e2e/util.go
zherman0: Bad cut and paste.

zherman0:
Still WIP, just did some tweaks to get things better aligned to the first set of recommendations

test/e2e/branding_test.go
```diff
@@ -0,0 +1,81 @@
...
+	gotBrand := testframework.GetBrand(t, client)
+	if !brandsEqual(online, gotBrand) {
+		t.Fatalf("error: brand expected %v but brand got was %v", online, gotBrand)
+	}
+}
+
+func TestBrandingFromManagedConfigMap(t *testing.T) {
+	client := testframework.MustNewClientset(t, nil)
+
+	t.Logf("merging Data on the console ConfigMap")
```
benjaminapetersen: Not merging, right? You are creating or updating a `configmap` in `openshift-config-managed`, then checking to see if the `configmap` in `openshift-console` gets the value (black box, the actual merging is unknown to the e2e test).

test/e2e/branding_test.go
```diff
@@ -0,0 +1,81 @@
...
+			"console-config.yaml": `kind: ConsoleConfig
+apiVersion: console.openshift.io/v1
+customization:
+  branding: ` + expectedBranding + `
+  documentationBaseURL: https://docs.okd.io/4.0/
+`,
+		},
+	}
+
+	_, err := client.ConfigMaps(consoleapi.OpenshiftConfigManagedNamespace).Update(cm)
```
benjaminapetersen: Perhaps extract out the "create or update" logic here into a func.  We use `resourceapply.ApplyConfigMap()` in the actual operator code (see `SyncConfigMap()`), which you could also just use for this (though it requires an event recorder).

pkg/testframework/config.go
```diff
@@ -39,3 +44,40 @@ func GetConfig() (*rest.Config, error) {
...
+	// set it
+	operatorConfig.Spec.Customization.Brand = brand
+	_, err = client.Consoles().Update(operatorConfig)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func GetBrand(t *testing.T, client *Clientset) string {
```
benjaminapetersen: Should the `GetConsoleConfigmap()` happen outside this func?  I would expect `GetBrand(configmap)` to be a helper to return a `Brand`, not sure if it should encapsulate an `http` request as well.

test/e2e/branding_test.go
```diff
@@ -0,0 +1,81 @@
...
+	t.Logf("polling for new branding value on the console ConfigMap")
+	err = wait.Poll(1*time.Second, pollTimeout, func() (stop bool, err error) {
+		gotBrand := testframework.GetBrand(t, client)
+		t.Logf("branding value in console: %s, in managed CM: %s", gotBrand, expectedBranding)
+		return (gotBrand == expectedBranding), nil
+	})
+
+}
+
+// helper funcs
```
benjaminapetersen: You are using both `if !brandsEqual(online, gotBrand)` and also `return (gotBrand == expectedBranding)`.  Is the helper needed?

test/e2e/branding_test.go
zherman0: yeah... that code was not meant to stay in there

pkg/testframework/config.go
zherman0: Well, part of the purpose of the function is to also reduce repeated code. So if I did not do the `GetConsoleConfigMap` inside this function, I would have to call that function before each call to `GetBrand` in each of my tests.  From an OOP standpoint, the getter should be simpler, but here this is a utility function that is getting called from multiple places.

test/e2e/branding_test.go
zherman0: Very good point, I will remove the helper funciton

pkg/testframework/config.go
```diff
@@ -2,10 +2,16 @@ package testframework
 
 import (
 	"fmt"
+	"github.com/openshift/api/operator/v1"
```
benjaminapetersen: ordering of imports

pkg/testframework/config.go
```diff
@@ -2,10 +2,16 @@ package testframework
...
 	"os"
 	"os/user"
 	"path/filepath"
+	"strings"
+	"testing"
 
+	consoleapi "github.com/openshift/console-operator/pkg/api"
```
benjaminapetersen: ordering, `kube` before `openshift`.

test/e2e/branding_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+// Test the customization features of the config map
+
+// the simpler test first would be branding set via the operator config
+func TestOperatorConfigBranding(t *testing.T) {
+	expectedBrand := operatorsv1.BrandDedicated
+	client := testframework.MustNewClientset(t, nil)
+	originalBrand := testframework.GetBrand(t, client)
+	// defer the cleanup
+	defer testframework.PutBackBrand(t, client, originalBrand)
```
benjaminapetersen: Should this just be `defer testframework.SetBrand(orig)` again?  Not sure if `PutBackBrand()` looks like it really adds value?

pkg/testframework/config.go
```diff
@@ -39,3 +45,90 @@ func GetConfig() (*rest.Config, error) {
...
+		// Need to clear the resource version
+		cm.ResourceVersion = ""
+		_, err = client.ConfigMaps(ns).Create(cm)
+		if err != nil {
+			t.Fatalf("Could not recreate original config map %s in namespace %s , %v", cm.Name, ns, err)
+		}
+	}
+}
+
+// Set Brand on the config
```
benjaminapetersen: So I think this naming is confusing.  `SetBrand()` is for the `operator-config`, but `GetBrand()` is for the `console-configmap`.  The `func` names are symetrical, the underlying resources are not.

pkg/testframework/config.go
```diff
@@ -39,3 +45,90 @@ func GetConfig() (*rest.Config, error) {
...
+	case "okd":
+		return v1.BrandOKD
+	case "ocp":
+		return v1.BrandOCP
+	case "openshift":
+		return v1.BrandOpenShift
+	case "azure":
+		return v1.BrandAzure
+	default:
+		return ""
```
benjaminapetersen: Should default throw an error?  Or is this ok?

test/e2e/branding_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/openshift/console-operator/pkg/testframework"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Test the customization features of the config map
+
+// the simpler test first would be branding set via the operator config
+func TestOperatorConfigBranding(t *testing.T) {
+	expectedBrand := operatorsv1.BrandDedicated
```
benjaminapetersen: Would there be value in:

```golang
// loop and test the valid options?
validBrands := []string{"okd", "ocp", "online", "dedicated", "azure"}
// as well as testing a few invalid?
invalidBrands := []string{"redhat", "ninnyhammers", ""}
```
How much would that increase our run time, and would be worth the assertions?

test/e2e/branding_test.go
```diff
@@ -0,0 +1,71 @@
...
+	if expectedBrand != gotBrand {
+		t.Fatalf("brand expected %v but brand got was %v", expectedBrand, gotBrand)
+	}
+}
+
+func TestBrandingFromManagedConfigMap(t *testing.T) {
+	client := testframework.MustNewClientset(t, nil)
+
+	t.Logf("update data for the config map in openshift-config-managed namespace")
+	expectedBranding := operatorsv1.BrandOpenShift
```
benjaminapetersen: Same as above, should we loop the valid options?

test/e2e/branding_test.go
zherman0: wow, yeah, the function started out more involved but got much simpler as I progressed

benjaminapetersen: see https://github.com/openshift/console-operator/pull/206/files#diff-142611e0533ca0e667c4d244cc42c3e4R22

Same pattern is probably sufficient here.

pkg/testframework/config.go
```diff
@@ -39,3 +45,90 @@ func GetConfig() (*rest.Config, error) {
...
+	brandingValue := ""
+	temp := strings.Split(data, "\n")
+	for _, item := range temp {
+		if strings.Contains(item, "branding") {
+			brandingValue = strings.Split(strings.TrimSpace(item), ":")[1]
+		}
+	}
+	return stringToBrand(strings.TrimSpace(brandingValue))
+}
+
```
benjaminapetersen: If going to emulate `ApplyConfigMap()`, I'd prob return `ConfigMap, err` (keep the signature).  Returning the new created or updated configmap could be valuable.

benjaminapetersen: And I would still suggest not combining functions for `operator-config`, top level `console-config`, configmap for `console` all in the same file.   Worried about the potential mess.

pkg/testframework/config.go
```diff
@@ -39,3 +45,90 @@ func GetConfig() (*rest.Config, error) {
...
+	}
+	// set it
+	operatorConfig.Spec.Customization.Brand = brand
+	_, err = client.Consoles().Update(operatorConfig)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
```
benjaminapetersen: I'd move this func, its for the configmap mounted in the console deployment, not the operator config.

pkg/testframework/config.go
benjaminapetersen: how about:
```golang
func stringToBrand(brandstr string) Brand {
	switch found := Brand(brandstr); found {
	case
		BrandDedicated,
		BrandOnline,
		BrandOKD,
		BrandOCP,
		BrandOpenShift,
		BrandAzure:
		return found
	default:
		fmt.Print("erm wut dis? %v", found)
		return BrandOKD
	}
}
```

test/e2e/branding_test.go
zherman0: So I keep trying this and I get through 4 of the 5 options but the fifth always fails with:
`Operation cannot be fulfilled on consoles.operator.openshift.io "cluster": the object has been modified; please apply your changes to the latest version and try again`
I change the order and still the first 4 pass then it errors out.
Additionally I added an invalid option and it does fail with a proper pattern statement:
`spec.customization.brand in body should match '^$|^(ocp|origin|okd|dedicated|online|azure)$`
 but I am not sure writing logic to catch it and not cause the test to fail is simple.

pkg/testframework/config.go
zherman0: yes I agree, these functions are confusing being in one file yet being used for different configs.  I will sort them out into proper place or make new files.

test/e2e/branding_test.go
benjaminapetersen: likely that just means your tests are too fast.  order doesn't really matter, just need a sleep in between them.

pkg/testframework/config.go
```diff
@@ -39,3 +46,89 @@ func GetConfig() (*rest.Config, error) {
 
 	return nil, fmt.Errorf("could not locate a kubeconfig")
 }
+
+// Reset the original cm stored in the openshift-config-managed namespace, if it exist
+func PutBackManagedConfigMap(t *testing.T, client *Clientset, ns string, cm *corev1.ConfigMap) {
```
benjaminapetersen: I'm not sure if you really need this wrapper function, your test function should know if the configmap existed from the first `GET` via `err != nil`, so using `if cm.Name != "" ` as a proxy is less than ideal:

```golang
func TestBrandingFromManagedConfigMap() {
  // I know I got a configmap, or I didn't. best to use the error as a signal and skip the func?
  originalManagedConfigMap, err := client.ConfigMaps(consoleapi.OpenshiftConfigManagedNamespace).Get(consoleapi.OpenShiftConsoleConfigMapName, metav1.GetOptions{})
   if err != nil {
      // no configmap originally
   }
   // make a copy so you can put it back
   configYouWillModify := originalManagedConfigMap.DeepCopy()
   // use update to put it back unmodified orig when done
   defer client.ConfigMaps(consoleapi.OpenshiftConfigManagedNamespace).Update(consoleapi.OpenShiftConsoleConfigMapName, metav1.GetOptions{})
   //
   // do your work...
   // do your work...
   //
}
```  
I think this probably reduces clarity.

test/e2e/branding_test.go
```diff
@@ -0,0 +1,74 @@
...
+func TestOperatorConfigBranding(t *testing.T) {
+	expectedBrand := operatorsv1.BrandDedicated
+	client := testframework.MustNewClientset(t, nil)
+	originalBrand := testframework.GetOperaterConfigBrand(t, client)
+
+	// helper to update the operator config
+	testframework.SetOperaterConfigBrand(t, client, expectedBrand)
+
+	// now check the console-config in openshift-console & verify it has brand set
+	gotBrand := testframework.GetOperaterConfigBrand(t, client)
```
benjaminapetersen: Your comment says `check console-config` brand but you are calling `GetOperatorConfigBrand()`?

I don't think this works.

pkg/testframework/config.go
benjaminapetersen: Different files.

```
// even if you
configmap.go // ApplyConfigmap() is generic for any configmap
operator-config.go
console-config.go
```
I think [this comment](https://github.com/openshift/console-operator/pull/212/files#r278976403) may prove my point.

zherman0:
My test `TestOperatorConfigBranding` is proving to be flaky. Still investigating.

pkg/testframework/config.go
```diff
@@ -39,3 +45,47 @@ func GetConfig() (*rest.Config, error) {
...
+
+	data := cm.Data["console-config.yaml"]
+	brandingValue := ""
+	temp := strings.Split(data, "\n")
+	for _, item := range temp {
+		if strings.Contains(item, "branding") {
+			brandingValue = strings.Split(strings.TrimSpace(item), ":")[1]
+		}
+	}
+	brand, _ := stringToBrand(strings.TrimSpace(brandingValue))
```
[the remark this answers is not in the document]
zherman0: Ben earlier suggested the change to indicate that we are defaulting since a match could not be found.

pkg/testframework/config.go
```diff
@@ -39,3 +45,47 @@ func GetConfig() (*rest.Config, error) {
...
+		v1.BrandDedicated,
+		v1.BrandOnline,
+		v1.BrandOKD,
+		v1.BrandOCP,
+		v1.BrandOpenShift,
+		v1.BrandAzure:
+		return found, true
+	default:
+		fmt.Printf("invalid brand (%s), defaulting to %s \n", brandstr, v1.BrandOKD)
+		return v1.BrandOKD, false
```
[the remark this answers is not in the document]
zherman0: Ben suggested this as default which I think is the right thing to default to.

pkg/testframework/config.go
```diff
@@ -39,3 +45,47 @@ func GetConfig() (*rest.Config, error) {
...
+		}
+	}
+	brand, _ := stringToBrand(strings.TrimSpace(brandingValue))
+	return brand
+}
+
+func ModifyConsoleConfigMap(client *Clientset, ns string, cm *corev1.ConfigMap) (*corev1.ConfigMap, error) {
+	// Check if configMap exist so we know whether to update or create
+	_, err := client.ConfigMaps(ns).Get(cm.Name, metav1.GetOptions{})
+	if err != nil {
```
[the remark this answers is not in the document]
zherman0: @jhadvig - Can you point me to example code to use NotFound?  My error that comes back is a string `configmaps "console-config" not found`.

pkg/testframework/config.go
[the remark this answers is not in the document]
jhadvig: If we decided that defaulting to OKD is the right choice then I'm ok with it :)

pkg/testframework/config.go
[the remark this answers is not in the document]
jhadvig: eg. https://github.com/openshift/console-operator/blob/master/pkg/boilerplate/controller/controller.go#L113

pkg/testframework/config.go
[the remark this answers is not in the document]
jhadvig: ok !

test/e2e/branding_test.go
```diff
@@ -0,0 +1,164 @@
...
+
+// Test the customization-branding features of the config map
+func setupBrandingTestCase(t *testing.T) *testframework.Clientset {
+	client := testframework.MustNewClientset(t, nil)
+	return client
+}
+
+// the simpler test first would be branding set via the operator config
+func TestOperatorConfigBranding(t *testing.T) {
+	validBrands := []operatorsv1.Brand{operatorsv1.BrandOKD, operatorsv1.BrandOCP, operatorsv1.BrandOnline, operatorsv1.BrandDedicated}
```
jhadvig: since this is common for both TCs lets pull it out of them
```go
var validBrands = []operatorsv1.Brand{operatorsv1.BrandOKD, operatorsv1.BrandOCP, operatorsv1.BrandOnline, operatorsv1.BrandDedicated}
```

test/e2e/branding_test.go
```diff
@@ -0,0 +1,164 @@
...
+func TestBrandingFromManagedConfigMap(t *testing.T) {
+	client := setupBrandingTestCase(t)
+	validBrands := []operatorsv1.Brand{operatorsv1.BrandOKD, operatorsv1.BrandOCP, operatorsv1.BrandOnline, operatorsv1.BrandDedicated}
+
+	t.Logf("update data for the config map in openshift-config-managed namespace")
+	for _, expectedBrand := range validBrands {
+		// Keep original config map so we can reset it after test
+		originalManagedConfigMap, err := client.ConfigMaps(consoleapi.OpenshiftConfigManagedNamespace).Get(consoleapi.OpenShiftConsoleConfigMapName, metav1.GetOptions{})
+		if err != nil {
+			defer client.ConfigMaps(consoleapi.OpenshiftConfigManagedNamespace).Update(originalManagedConfigMap)
```
jhadvig: As mentioned before this needs be pulled out of the loop, right after the `setupBrandingTestCase()` func
```go
	client := setupBrandingTestCase(t)
	originalManagedConfigMap, err := client.ConfigMaps(consoleapi.OpenshiftConfigManagedNamespace).Get(consoleapi.OpenShiftConsoleConfigMapName, metav1.GetOptions{})
	if err != nil {
		t.Fatalf("could not get console-config configmap, %v", err)
	}
	defer client.ConfigMaps(consoleapi.OpenshiftConfigManagedNamespace).Update(originalManagedConfigMap)
	for _, expectedBrand := range validBrands {
		t.Logf("update 'console-config' configmap data with %s brand", expectedBrand)
                ...

File: test/e2e/branding_test.go [hunk 26]
```diff
@@ -0,0 +1,164 @@
...
+		}
+	}
+
+}
+
+func TestBrandingFromManagedConfigMap(t *testing.T) {
+	client := setupBrandingTestCase(t)
+	validBrands := []operatorsv1.Brand{operatorsv1.BrandOKD, operatorsv1.BrandOCP, operatorsv1.BrandOnline, operatorsv1.BrandDedicated}
+
+	t.Logf("update data for the config map in openshift-config-managed namespace")
```
jhadvig: Lets pull the log output into the range loop to see which brands are we updating. Check the example beneath...

test/e2e/branding_test.go
```diff
@@ -0,0 +1,164 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	operatorsv1 "github.com/openshift/api/operator/v1"
+	consoleapi "github.com/openshift/console-operator/pkg/api"
+	"github.com/openshift/console-operator/pkg/testframework"
+)
+
+// Test the customization-branding features of the config map
```
jhadvig: This comment doesnt feel right

test/e2e/branding_test.go
```diff
@@ -0,0 +1,164 @@
...
+		},
+	}
+	operatorConfig.Spec = spec
+	_, err = client.Consoles().Update(operatorConfig)
+	if err != nil {
+		t.Fatalf("could not update operator config, %v", err)
+	}
+}
+
+// Set Brand on the config
```
jhadvig: This comment is not right

test/e2e/branding_test.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/openshift/console-operator/pkg/testframework"
+)
+
+// Test the customization-branding features of the config map
+func setupBrandingTestCase(t *testing.T) *testframework.Clientset {
+	client := testframework.MustNewClientset(t, nil)
+	return client
+}
+
+// the simpler test first would be branding set via the operator config
```
jhadvig: This comment does not make much of a sense. Check comments of other TCs for inspiration

test/e2e/branding_test.go
```diff
@@ -0,0 +1,164 @@
...
+		// now check if it has set the brand
+		gotBrand := getOperaterConfigBrand(t, client)
+		if expectedBrand != gotBrand {
+			t.Fatalf("brand expected %v but brand got was %v", expectedBrand, gotBrand)
+		}
+	}
+
+}
+
+func TestBrandingFromManagedConfigMap(t *testing.T) {
```
jhadvig: Lets add comment that would describe the TC so anybody that will look into this will get the feel what it does.

test/e2e/branding_test.go
zherman0: This was a total miscopy/paste - fixed now

test/e2e/branding_test.go
```diff
@@ -0,0 +1,169 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	operatorsv1 "github.com/openshift/api/operator/v1"
+	consoleapi "github.com/openshift/console-operator/pkg/api"
+	"github.com/openshift/console-operator/pkg/testframework"
+)
+
+var validBrands = []operatorsv1.Brand{operatorsv1.BrandOKD, operatorsv1.BrandOCP, operatorsv1.BrandOnline, operatorsv1.BrandDedicated}
```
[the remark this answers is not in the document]
zherman0: I will add Azure.  Openshift is defined but not an accepted value.

test/e2e/branding_test.go
```diff
@@ -0,0 +1,168 @@
...
+		operatorsv1.BrandAzure:
+		return found, true
+	default:
+		fmt.Printf("invalid brand (%s), defaulting to %s \n", brandstr, operatorsv1.BrandOKD)
+		return operatorsv1.BrandOKD, false
+	}
+}
+
+// Helper function that decides whether to update a config map (if it exists) or create a new one
+func modifyConsoleConfigMap(client *testframework.Clientset, ns string, cm *v1.ConfigMap) (*v1.ConfigMap, error) {
```
benjaminapetersen: `modify` implies it would change an existing, not create a new.  Best name?

benjaminapetersen: Comment seems to be needed only because the func name is unclear / implies something different.

zherman0: What about `updateOrCreateConsoleConfigMap`?

zherman0:
@spadgett @benjaminapetersen @jhadvig - I have tried numerous things on my cluster to try to get the error seen in the e2e-aws-operator test failure with no luck. Any ideas?
`test/e2e/branding_test.go:53:53: undefined: "github.com/openshift/console-operator/pkg/api".OpenshiftConfigManagedNamespace`

spadgett:
> @spadgett @benjaminapetersen @jhadvig - I have tried numerous things on my cluster to try to get the error seen in the e2e-aws-operator test failure with no luck. Any ideas?
> `test/e2e/branding_test.go:53:53: undefined: "github.com/openshift/console-operator/pkg/api".OpenshiftConfigManagedNamespace`

You need to rebase. That was renamed to `OpenShiftConfigManagedNamespace`

test/e2e/branding_test.go
```diff
@@ -0,0 +1,168 @@
...
+			t.Fatalf("error: could not apply managed config map %v", err)
+		}
+
+		err = wait.Poll(1*time.Second, pollTimeout, func() (stop bool, err error) {
+			gotBrand := getConsoleBrand(t, client)
+			return (gotBrand == expectedBrand), nil
+		})
+	}
+}
+
```
benjaminapetersen: Test for both?  What if both managed config and operator config are set?  Which one wins?

zherman0: From the console's perspective, the configmap for console comes from the merge that happens in the  configmap->DefaultConfigMap, so the `openshift-config-managed` will win.

benjaminapetersen: Right, that is incorrect actually.

benjaminapetersen: A unit test might be sufficient, however (once fixed).

benjaminapetersen: The configmap in managed is internal. The operator config is an API intended to be used (and have an effect always when set).

spadgett: Right, the config map in openshift-config-managed is just the default when no value is specified. The value in the operator config should win otherwise.

zherman0: I put this on hold until the changes in PR228 merge since this test does work correctly for the current code. PR228 will fix the precedence order for the various configs.

benjaminapetersen:
Squash - tho tests are failing @zherman0

test/e2e/branding_test.go
```diff
@@ -0,0 +1,198 @@
...
+	if err != nil && !errors.IsNotFound(err) {
+		t.Fatalf("could not get console-config configmap, %v", err)
+	}
+
+	return client, originalConfigBrand, originalManagedConfigMap
+}
+
+func cleanupBrandingTestCase(t *testing.T, client *testframework.Clientset, originalConfigBrand operatorsv1.Brand, originalManagedConfigMap *v1.ConfigMap) {
+
+	time.Sleep(1)
```
benjaminapetersen: Did this work consistently or did it introduce more flakes?

zherman0: I reduced this from 5 second to 1 sec with no change to consistency.  I did not try removing it completely.

test/e2e/branding_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+	setOperatorConfigBrand(t, client, originalConfigBrand)
+
+	if originalManagedConfigMap.Name != "" {
+		originalManagedConfigMap.SetResourceVersion("0")
```
benjaminapetersen: `.SetResourceVersion("0")`?

zherman0: The api trhows an error if the ResourceVersion is set when I try doing a create/update.

test/e2e/branding_test.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+
+		err = wait.Poll(1*time.Second, pollTimeout, func() (stop bool, err error) {
+			gotBrand := getConsoleBrand(t, client)
+			return (gotBrand == expectedBrand), nil
+		})
+	}
+}
+
+func createConsoleConfigMap(brand operatorsv1.Brand) *v1.ConfigMap {
```
benjaminapetersen: So I would think `createConsoleConfigMap()` would create the final configmap read by the console.  This actually creates a specific configmap in the `openshift-config-managed` namespace.

We have a lot of configmaps that are similar, I think we need the names to be correct.

test/e2e/branding_test.go
```diff
@@ -0,0 +1,198 @@
...
+func cleanupBrandingTestCase(t *testing.T, client *testframework.Clientset, originalConfigBrand operatorsv1.Brand, originalManagedConfigMap *v1.ConfigMap) {
+
+	time.Sleep(1)
+
+	setOperatorConfigBrand(t, client, originalConfigBrand)
+
+	if originalManagedConfigMap.Name != "" {
+		originalManagedConfigMap.SetResourceVersion("0")
+	} else {
+		originalManagedConfigMap = createConsoleConfigMap(operatorsv1.BrandOKD)
```
benjaminapetersen: Should you be creating a new `configmap` in cleanup?

test/e2e/branding_test.go
benjaminapetersen: I don't think you should have to manipulate this.

test/e2e/branding_test.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+}
+
+// Set Brand on the operator config
+func setOperatorConfigBrand(t *testing.T, client *testframework.Clientset, brand operatorsv1.Brand) {
+	operatorConfig, err := client.Consoles().Get(consoleapi.ConfigResourceName, metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("could not get operator config, %v", err)
+	}
+	// set it
```
benjaminapetersen: nit: can drop this comment, func name & comment on the func already say "set".

test/e2e/branding_test.go
benjaminapetersen: Try clone the latest one, copy over `original.Spec`, then `Update()`?

test/e2e/branding_test.go
benjaminapetersen: Also, `create` seems to imply an `API` call.

test/e2e/branding_test.go
```diff
@@ -0,0 +1,198 @@
...
+		operatorsv1.BrandAzure:
+		return found, true
+	default:
+		fmt.Printf("invalid brand (%s), defaulting to %s \n", brandstr, operatorsv1.BrandOKD)
+		return operatorsv1.BrandOKD, false
+	}
+}
+
+// Helper function that decides whether to update a config map (if it exists) or create a new one
+func updateOrCreateConsoleConfigMap(client *testframework.Clientset, ns string, cm *v1.ConfigMap) (*v1.ConfigMap, error) {
```
benjaminapetersen: I think you can drop `ns string` arg and just use the `cm.ObjectMeta.Namespace`.  No need to pass it twice.

benjaminapetersen: Can you use `resourceapply.ApplyConfigMap()` instead of this func?  See https://github.com/openshift/console-operator/blob/master/pkg/console/operator/sync_v400.go#L228

test/e2e/branding_test.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+}
+
+// Test setting brand using the operator config which overrides the default and managed configmap
+func TestOperatorConfigBranding(t *testing.T) {
+	client, originalConfigBrand, originalManagedConfigMap := setupBrandingTestCase(t)
+	defer cleanupBrandingTestCase(t, client, originalConfigBrand, originalManagedConfigMap)
+	// Set a temporary managed config to test it does not override the operator config values
+	_, err := updateOrCreateConsoleConfigMap(client, consoleapi.OpenShiftConfigManagedNamespace, createConsoleConfigMap(operatorsv1.BrandOKD))
```
benjaminapetersen: So the `console-config` in `openshift-config-managed` is responsible for altering the default `brand`, if the configmap exists.  Do we need to retrieve it in the test for the operator config?  This test just needs to verify that changing the value in the operator config does what we expect to the `console-config` in `openshift-console` namespace, correct?

test/e2e/branding_test.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+}
+
+// Test setting brand via the config map in the openshift-config-managed namespace, this requires the operator config not be set
+func TestBrandingFromManagedConfigMap(t *testing.T) {
+	client, originalConfigBrand, originalManagedConfigMap := setupBrandingTestCase(t)
+	defer cleanupBrandingTestCase(t, client, originalConfigBrand, originalManagedConfigMap)
+	// Set operator config to empty so it does not override the managed config map values
+	setOperatorConfigBrand(t, client, "")
```
benjaminapetersen: I think this test is to check:
- if `console-config` in `openshift-config-managed` does not exist, branding is `ocp`.
- if `console-config` in `openshift-config-managed` does exist,  branding is `okd`.

We don't need the `operator config` correct?  
the `console-config` in `openshift-config-managed` isn't a public API, it should never exist with a different value than `ocp`. We don't need to loop all the brands and change it many times.

test/e2e/branding_test.go
benjaminapetersen: `resourceapply.ApplyConfigMap()` instead of the custom `func`?

test/e2e/branding_test.go
```diff
@@ -0,0 +1,198 @@
...
+		err = wait.Poll(1*time.Second, pollTimeout, func() (stop bool, err error) {
+			gotBrand := getConsoleBrand(t, client)
+			//gotBrand := getOperaterConfigBrand(t, client)
+			return (gotBrand == expectedBrand), nil
+		})
+	}
+
+}
+
+// Test setting brand via the config map in the openshift-config-managed namespace, this requires the operator config not be set
```
benjaminapetersen:
```golang
// this e2e test validates that 
// - if console-config in openshift-config-managed does not exist, branding is okd 
// - if console-config in openshift-config-managed does exist, branding is ocp
func TestBrandingFromManagedConfigMap(t *testing.T) {
	// create client 
	// delete managed config, start from default state of no managed config 
	// check console-config in openshift-console, see if brand is ocp 
	// throw error if incorrect
	// create managed config 
	// check console-config in openshift-console, see if brand is okd 
	// throw error if incorrect
}
```

test/e2e/branding_test.go
```diff
@@ -0,0 +1,198 @@
...
+		t.Fatalf("could not delete managed config map %s, %v", originalManagedConfigMap.GetName(), err)
+	}
+	_, err = client.ConfigMaps(consoleapi.OpenShiftConfigManagedNamespace).Create(originalManagedConfigMap)
+	if err != nil {
+		t.Fatalf("could not reset managed config map %s, %v", originalManagedConfigMap.GetName(), err)
+	}
+
+}
+
+// Test setting brand using the operator config which overrides the default and managed configmap
```
benjaminapetersen:
```golang 
// this e2e test validates that 
// - brand value set via the operator config is set on console-config in openshift-console 
func TestOperatorConfigBranding(t *testing.T) {
	// create client 
	// get operator config
	// loop each $brand
	// set $brand value on operator config 
	// get console-config in openshift-console 
	// check and see if brand is $brand
	// throw error if incorrect 
}
```

test/e2e/branding_test.go
```diff
@@ -0,0 +1,198 @@
...
+			t.Fatalf("error: could not apply managed config map %v", err)
+		}
+
+		err = wait.Poll(1*time.Second, pollTimeout, func() (stop bool, err error) {
+			gotBrand := getConsoleBrand(t, client)
+			return (gotBrand == expectedBrand), nil
+		})
+	}
+}
+
```
benjaminapetersen: We could have a third if we wanted to:

```golang
// this e2e test validates that 
// - if brand is set on console-config in openshift-config-managed
// - and if brand is set on operator-config 
// - operator-config brand wins 
func TestBrandingFromManagedConfigAndOperatorConfig(t *testing.T) { 
	// create client 
	// get managed config 
	// if not found, create default 
	// get operator config 
	// set $brand on operator config 
	// get console-config in openshift-console 
	// check if console-config matches $brand 
	// throw error if incorrect
}
```

test/e2e/branding_test.go
benjaminapetersen: if you want to clarify:
```
// TestOperatorConfigBranding() tests that changing the brand value on the operator-config
// will result in the brand being set on the console-config in openshift-console.
// Implicitly it ensures that the operator-config brand overrides brand set on 
// console-config in openshift-config-managed, if the managed configmap exists.
```

test/e2e/branding_test.go
zherman0: `ApplyConfigMap` expects a ConfigMapsGetter that is available from the consoleOperator object in the sync routines.  The E2E framework does not have that currently and I am not sure how to add it without a bit of code.  I do agree that `updateOrCreateConsoleConfigMap` function is very similar.

test/e2e/branding_test.go
zherman0: Only doing `updates` now

test/e2e/branding_test.go
zherman0: Implemented this smaller update instead of creating a new instance

test/e2e/branding_test.go
zherman0: This is adding an element to the test that verifies the operator config will have higher priority than the managed config. I could separate these into 2 tests but I think this is clean and proves the point that the operator config wins out.

zherman0: From Prior Comment:
ApplyConfigMap expects a ConfigMapsGetter that is available from the consoleOperator object in the sync routines. The E2E framework does not have that currently and I am not sure how to add it without a bit of code. I do agree that updateOrCreateConsoleConfigMap function is very similar.

test/e2e/branding_test.go
zherman0: Removed the wait altogether and it seems robust.

test/e2e/branding_test.go
zherman0: This is what the current version of TestOperatorConfigBranding is doing.

test/e2e/branding_test.go
zherman0: We just need to make sure the `operator-config` is not set to any value or that will override the managed config.

test/e2e/branding_test.go
zherman0: I worry about this test (default is `ocp` when nothing else is set).  The brand is only `ocp` when built as a container with the dockerfile.  If someone builds this manually without the `ADDITIONAL_GOTAGS="ocp"` tags, then the default is actually `okd` and a test like this would fail.  E2E test designed around people building the code with specific tags seems like a bad idea.

zherman0:
./retest

test/e2e/branding_test.go
benjaminapetersen: I'm good with this.

test/e2e/branding_test.go
benjaminapetersen: So the `clientset` is this:
https://github.com/openshift/console-operator/blob/master/pkg/testframework/clientset.go#L15
can compare to how the operator itself is built up:
https://github.com/openshift/console-operator/blob/master/pkg/console/operator/operator.go#L61

`ConfigMaps` are `core`, so I think you have what you need with `clientset.ConfigMaps()`.

test/e2e/branding_test.go
benjaminapetersen: So the `clientset` is this:
https://github.com/openshift/console-operator/blob/master/pkg/testframework/clientset.go#L15
can compare to how the operator itself is built up:
https://github.com/openshift/console-operator/blob/master/pkg/console/operator/operator.go#L61

`ConfigMaps` are `core`, so I think you have what you need with `clientset.ConfigMaps()`.

test/e2e/branding_test.go
benjaminapetersen: Thinking about `ADDITIONAL_GOTAGS="ocp"`.
Honestly trying to remember if we can deprecate that, since we added the `managed configmap`.  I don't know why we want yet another layer of `brand` fiddling.

benjaminapetersen: (Meaning I had thought that `ADDITIONAL_GOTAGS="ocp"` was the old way, and we decided against that because the `managed` configmap allows other components, such as the oauth client, to also retrieve the brand, as a canonical source of truth. )

zherman0: If the `GOTAGS` are deprecated, we should remove that from the Dockerfile in a different PR.  That would make the default `okd` and then the managed config would set it to `ocp`. That work for you guys @benjaminapetersen  @spadgett ?

test/e2e/branding_test.go
```diff
@@ -0,0 +1,192 @@
...
+	if err != nil {
+		t.Fatalf("error: could not apply managed config map %v", err)
+	}
+
+	for _, expectedBrand := range validOperatorConfigBrands {
+		t.Logf("update operator with %v", expectedBrand)
+		// helper to update the operator config
+		setOperatorConfigBrand(t, client, expectedBrand)
+		// now check if it has set the brand
+		err = wait.Poll(1*time.Second, pollTimeout, func() (stop bool, err error) {
```
benjaminapetersen: Need to use the `err`?

`t.Fatalf("error: %s", err)` else I don't think the test fails.

test/e2e/branding_test.go
```diff
@@ -0,0 +1,192 @@
...
+
+	for _, expectedBrand := range validManagedConfigMapBrands {
+		t.Logf("update data for the config map in openshift-config-managed namespace with %v", expectedBrand)
+		// Create new configmap for test
+		_, err := updateOrCreateConsoleConfigMap(client, generateTestConfigMap(expectedBrand))
+		if err != nil {
+			t.Fatalf("error: could not apply managed config map %v", err)
+		}
+
+		err = wait.Poll(1*time.Second, pollTimeout, func() (stop bool, err error) {
```
benjaminapetersen: Same as above, your `t.Fatalf()` is gone, I don't think your test is going to report anything if it fails.

test/e2e/branding_test.go
benjaminapetersen: Ok thats fine. We can discuss removing `GOTAGS` as a follow.  Wanted to make sure we are testing the right things.

test/e2e/branding_test.go
zherman0: Good catch, I somehow deleted that.  Added it back and test all pass.

test/e2e/branding_test.go
zherman0: fixed

benjaminapetersen:
/lgtm

lets get these merged & follow with the `Status.Conditions` check in between tests to ensure we don't flake!

benjaminapetersen:
Weird flake:

```
level=info msg="Waiting up to 30m0s for bootstrapping to complete..."
level=info msg="Destroying the bootstrap resources..."
level=info msg="Waiting up to 30m0s for the cluster at https://api.ci-op-js6t5s5j-ab436.origin-ci-int-aws.dev.rhcloud.com:6443 to initialize..."
level=fatal msg="failed to initialize the cluster: Some cluster operators are still updating: authentication, console: timed out waiting for the condition"
```
Not sure why we are still updating.  Our pod log:

```
W0531 18:29:53.019093       1 reflector.go:270] github.com/openshift/client-go/oauth/informers/externalversions/factory.go:101: watch of *v1.OAuthClient ended with: The resourceVersion for the provided watch is too old.
```
Quite a few "`ended with: too old resource version:`" errors.  We might want to open a bugzilla if we keep getting something like that.

PR #218, opened 2019-05-01
Add public console config to publish console URL

spadgett:
The global config resource is not readable by normal users. Add a public
config map that has the console URL so it can be discovered by all
authenticated users.

### DIFF
--- a/manifests/03-rbac-role-ns-openshift-config-managed.yaml
+++ b/manifests/03-rbac-role-ns-openshift-config-managed.yaml
@@ -12,3 +12,26 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  resourceNames:
+  - console-public
+  verbs:
+  - update
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: console-public
+  namespace: openshift-config-managed
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  resourceNames:
+  - console-public
+  verbs:
+  - get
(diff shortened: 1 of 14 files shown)


manifests/03-rbac-role-ns-openshift-config-managed.yaml
```diff
@@ -12,3 +12,21 @@ rules:
...
   - list
   - watch
+  - create
+  - update
+  - delete
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: console-config-public
```
spadgett: @enj I followed the naming for our existing console operator roles here, but it looks like we should be using a `system:` prefix?

spadgett:
Creates this config map, which is readable to any authenticated user. @deads2k @mfojtik fyi

```yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: console-config-public
  namespace: openshift-config-managed
  selfLink: /api/v1/namespaces/openshift-config-managed/configmaps/console-config-public
  uid: f4a85172-6c11-11e9-a140-060bde40ce6a
  resourceVersion: '1886802'
  creationTimestamp: '2019-05-01T13:06:38Z'
  labels:
    app: console
data:
  consoleURL: 'https://console-openshift-console.apps.spadgett.devcluster.openshift.com'
```

spadgett:
Renamed the config map `console-public`

manifests/03-rbac-role-ns-openshift-config-managed.yaml
```diff
@@ -12,3 +12,21 @@ rules:
   - get
   - list
   - watch
+  - create
```
enj: This role is too broad IMO.

spadgett:
@enj updated

manifests/03-rbac-role-ns-openshift-config-managed.yaml
enj: I think this is fine.  It does not escape the namespace and nothing else needs to bind to this role.

PR #220, opened 2019-05-01
Add custom branding to config map

zherman0:
Add custom product name and logo to the config map and update unit/e2e tests.

For testing:
Creating config map with image (image needs to be less than 1MB)
`oc create configmap myconfig --from-file=/mypic.jpg -n openshift-config`
Creating operator config:
```
apiVersion: operator.openshift.io/v1
kind: Console
metadata:
  annotations:
    release.openshift.io/create-only: "true"
  name: cluster
spec:
  customization:
    brand: azure
    customProductName: my-custom-name
    customLogoFile:
      name:  myconfig
      key: mypic.jpg
  managementState: Managed
```
Update Operator Config:
`oc apply -f custom.yaml`

Verify Changes in Deployment:
`oc get deployment console -n openshift-console -o yaml`

If testing this locally before merge, you will need to make sure console code is the latest along with the console-operator code.
Additionally, you will need to manually run the `oc apply -f manifests/03-rbac-role-ns-openshift-config.yaml ` and `oc apply -f manifests/04-rbac-rolebinding.yaml`

### DIFF
--- a/manifests/03-rbac-role-ns-openshift-config.yaml
+++ b/manifests/03-rbac-role-ns-openshift-config.yaml
@@ -1,0 +1,14 @@
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: console-operator
+  namespace: openshift-config
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - configmaps
+    verbs:
+      - get
+      - list
+      - watch
--- a/pkg/console/operator/operator.go
+++ b/pkg/console/operator/operator.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
 	"github.com/openshift/library-go/pkg/operator/status"
 
 	"monis.app/go/openshift/operator"
@@ -68,7 +69,8 @@ type consoleOperator struct {
 	infrastructureConfigClient configclientv1.InfrastructureInterface
 	versionGetter              status.VersionGetter
 	// recorder
-	recorder events.Recorder
+	recorder       events.Recorder
+	resourceSyncer resourcesynccontroller.ResourceSyncer
 }
 
 func NewConsoleOperator(
@@ -92,6 +94,7 @@ func NewConsoleOperator(
 	versionGetter status.VersionGetter,
 	// recorder
 	recorder events.Recorder,
+	resourceSyncer resourcesynccontroller.ResourceSyncer,
 ) operator.Runner {
 	c := &consoleOperator{
 		// configs
@@ -109,7 +112,8 @@ func NewConsoleOperator(
 		oauthClient:   oauthv1Client,
 		versionGetter: versionGetter,
 		// recorder
-		recorder: recorder,
+		recorder:       recorder,
+		resourceSyncer: resourceSyncer,
 	}
 
 	secretsInformer := coreV1.Secrets()
@@ -132,7 +136,7 @@ func NewConsoleOperator(
 		operator.WithInformer(serviceInformer, targetNameFilter),
 		operator.WithInformer(oauthClients, targetNameFilter),
 		// special resources with unique names
-		operator.WithInformer(configMapInformer, operator.FilterByNames(api.OpenShiftConsoleConfigMapName, api.ServiceCAConfigMapName)),
+		operator.WithInformer(configMapInformer, operator.FilterByNames(api.OpenShiftConsoleConfigMapName, api.ServiceCAConfigMapName, api.OpenShiftCustomLogoConfigMap)),
 		operator.WithInformer(managedConfigMapInformer, operator.FilterByNames(api.OpenShiftConsoleConfigMapName, api.OpenShiftConsolePublicConfigMapName)),
 		operator.WithInformer(secretsInformer, operator.FilterByNames(deployment.ConsoleOauthConfigName)),
 	)
@@ -250,4 +254,36 @@ func (c *consoleOperator) removeConsole(cr *operatorsv1.Console) error {
 	errs = append(errs, updateConfigErr)
 
 	return utilerrors.FilterOut(utilerrors.NewAggregate(errs), errors.IsNotFound)
+}
+
+func (c *consoleOperator) SyncCustomLogoConfigMap(operatorConfig *operatorsv1.Console) error {
+	c.CheckCustomLogoImageStatus(operatorConfig)
+	logoName := operatorConfig.Spec.Customization.CustomLogoFile.Name
+	// add syncing for the custom logo config map
+	if logoName != "" {
+		return c.resourceSyncer.SyncConfigMap(
+			resourcesynccontroller.ResourceLocation{Namespace: api.OpenShiftConsoleNamespace, Name: api.OpenShiftCustomLogoConfigMap},
+			resourcesynccontroller.ResourceLocation{Namespace: api.OpenShiftConfigNamespace, Name: logoName},
+		)
+	}
+	return nil
+}
+
+func (c *consoleOperator) CheckCustomLogoImageStatus(operatorConfig *operatorsv1.Console) {
+	logo := operatorConfig.Spec.Customization.CustomLogoFile
+	if logo.Name != "" && logo.Key != "" {
+		//Check that configmap for custom logo exists
+		logoConfigMap, err := c.configMapClient.ConfigMaps(api.OpenShiftConfigNamespace).Get(logo.Name, metav1.GetOptions{})
+		if err != nil {
+			// Set Operator Status to CustomLogoInvalid
+			c.SetStatusCondition(operatorConfig, operatorsv1.OperatorStatusTypeDegraded, operatorsv1.ConditionTrue, reasonCustomLogoInvalid, "custom logo config map does not exist or has error")
+			klog.Errorf("customLogo configmap not valid, %v", err)
+		} else {
+			if logoConfigMap.BinaryData[logo.Key] == nil {
+				//Key does not exist so activate status condition
+				c.SetStatusCondition(operatorConfig, operatorsv1.OperatorStatusTypeDegraded, operatorsv1.ConditionTrue, reasonCustomLogoInvalid, "custom logo file does not exist ")
+				klog.Errorf("customLogo key is null")
+			}
+		}
+	}
 }
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -109,6 +109,14 @@ func sync_v400(co *consoleOperator, operatorConfig *operatorv1.Console, consoleC
 	}
 	toUpdate = toUpdate || depChanged
 
+	sErr := co.SyncCustomLogoConfigMap(operatorConfig)
+	if sErr != nil {
+		msg := fmt.Sprintf("%q: %v", "customLogoSync", sErr)
+		klog.V(4).Infof("incomplete sync: %v", msg)
+		co.ConditionResourceSyncProgressing(operatorConfig, msg)
+		return sErr
+	}
+
 	resourcemerge.SetDeploymentGeneration(&operatorConfig.Status.Generations, actualDeployment)
 	operatorConfig.Status.ObservedGeneration = operatorConfig.ObjectMeta.Generation
 
--- a/pkg/console/subresource/configmap/configmap.go
+++ b/pkg/console/subresource/configmap/configmap.go
@@ -17,11 +17,7 @@ import (
 
 const (
 	consoleConfigYamlFile = "console-config.yaml"
-)
-
-// overridden by console config
-const (
-	defaultLogoutURL = ""
+	defaultLogoutURL      = ""
 )
 
 func getApiUrl(infrastructureConfig *configv1.Infrastructure) string {
@@ -54,13 +50,14 @@ func DefaultConfigMap(
 		ConfigYAML()
 
 	extractedManagedConfig := extractYAML(managedConfig)
-
 	userDefinedBuilder := &consoleserver.ConsoleServerCLIConfigBuilder{}
 	userDefinedConfig, err := userDefinedBuilder.Host(rt.Spec.Host).
 		LogoutURL(consoleConfig.Spec.Authentication.LogoutRedirect).
 		Brand(operatorConfig.Spec.Customization.Brand).
 		DocURL(operatorConfig.Spec.Customization.DocumentationBaseURL).
 		APIServerURL(getApiUrl(infrastructureConfig)).
+		CustomLogoFile(operatorConfig.Spec.Customization.CustomLogoFile.Key).
+		CustomProductName(operatorConfig.Spec.Customization.CustomProductName).
 		StatusPageID(statusPageId(operatorConfig)).
 		ConfigYAML()
 
--- a/pkg/console/subresource/configmap/configmap_test.go
+++ b/pkg/console/subresource/configmap/configmap_test.go
@@ -198,7 +198,77 @@ providers: {}
 `,
 				},
 			},
-		}, {
+		},
+		{
+			name: "Test operator config with Custom Branding Values",
+			args: args{
+				operatorConfig: &operatorv1.Console{
+					Spec: operatorv1.ConsoleSpec{
+						OperatorSpec: operatorv1.OperatorSpec{},
+						Customization: operatorv1.ConsoleCustomization{
+							Brand:                operatorv1.BrandDedicated,
+							DocumentationBaseURL: mockOperatorDocURL,
+							CustomProductName:    "custom-product-name",
+							CustomLogoFile: configv1.ConfigMapFileReference{
+								Name: "custom-logo-file",
+								Key:  "logo.svg",
+							},
+						},
+					},
+					Status: operatorv1.ConsoleStatus{},
+				},
+				consoleConfig: &configv1.Console{},
+				managedConfig: &corev1.ConfigMap{
+					Data: map[string]string{configKey: `kind: ConsoleConfig
+apiVersion: console.openshift.io/v1
+customization:
+  branding: online
+  documentationBaseURL: https://docs.okd.io/4.1/
+`,
+					},
+				},
+				infrastructureConfig: &configv1.Infrastructure{
+					Status: configv1.InfrastructureStatus{
+						APIServerURL: mockAPIServer,
+					},
+				},
+				rt: &routev1.Route{
+					Spec: routev1.RouteSpec{
+						Host: host,
+					},
+				},
+			},
+			want: &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:        api.OpenShiftConsoleConfigMapName,
+					Namespace:   api.OpenShiftConsoleNamespace,
+					Labels:      map[string]string{"app": api.OpenShiftConsoleName},
+					Annotations: map[string]string{},
+				},
+				Data: map[string]string{configKey: `kind: ConsoleConfig
+apiVersion: console.openshift.io/v1
+auth:
+  clientID: console
+  clientSecretFile: /var/oauth-config/clientSecret
+  oauthEndpointCAFile: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+clusterInfo:
+  consoleBaseAddress: https://` + host + `
+  masterPublicURL: ` + mockAPIServer + `
+customization:
+  branding: ` + string(operatorv1.BrandDedicated) + `
+  documentationBaseURL: ` + mockOperatorDocURL + `
+  customLogoFile: /var/logo/logo.svg
+  customProductName: custom-product-name
+servingInfo:
+  bindAddress: https://0.0.0.0:8443
+  certFile: /var/serving-cert/tls.crt
+  keyFile: /var/serving-cert/tls.key
+providers: {}
+`,
+				},
+			},
+		},
+		{
 			name: "Test operator config with Statuspage pageID",
 			args: args{
 				operatorConfig: &operatorv1.Console{
--- a/pkg/console/subresource/deployment/deployment.go
+++ b/pkg/console/subresource/deployment/deployment.go
@@ -54,33 +54,6 @@ type volumeConfig struct {
 	isConfigMap bool
 }
 
-var volumeConfigList = []volumeConfig{
-	{
-		name:     ConsoleServingCertName,
-		readOnly: true,
-		path:     "/var/serving-cert",
-		isSecret: true,
-	},
-	{
-		name:     ConsoleOauthConfigName,
-		readOnly: true,
-		path:     "/var/oauth-config",
-		isSecret: true,
-	},
-	{
-		name:        api.OpenShiftConsoleConfigMapName,
-		readOnly:    true,
-		path:        "/var/console-config",
-		isConfigMap: true,
-	},
-	{
-		name:        api.ServiceCAConfigMapName,
-		readOnly:    true,
-		path:        "/var/service-ca",
-		isConfigMap: true,
-	},
-}
-
 func DefaultDeployment(operatorConfig *operatorv1.Console, cm *corev1.ConfigMap, serviceCAConfigMap *corev1.ConfigMap, sec *corev1.Secret, rt *routev1.Route) *appsv1.Deployment {
 	labels := util.LabelsForConsole()
 	meta := util.SharedMeta()
@@ -95,6 +68,10 @@ func DefaultDeployment(operatorConfig *operatorv1.Console, cm *corev1.ConfigMap,
 	replicas := int32(ConsoleReplicas)
 	gracePeriod := int64(30)
 	tolerationSeconds := int64(120)
+	volumeConfig := defaultVolumeConfig()
+	if hasCustomLogo(operatorConfig) {
+		volumeConfig = append(volumeConfig, customLogoVolume(operatorConfig))
+	}
 
 	deployment := &appsv1.Deployment{
 		ObjectMeta: meta,
@@ -160,9 +137,9 @@ func DefaultDeployment(operatorConfig *operatorv1.Console, cm *corev1.ConfigMap,
 					TerminationGracePeriodSeconds: &gracePeriod,
 					SecurityContext:               &corev1.PodSecurityContext{},
 					Containers: []corev1.Container{
-						consoleContainer(operatorConfig),
+						consoleContainer(operatorConfig, volumeConfig),
 					},
-					Volumes: consoleVolumes(volumeConfigList),
+					Volumes: consoleVolumes(volumeConfig),
 				},
 			},
 		},
@@ -253,8 +230,8 @@ func GetLogLevelFlag(logLevel operatorv1.LogLevel) string {
 	return flag
 }
 
-func consoleContainer(cr *operatorv1.Console) corev1.Container {
-	volumeMounts := consoleVolumeMounts(volumeConfigList)
+func consoleContainer(cr *operatorv1.Console, volConfigList []volumeConfig) corev1.Container {
+	volumeMounts := consoleVolumeMounts(volConfigList)
 	// Since the console-operator logging has different logging levels then the capnslog,
 	// that we use for console server(bridge) we need to map them to each other
 	flag := GetLogLevelFlag(cr.Spec.LogLevel)
@@ -331,4 +308,47 @@ func IsAvailableAndUpdated(deployment *appsv1.Deployment) bool {
 	return deployment.Status.AvailableReplicas > 0 &&
 		deployment.Status.ObservedGeneration >= deployment.Generation &&
 		deployment.Status.UpdatedReplicas == deployment.Status.Replicas
+}
+
+func defaultVolumeConfig() []volumeConfig {
+	return []volumeConfig{
+		{
+			name:     ConsoleServingCertName,
+			readOnly: true,
+			path:     "/var/serving-cert",
+			isSecret: true,
+		},
+		{
+			name:     ConsoleOauthConfigName,
+			readOnly: true,
+			path:     "/var/oauth-config",
+			isSecret: true,
+		},
+		{
+			name:        api.OpenShiftConsoleConfigMapName,
+			readOnly:    true,
+			path:        "/var/console-config",
+			isConfigMap: true,
+		},
+		{
+			name:        api.ServiceCAConfigMapName,
+			readOnly:    true,
+			path:        "/var/service-ca",
+			isConfigMap: true,
+		},
+	}
+}
+
+func hasCustomLogo(operatorConfig *operatorv1.Console) bool {
+	if logoName := operatorConfig.Spec.Customization.CustomLogoFile.Name; logoName != "" {
+		return true
+	}
+	return false
+}
+
+func customLogoVolume(operatorConfig *operatorv1.Console) volumeConfig {
+	return volumeConfig{
+		name:        api.OpenShiftCustomLogoConfigMap,
+		path:        "/var/logo/",
+		isConfigMap: true}
 }
--- a/pkg/console/subresource/deployment/deployment_test.go
+++ b/pkg/console/subresource/deployment/deployment_test.go
@@ -170,9 +170,9 @@ func TestDefaultDeployment(t *testing.T) {
 							TerminationGracePeriodSeconds: &gracePeriod,
 							SecurityContext:               &corev1.PodSecurityContext{},
 							Containers: []corev1.Container{
-								consoleContainer(consoleOperatorConfig),
+								consoleContainer(consoleOperatorConfig, defaultVolumeConfig()),
 							},
-							Volumes: consoleVolumes(volumeConfigList),
+							Volumes: consoleVolumes(defaultVolumeConfig()),
 						},
 					},
 					Strategy:                appsv1.DeploymentStrategy{},
@@ -252,7 +252,7 @@ func Test_consoleVolumes(t *testing.T) {
 		{
 			name: "Test console volumes creation",
 			args: args{
-				vc: volumeConfigList,
+				vc: defaultVolumeConfig(),
 			},
 			want: []corev1.Volume{
 				{
@@ -326,7 +326,7 @@ func Test_consoleVolumeMounts(t *testing.T) {
 	}{
 		{name: "Test console volumes Mounts",
 			args: args{
-				vc: volumeConfigList,
+				vc: defaultVolumeConfig(),
 			},
 			want: []corev1.VolumeMount{
 				{
--- a/manifests/04-rbac-rolebinding.yaml
+++ b/manifests/04-rbac-rolebinding.yaml
@@ -65,3 +65,17 @@ subjects:
   - kind: Group
     name: system:authenticated
     apiGroup: rbac.authorization.k8s.io
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: console-operator
+  namespace: openshift-config
+roleRef:
+  kind: Role
+  name: console-operator
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+  - kind: ServiceAccount
+    name: console-operator
+    namespace: openshift-console-operator
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -19,4 +19,6 @@ const (
 	OpenShiftConsoleRouteName           = OpenShiftConsoleName
 	OAuthClientName                     = OpenShiftConsoleName
 	OpenShiftConfigManagedNamespace     = "openshift-config-managed"
+	OpenShiftConfigNamespace            = "openshift-config"
+	OpenShiftCustomLogoConfigMap        = "custom-logo"
 )
--- a/pkg/console/clientwrapper/without_secret.go
+++ b/pkg/console/clientwrapper/without_secret.go
@@ -1,0 +1,39 @@
+package clientwrapper
+
+import (
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/fake"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+func WithoutSecret(client kubernetes.Interface) kubernetes.Interface {
+	return &clientWrapper{
+		Interface: client,
+		core: &coreWrapper{
+			CoreV1Interface: client.CoreV1(),
+			secrets:         fake.NewSimpleClientset().CoreV1(),
+		},
+	}
+}
+
+type clientWrapper struct {
+	kubernetes.Interface
+	core *coreWrapper
+}
+
+func (c *clientWrapper) CoreV1() corev1.CoreV1Interface {
+	return c.core
+}
+
+func (c *clientWrapper) Secrets(namespace string) corev1.SecretInterface {
+	return c.core.Secrets(namespace)
+}
+
+type coreWrapper struct {
+	corev1.CoreV1Interface
+	secrets corev1.SecretsGetter
+}
+
+func (c *coreWrapper) Secrets(namespace string) corev1.SecretInterface {
+	return c.secrets.Secrets(namespace)
+}
--- a/pkg/console/operator/status.go
+++ b/pkg/console/operator/status.go
@@ -58,6 +58,7 @@ const (
 	reasonSyncError           = "SynchronizationError"
 	reasonNoPodsAvailable     = "NoPodsAvailable"
 	reasonSyncLoopError       = "SyncLoopError"
+	reasonCustomLogoInvalid   = "CustomLogoInvalid"
 )
 
 // TODO:
--- a/pkg/console/starter/starter.go
+++ b/pkg/console/starter/starter.go
@@ -2,9 +2,6 @@ package starter
 
 import (
 	"fmt"
-
-	"github.com/openshift/api/oauth"
-
 	"os"
 	"time"
 
@@ -17,12 +14,15 @@ import (
 
 	// openshift
 	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/api/oauth"
 	operatorv1 "github.com/openshift/api/operator"
 	"github.com/openshift/console-operator/pkg/api"
-	operatorclient "github.com/openshift/console-operator/pkg/console/operatorclient"
+	"github.com/openshift/console-operator/pkg/console/operatorclient"
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
+	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
 	"github.com/openshift/library-go/pkg/operator/status"
 	"github.com/openshift/library-go/pkg/operator/unsupportedconfigoverridescontroller"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
 
 	// clients
 	configclient "github.com/openshift/client-go/config/clientset/versioned"
@@ -37,6 +37,7 @@ import (
 	routesclient "github.com/openshift/client-go/route/clientset/versioned"
 	routesinformers "github.com/openshift/client-go/route/informers/externalversions"
 
+	"github.com/openshift/console-operator/pkg/console/clientwrapper"
 	"github.com/openshift/console-operator/pkg/console/operator"
 )
 
@@ -88,7 +89,8 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		resync,
 		informers.WithNamespace(api.OpenShiftConfigManagedNamespace),
 	)
-	// configs are all named "cluster", but our clusteroperator is named "console"
+
+	//configs are all named "cluster", but our clusteroperator is named "console"
 	configInformers := configinformers.NewSharedInformerFactoryWithOptions(
 		configClient,
 		resync,
@@ -122,6 +124,8 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 
 	versionGetter := status.NewVersionGetter()
 
+	resourceSyncerInformers, resourceSyncer := getResourceSyncer(ctx, clientwrapper.WithoutSecret(kubeClient), operatorClient)
+
 	// TODO: rearrange these into informer,client pairs, NOT separated.
 	consoleOperator := operator.NewConsoleOperator(
 		// informers
@@ -142,6 +146,7 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		oauthClient.OauthV1(),
 		versionGetter,
 		recorder,
+		resourceSyncer,
 	)
 
 	versionRecorder := status.NewVersionGetter()
@@ -175,6 +180,7 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 	}{
 		kubeInformersNamespaced,
 		kubeInformersManagedNamespaced,
+		resourceSyncerInformers,
 		operatorConfigInformers,
 		configInformers,
 		routesInformersNamespaced,
@@ -184,9 +190,26 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 	}
 
 	go consoleOperator.Run(ctx.Done())
+	go resourceSyncer.Run(1, ctx.Done())
 	go clusterOperatorStatus.Run(1, ctx.Done())
 	go configUpgradeableController.Run(1, ctx.Done())
 
 	<-ctx.Done()
 	return fmt.Errorf("stopped")
+}
+
+func getResourceSyncer(ctx *controllercmd.ControllerContext, kubeClient kubernetes.Interface, operatorClient v1helpers.OperatorClient) (v1helpers.KubeInformersForNamespaces, *resourcesynccontroller.ResourceSyncController) {
+	resourceSyncerInformers := v1helpers.NewKubeInformersForNamespaces(
+		kubeClient,
+		api.OpenShiftConfigNamespace,
+		api.OpenShiftConsoleNamespace,
+	)
+	resourceSyncer := resourcesynccontroller.NewResourceSyncController(
+		operatorClient,
+		resourceSyncerInformers,
+		v1helpers.CachedSecretGetter(kubeClient.CoreV1(), resourceSyncerInformers),
+		v1helpers.CachedConfigMapGetter(kubeClient.CoreV1(), resourceSyncerInformers),
+		ctx.EventRecorder,
+	)
+	return resourceSyncerInformers, resourceSyncer
 }
--- a/pkg/console/subresource/consoleserver/config_builder.go
+++ b/pkg/console/subresource/consoleserver/config_builder.go
@@ -34,8 +34,6 @@ type ConsoleServerCLIConfigBuilder struct {
 	docURL            string
 	apiServerURL      string
 	statusPageID      string
-	// TODO: pipe these in when they are added via the future custom branding PR
-	// This should be trivial
 	customProductName string
 	customLogoFile    string
 }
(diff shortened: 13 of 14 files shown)


zherman0:
Everything expected to fail until api changes are in place
/hold

zherman0:
@spadgett  @benjaminapetersen 
/assign @benjaminapetersen

pkg/console/subresource/deployment/deployment.go
```diff
@@ -321,3 +335,15 @@ func IsAvailableAndUpdated(deployment *appsv1.Deployment) bool {
...
 		deployment.Status.UpdatedReplicas == deployment.Status.Replicas
 }
+
+func AddVolumeForCustomLogo(operatorConfig *operatorv1.Console) {
+	if logoFile := configmap.GetCustomLogoFile(operatorConfig); logoFile != "" {
+		logoPVC := volumeConfig{
+			name:                    CustomLogoPersistentVolumeClaim,
+			readOnly:                true,
+			path:                    logoFile,
+			isPersistentVolumeClaim: true,
```
zherman0: I was thinking of making a pvc to hold the logo, but that might be overkill.

pkg/console/subresource/configmap/configmap.go
```diff
@@ -49,15 +51,17 @@ func DefaultConfigMap(operatorConfig *operatorv1.Console, consoleConfig *configv
...
+	defaultConfig := NewYamlConfig(host, defaultLogoutURL, DEFAULT_BRAND, DEFAULT_DOC_URL, apiServerURL, defaultCustomProductName, defaultCustomLogoFile)
 	// Get console config from the openshift-config-managed namespace
 	extractedManagedConfig := extractYAML(managedConfig)
 
 	// Derive a user defined config using operator config
 	logoutRedirect := consoleConfig.Spec.Authentication.LogoutRedirect
 	brand := operatorConfig.Spec.Customization.Brand
 	docURL := operatorConfig.Spec.Customization.DocumentationBaseURL
-	userDefinedConfig := NewYamlConfig(host, logoutRedirect, brand, docURL, apiServerURL)
+	productName := operatorConfig.Spec.Customization.CustomProductName
```
benjaminapetersen: nit: I'd prob keep `custom` in the var name.  Thats the key thing here.
`customProductName`, or `customName` if you want to keep it short.

pkg/console/subresource/configmap/configmap.go
```diff
@@ -119,7 +123,7 @@ func Stub() *corev1.ConfigMap {
 	return configMap
 }
 
-func NewYamlConfig(host string, logoutRedirect string, brand operatorv1.Brand, docURL string, apiServerURL string) []byte {
+func NewYamlConfig(host string, logoutRedirect string, brand operatorv1.Brand, docURL string, apiServerURL string, customProductName string, customLogoFile string) []byte {
```
benjaminapetersen: nit: This, however, should be generic.  `productName`, `logoFile`.  This `func` has no idea if its custom or default.

pkg/console/subresource/configmap/configmap.go
```diff
@@ -155,14 +159,20 @@ func servingInfo() yaml.MapSlice {
 	}
 }
 
-func customization(brand operatorv1.Brand, docURL string) map[string]string {
+func customization(brand operatorv1.Brand, docURL string, customProductName string, customLogoFile string) map[string]string {
```
benjaminapetersen: nit: same inversion as above, this `func` is lower down the stack, doesn't know if its custom or not.  Just `productName`, and `logoFile`.

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -311,18 +317,20 @@ func TestNewYamlConfig(t *testing.T) {
 		{
 			name: "TestNewYamlConfig",
 			args: args{
-				host:           host,
-				logoutRedirect: "",
-				brand:          DEFAULT_BRAND,
-				docURL:         DEFAULT_DOC_URL,
-				apiServerURL:   mockAPIServer,
+				host:              host,
```
benjaminapetersen: Need to test with both.
- `name: "Test generate default yaml config"`
- `name: "Test generate yaml config with custom product values"`

pkg/console/subresource/deployment/deployment.go
```diff
@@ -334,3 +337,15 @@ func IsAvailableAndUpdated(deployment *appsv1.Deployment) bool {
 		deployment.Status.ObservedGeneration >= deployment.Generation &&
 		deployment.Status.UpdatedReplicas == deployment.Status.Replicas
 }
+
+func AddVolumeForCustomLogo(operatorConfig *operatorv1.Console) {
+	if logoFile := operatorConfig.Spec.Customization.CustomLogoFile.Name; logoFile != "" {
+		logoPVC := volumeConfig{
```
benjaminapetersen: Should this `func` take a `copy` of `volumeConfigList` as an arg so we keep a pristine one around?  Have you tested to ensure this can't append multiple logo file volumes, perhaps if you swap the `logoFile` value several times?

pkg/console/subresource/deployment/deployment_test.go
```diff
@@ -51,22 +51,10 @@ func TestDefaultDeployment(t *testing.T) {
...
-						CreationTimestamp:          metav1.Time{},
-						DeletionTimestamp:          nil,
-						DeletionGracePeriodSeconds: nil,
-						Labels:          labels,
-						Annotations:     nil,
-						OwnerReferences: nil,
-						Initializers:    nil,
-						Finalizers:      nil,
-						ClusterName:     "",
+						Name:              "console-config",
```
benjaminapetersen: Thanks for eliminating the noise!

benjaminapetersen:
Gonna add e2e for this feature?

pkg/console/subresource/deployment/deployment.go
zherman0: Excellent point otherwise the user can never remove the custom logo volume.  I made a working copy and have tested it by passing several different logos in a row and it works well. There can only ever be one mounted logo.

benjaminapetersen: Ok great.

zherman0:
@benjaminapetersen @spadgett - This is ready for a full review.  There are both unit test and E2E tests. 
Here are some results from manual testing.
`oc create configmap myconfig --from-file=/home/zherman/Pictures/small-tractor.jpg -n openshift-config`
The code creates the following in the console-config:
```
 customization:
      branding: azure
      customLogoFile: myconfig
      customProductName: zac-test-1
      documentationBaseURL: https://docs.openshift.com/container-platform/4.1/
```
and the deployment:
```
volumeMounts:
        - mountPath: /var/serving-cert
          name: console-serving-cert
          readOnly: true
        - mountPath: /var/oauth-config
          name: console-oauth-config
          readOnly: true
        - mountPath: /var/console-config
          name: console-config
          readOnly: true
        - mountPath: /var/service-ca
          name: service-ca
          readOnly: true
        - mountPath: /app/small-tractor.jpg
          name: myconfig
          readOnly: true
          subPath: small-tractor.jpg
volumes:
      - name: console-serving-cert
        secret:
          defaultMode: 420
          secretName: console-serving-cert
      - name: console-oauth-config
        secret:
          defaultMode: 420
          secretName: console-oauth-config
      - configMap:
          defaultMode: 420
          name: console-config
        name: console-config
      - configMap:
          defaultMode: 420
          name: service-ca
        name: service-ca
      - configMap:
          defaultMode: 420
          name: myconfig
        name: myconfig
```

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -64,6 +64,28 @@ servingInfo:
...
+  clientSecretFile: /var/oauth-config/clientSecret
+  logoutRedirect: ""
+  oauthEndpointCAFile: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+clusterInfo:
+  consoleBaseAddress: https://` + host + `
+  consoleBasePath: ""
+  masterPublicURL: ` + mockAPIServer + `
+customization:
+  branding: ` + DEFAULT_BRAND + `
+  customLogoFile: custom-logo-file
```
spadgett: I'd use a more realistic value like

```suggestion
  customLogoFile: /var/logo/logo.svg
```

pkg/console/subresource/deployment/deployment.go
```diff
@@ -334,3 +340,17 @@ func IsAvailableAndUpdated(deployment *appsv1.Deployment) bool {
 		deployment.Status.ObservedGeneration >= deployment.Generation &&
 		deployment.Status.UpdatedReplicas == deployment.Status.Replicas
 }
+
+func AddVolumeForCustomLogo(operatorConfig *operatorv1.Console, volConfigList []volumeConfig) []volumeConfig {
+	if logoFile := operatorConfig.Spec.Customization.CustomLogoFile.Name; logoFile != "" {
+		logoVolume := volumeConfig{
+			name:        logoFile,
+			readOnly:    true,
+			path:        "/app/" + operatorConfig.Spec.Customization.CustomLogoFile.Key,
```
spadgett: This shouldn't include the filename.

```suggestion
			path:        "/var/logo",
```

pkg/console/subresource/deployment/deployment.go
```diff
@@ -334,3 +340,17 @@ func IsAvailableAndUpdated(deployment *appsv1.Deployment) bool {
...
+			subPath:     operatorConfig.Spec.Customization.CustomLogoFile.Key,
```
spadgett: You shouldn't need a `subPath`

pkg/console/subresource/configmap/configmap.go
```diff
@@ -59,7 +61,9 @@ func DefaultConfigMap(operatorConfig *operatorv1.Console, consoleConfig *configv
 	brand := operatorConfig.Spec.Customization.Brand
 	docURL := operatorConfig.Spec.Customization.DocumentationBaseURL
 	providers := operatorConfig.Spec.Providers
-	userDefinedConfig := NewYamlConfig(host, logoutRedirect, brand, docURL, apiServerURL, providers)
+	customProductName := operatorConfig.Spec.Customization.CustomProductName
+	customLogoFile := operatorConfig.Spec.Customization.CustomLogoFile.Name
```
spadgett: Don't you need to append the key to this path to get the full path?

pkg/console/subresource/deployment/deployment.go
```diff
@@ -27,6 +27,7 @@ const (
 	publicURLName          = "BRIDGE_DEVELOPER_CONSOLE_URL"
 	ConsoleServingCertName = "console-serving-cert"
 	ConsoleOauthConfigName = "console-oauth-config"
+	ConsoleCustomLogo      = "console-custom-logo"
```
benjaminapetersen: is this used?

pkg/console/subresource/deployment/deployment.go
```diff
@@ -97,6 +99,9 @@ func DefaultDeployment(operatorConfig *operatorv1.Console, cm *corev1.ConfigMap,
 	replicas := int32(ConsoleReplicas)
 	gracePeriod := int64(30)
 	tolerationSeconds := int64(120)
+	workingVolumeConfigList := make([]volumeConfig, len(volumeConfigList))
```
benjaminapetersen: Couple things here that might make this a little cleaner & easier to follow:

```golang
func DefaultDeployment() {
   // ....
   // build up what you need at the top level seems cleaner
   volumeConfig := defaultVolumeConfig()
   if hasCustomLogo(operatorConfig) {
      volumeConfig = append(volumeConfig, customLogoVolume(operatorConfig))
   }
   // finish all the building logic here?  
   volumeMounts := consoleVolumeMounts(volumeConfig)
   volumes := consoleVolumes(volumeConfig)
   //
   // use as you would...
   // ....
}

// wrap the list in a func & return it is easier/clearer than all the make() stuff 
func defaultVolumeConfig() {
    return []volumeConfig{ ... the current volumeConfigList... }
}

// a bool helper for the if statement 
func hasCustomLogo(operatorConfig) bool {}

// keep this simple
// doesn't need an internal append() nor does it need if checks on operator config
// call it, or don't, see below
func customLogoVolume(operatorConfig) {
   return volumeConfig{ ...for the custom logo... }
}

```

pkg/console/subresource/deployment/deployment.go
```diff
@@ -334,3 +340,17 @@ func IsAvailableAndUpdated(deployment *appsv1.Deployment) bool {
 		deployment.Status.ObservedGeneration >= deployment.Generation &&
 		deployment.Status.UpdatedReplicas == deployment.Status.Replicas
 }
+
+func AddVolumeForCustomLogo(operatorConfig *operatorv1.Console, volConfigList []volumeConfig) []volumeConfig {
```
benjaminapetersen: See above comment, would simplify this func.

test/e2e/customization_test.go
```diff
@@ -0,0 +1,176 @@
+package e2e
```
benjaminapetersen: `brand_customization_test.go`?  Just to be clear on what we are customizing.

benjaminapetersen:
/hold

We should hold until we have confirmed that the pipes work & the custom logo appears in the console.

test/e2e/customization_test.go
```diff
@@ -0,0 +1,176 @@
...
+		t.Fatalf("could not delete test configmap, %v", err)
+	}
+
+}
+
+// TestOperatorConfigCustomization() tests that changing the customization values on the operator-config
+// will result in the customization being set on the console-config in openshift-console.
+// Implicitly it ensures that the operator-config customization overrides customization set on
+// console-config in openshift-config-managed, if the managed configmap exists.
+func TestOperatorConfigCustomization(t *testing.T) {
```
benjaminapetersen: Contemplating if we need to check that the console pods are happy, or if they crashloop, since we can't easily open a browser & verify the logo appeared.

This was likely a good question for the previous `providers_test.go` as well, but we didn't think about it.  I would expect if we did want to do this as a basic "it didn't fall over" check, it could be a write-once-use-everywhere shared func.

pkg/console/subresource/deployment/deployment.go
```diff
@@ -162,9 +167,9 @@ func DefaultDeployment(operatorConfig *operatorv1.Console, cm *corev1.ConfigMap,
 					TerminationGracePeriodSeconds: &gracePeriod,
 					SecurityContext:               &corev1.PodSecurityContext{},
 					Containers: []corev1.Container{
-						consoleContainer(operatorConfig),
+						consoleContainer(operatorConfig, workingVolumeConfigList),
```
benjaminapetersen: We need to verify that changes here cause new pods to roll out.  I would assume so, but as I'm tinkering with this, I don't see it happening.  

We might have to pull the `meta.Annotations[customLogoFile] = <hackery>` trick if its not automatic.

spadgett: Changes to the pod template trigger a rollout.

pkg/console/subresource/deployment/deployment.go
```diff
@@ -334,3 +340,17 @@ func IsAvailableAndUpdated(deployment *appsv1.Deployment) bool {
...
+
+func AddVolumeForCustomLogo(operatorConfig *operatorv1.Console, volConfigList []volumeConfig) []volumeConfig {
+	if logoFile := operatorConfig.Spec.Customization.CustomLogoFile.Name; logoFile != "" {
+		logoVolume := volumeConfig{
+			name:        logoFile,
+			readOnly:    true,
+			path:        "/app/" + operatorConfig.Spec.Customization.CustomLogoFile.Key,
+			subPath:     operatorConfig.Spec.Customization.CustomLogoFile.Key,
+			isConfigMap: true,
+		}
```
benjaminapetersen: Also, this `configmap` exists in `openshift-config`. I don't believe it will be loaded w/o some code.

pkg/console/subresource/configmap/configmap.go
benjaminapetersen: Yup:

```
 oc logs -f console-6694c875d5-rp8qw
2019/06/6 22:04:01 cmd/main: could not read logo file: stat my-logo-file: no such file or directory
```

zherman0: done

pkg/console/operator/sync_v400.go
```diff
@@ -301,6 +301,22 @@ func SyncConfigMap(co *consoleOperator, recorder events.Recorder, operatorConfig
 		return nil, false, mcErr
 	}
 
+	// Check if a custom logo is set on the operator config
```
benjaminapetersen: The problem here is that our `watches` are on everything but this resource.  So if the user edits the `console operator config` first, then creates the `configmap` after, it will be too late, the logo configmap won't get picked up.  We will have a timing issue.

We need to make sure we don't miss the update.  That means we need to setup a watch.

In `starter.go` we can use the `resourceSyncer` (since we know for certain we need this resource copied into our namespace) to accomplish the watch & the copy:

```golang
     resourceSyncerInformers := v1helpers.NewKubeInformersForNamespaces(
          // just the configmap, in the config namespace
     )

     resourceSyncer := resourcesynccontroller.NewResourceSyncController(
          // see:
          // https://github.com/openshift/cluster-authentication-operator/blob/master/pkg/operator2/starter.go#L89
          // if you need an example for setting this up 
     )

    if err := resourceSyncer.SyncConfigMap(
        // sync it into the console namespace
        // we are already watching configmaps in the console namespace, so our sync loop will kick when 
        // it picks up changes
    )
```

manifests/03-rbac-role-ns-openshift-config-managed.yaml
```diff
@@ -35,3 +35,16 @@ rules:
   - console-public
   verbs:
   - get
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: console-operator
+  namespace: openshift-config
```
benjaminapetersen: This should be a different file, `ns-openshift-config.yaml`, since its a different namespace than `ns-openshift-config-managed.yaml`.  The `-ns-` indicates the namespace.

benjaminapetersen: I don't know if this is exactly what you will need with the `syncer`.  I imagine you might need `list`, `watch` as well.

pkg/console/subresource/configmap/configmap.go
```diff
@@ -59,7 +61,13 @@ func DefaultConfigMap(operatorConfig *operatorv1.Console, consoleConfig *configv
 	brand := operatorConfig.Spec.Customization.Brand
 	docURL := operatorConfig.Spec.Customization.DocumentationBaseURL
 	providers := operatorConfig.Spec.Providers
-	userDefinedConfig := NewYamlConfig(host, logoutRedirect, brand, docURL, apiServerURL, providers)
+	customProductName := operatorConfig.Spec.Customization.CustomProductName
+	var customLogoFile string
+	if operatorConfig.Spec.Customization.CustomLogoFile.Key != "" {
```
benjaminapetersen: If it doesn't exist, you might get a `Pointer error` here.

pkg/console/subresource/deployment/deployment.go
benjaminapetersen: Right, it should work.  But when I was running the code, I had to manually delete pods to see updates.  So something isn't quite wired up correctly.

benjaminapetersen: Also, we should ensure this works:

```bash
// user gives us a logo
oc create configmap my-logo-file --from-file ~/Desktop/fake-logo.png -n openshift-console
oc create configmap my-logo-file --from-file ~/Desktop/another/dirfake-logo.png -n openshift-console
```
Either way, the configmap will generate:

```
binaryData:
  fake-logo.png: iVBORw0KGgoAAA.....5ErkJggg==
```
The `fake-logo.png` will be the same, but the image will be different.  When we create the volume mount this will probably not trigger a pod rollout.

zherman0: Correct, changing the key or image file will not change the deployment and therefore not force a rollout of the console pods. Changing the configmap name or simply deleting the deployment does work, but that would be a manual step for the user.

spadgett: We shouldn't need to trigger a rollout in the case when only the image itself changes. The mounted config map will update in the pod, and we server the logo file directly from the filesystem.

zherman0: I was able to verify (by manually moving the config map to the openshift-console ns for now) that changing the image in the configmap will show up in the pod and then on the console UI.

pkg/console/operator/sync_v400.go
zherman0: The resourceSyncer requires the secret getter to be set. If that is reasonable, then I can go down this path.  I am looking at the kubeInformer logic, which is what I used for the configmaps in openshift-config-manged.

zherman0: I created a solution using the `informer`.  Check it out and we can discuss if needed.

pkg/console/subresource/deployment/deployment.go
zherman0: The configmap now gets copied automatically using an `informer` to trigger the update.

test/e2e/customization_test.go
zherman0: Yes, we should test not only the console pods but also the console-operator pods since I noticed the e2e will run and all fail when the pods are having an issue (which took me a second to realize what was going on).

zherman0:
/assign @benjaminapetersen
/hold cancel
This is working end to end now so let's review it for real.

pkg/boilerplate/controller/filter.go
```diff
@@ -67,3 +67,17 @@ func FilterByNames(parentFunc ParentFunc, names ...string) ParentFilter {
 		DeleteFunc: has,
 	}
 }
+
```
benjaminapetersen: We treat `pkg/boilerplate` like a library, should not add arbitrary funcs to it.  We have a JIRA to bump it:
https://jira.coreos.com/browse/CONSOLE-1497

Do we need this to live somewhere?

benjaminapetersen: The above JIRA references doing an update by extracting our `pkg/boilerplate` and pulling it in as a dependency, where it now lives:
https://github.com/enj/go

pkg/console/operator/sync_v400.go
```diff
@@ -408,3 +410,31 @@ func getDeploymentGeneration(co *consoleOperator) int64 {
 	}
 	return deployment.Generation
 }
+
+func copyCustomLogoConfigMap(co *consoleOperator, operatorConfig *operatorv1.Console) {
+	// Check if a custom logo is set on the operator config
+	logoName := operatorConfig.Spec.Customization.CustomLogoFile.Name
```
benjaminapetersen: This should return an `error` like other funcs that make API calls so it can bubble up.

pkg/console/operator/sync_v400.go
```diff
@@ -408,3 +410,31 @@ func getDeploymentGeneration(co *consoleOperator) int64 {
...
+	if logoName != "" {
```
benjaminapetersen: You can return early to avoid some of the nesting here.

pkg/console/operator/sync_v400.go
```diff
@@ -301,6 +301,8 @@ func SyncConfigMap(co *consoleOperator, recorder events.Recorder, operatorConfig
 		return nil, false, mcErr
 	}
 
+	copyCustomLogoConfigMap(co, operatorConfig)
```
benjaminapetersen: With this approach, I don't know that I would bury it down inside `SyncConfigMap()`.  Any reason to not make the call a sibling to `SyncConfigMap()` and `SyncServiceCAConfigMap()` a level up?

pkg/console/operator/sync_v400.go
```diff
@@ -408,3 +410,31 @@ func getDeploymentGeneration(co *consoleOperator) int64 {
 	}
 	return deployment.Generation
 }
+
+func copyCustomLogoConfigMap(co *consoleOperator, operatorConfig *operatorv1.Console) {
```
spadgett: Isn't there a resource syncer in library-go that we can use?

spadgett: I see this is discussed above, but I'm unclear why we can't use it.

benjaminapetersen: I would follow the pattern we have set:

```golang
func SyncCustomLogoConfigMap(co *consoleOperator, operatorConfig *operatorv1.Console) (*corev1.ConfigMap, bool, error) {
```
Caller can ignore the `configmap`, but `errors` should bubble.

pkg/boilerplate/controller/filter.go
```diff
@@ -67,3 +67,17 @@ func FilterByNames(parentFunc ParentFunc, names ...string) ParentFilter {
...
+func AcceptAll(parentFunc ParentFunc) ParentFilter {
+	has := func(obj v1.Object) bool {
+		return true
+	}
```
benjaminapetersen: This will be noisy... do we want to kick our sync loop for ANY config changed in the openshift config namespace?

pkg/console/subresource/deployment/deployment.go
```diff
@@ -334,3 +311,46 @@ func IsAvailableAndUpdated(deployment *appsv1.Deployment) bool {
...
+			name:        api.ServiceCAConfigMapName,
+			readOnly:    true,
+			path:        "/var/service-ca",
+			isConfigMap: true,
+		},
+	}
+}
+
+func hasCustomLogo(operatorConfig *operatorv1.Console) bool {
+	if operatorConfig.Spec.Customization.CustomLogoFile.Name != "" {
```
benjaminapetersen: Does this error if `CustomLogoFile` is `nil`?  See previous comment on this func.

pkg/console/operator/sync_v400.go
```diff
@@ -408,3 +410,31 @@ func getDeploymentGeneration(co *consoleOperator) int64 {
...
+	logoName := operatorConfig.Spec.Customization.CustomLogoFile.Name
+	if logoName != "" {
+		// If it exist, copy the configMap in openshift-config into openshift-console
+		customLogoConfigMap, logoErr := co.configMapClient.ConfigMaps(api.OpenShiftConfigNamespace).Get(logoName, metav1.GetOptions{})
+		if logoErr != nil && !apierrors.IsNotFound(logoErr) {
+			logrus.Errorf("custom logo configMap error: %v \n", logoErr)
+		} else if apierrors.IsNotFound(logoErr) {
+			logrus.Errorf("warning: custom logo config map name is set but configmap is not found %v \n", logoErr)
+		} else {
+			customLogoConfigMap.Namespace = "openshift-console"
```
benjaminapetersen: Can you choose to do something like this instead:

```golang
	logoConfigMap := &corev1.ConfigMap{
		ObjectMeta: metav1.ObjectMeta{
                         // We have to honor the ConfigMapNameRef in our operator config, 
                         // but we can copy it into `openshift-console` and name it something stable.
                         //  Could this simplify some of the logic?
			Name:     "user-custom-logo-configmap",
			Namespace: Namespace:   api.OpenShiftConsoleNamespace,
		},
                 // copy over the data section instead of manipulating the original?
		Data: OriginalConfigMapFromOpenShiftConfig.Data
	}
```

benjaminapetersen: I still think whenever you resort to clearing the resource version:

```golang
customLogoConfigMap.ResourceVersion = ""
```
the approach is incorrect.

benjaminapetersen:
Instead of burying that copy func way down deep, Can we can pull it up to the top:

```golang
// starter.go 

// with the loops we already have 
consoleOperator := operator.NewConsoleOperator( .... )
clusterOperatorStatus := status.NewClusterOperatorStatusController( .... )
configUpgradeableController := unsupportedconfigoverridescontroller.NewUnsupportedConfigOverridesController( ....)

// we add a new controller for our copy efforts
configCopyController := operator.ConfigCopyController( 
   // give it the informer to watch the openshift-config namespace for configmaps 
   // and client to copy the configmap to openshift-console, if its the correct configmap
   // based on our operator config.
   // it can be triggered when a configmap change happens, but decide if its the correct configmap
   // if so, copy the configmap into the openshift-console namespace
   // (this can even be to a hard-coded configmap name, if you see my comment above, to reduce some plumbing)
   // then the above `clusterOperator` sync loop we already have doesn't have to do much more 
   // to use the custom-logo configmap if it exists.
)

```

benjaminapetersen:
For documentation purposes, If we want to use the `resourceSyncer`, we can prob do something like this:

1. Set it up as its own sync loop ~
1. `kubeClient` is `k8s.io/client-go/kubernetes.Interface`
    1. write a wrapper to return `Secrets` as a dummy client, no RBAC needed then.
1. Set up the `resourceSyncer` in starter.go, mostly copy-paste from another operator such as [auth operator](https://github.com/openshift/cluster-authentication-operator/blob/master/pkg/operator2/starter.go#L89).  It can watch all `configmap`s in `openshift-config`, but we will only want it to copy the one we care about (so most will be a skip).
1. pass the syncer to the `NewConsoleOperator()` func.  
1.  in our `handleSync()` func, add a call to something like `handleConfigMapSync()` (will look a bit like `handleOAuthConfig()` in auth operator), in there, you can call `c.resourceSyncer.SyncConfigMap` but give it the updated source.  We would use this to check the `operator config` for our `configmapnamereference`, and swap in the named `configmap` to sync.
1. if this works, our current sync watching the `openshift-console` namespace will already see any new configmaps added to the namespace.

manifests/03-rbac-role-ns-openshift-config.yaml
```diff
@@ -0,0 +1,15 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: console-operator
+  namespace: openshift-config
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - configmaps
+      - secrets
```
benjaminapetersen: We don't want `secrets`, just `configmaps`.

pkg/console/subresource/deployment/deployment.go
zherman0: Code now uses a `ResourceSyncer` instead of an `informer`

knobunc:
Syntax error -- https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_console-operator/220/pull-ci-openshift-console-operator-master-images/888

manifests/03-rbac-role-ns-openshift-config.yaml
zherman0: The deafult code for ResourceSync expects a secretsGetter and that getter will only work with RBAC to the namespaces.  I am working on a way around this.

zherman0:
So, I have been trying to remove the RBAC requirement of the ResourceSyncer needing list/watch privileges for Secrets in the `openshift-config`.  I created a `dummySecretGetter` that had the list/watch functions do basically no-ops but I was still getting an error that I needed permissions to the `openshift-config` namespace for secrets.
I then noticed that the ResourceSyncController does the following:
```
informers := kubeInformersForNamespaces.InformersFor(namespace)
informers.Core().V1().ConfigMaps().Informer().AddEventHandler(c.eventHandler())
informers.Core().V1().Secrets().Informer().AddEventHandler(c.eventHandler())
```
https://github.com/openshift/console-operator/blob/7fb2cc8217fb7c80060817c5f51fbed72ede0216/vendor/github.com/openshift/library-go/pkg/operator/resourcesynccontroller/resourcesync_controller.go#L86

Won't that last line make the informer check for secrets no matter how I override the SecretGetter?
@spadgett @benjaminapetersen @jhadvig

benjaminapetersen:
@zherman0 So had a chat with @enj about this to leverage it but avoid the problem with secrets.  To pass along the relevant bits:

Borrowing this code from the authentication operator:

```golang
    resourceSyncerInformers := v1helpers.NewKubeInformersForNamespaces(
        kubeClient,
        targetNamespace,
        userConfigNamespace,
        machineConfigNamespace,
    )

    operatorClient := &OperatorClient{
        authOperatorConfigInformers,
        authConfigClient.OperatorV1(),
    }

    resourceSyncer := resourcesynccontroller.NewResourceSyncController(
        operatorClient,
        resourceSyncerInformers,
        v1helpers.CachedSecretGetter(kubeClient.CoreV1(), resourceSyncerInformers),
        v1helpers.CachedConfigMapGetter(kubeClient.CoreV1(), resourceSyncerInformers),
        ctx.EventRecorder,
    )
```
Then:
1. `kubeClient` is `k8s.io/client-go/kubernetes.Interface`
2. Write a wrapper interface, so that when it tries to access the `Secrets`we return a dummy fake client
3.  Proceed to use it to access `configmaps`, it will not attempt `secrets`.

pkg/console/subresource/configmap/configmap.go
```diff
@@ -51,16 +49,26 @@ func DefaultConfigMap(
...
 		APIServerURL(getApiUrl(infrastructureConfig)).
+		CustomLogoFile(defaultCustomLogoFile).
+		CustomProductName(defaultCustomProductName).
 		ConfigYAML()
 
 	extractedManagedConfig := extractYAML(managedConfig)
 
+	customProductName := operatorConfig.Spec.Customization.CustomProductName
+	customLogoFile := operatorConfig.Spec.Customization.CustomLogoFile.Key
+	if customLogoFile != "" {
```
benjaminapetersen: Push this little prefixer of `/var/logo` up into the `Builder. CustomLogoFile()` func?  We should be able to stay tidy here now.

pkg/console/subresource/configmap/configmap.go
```diff
@@ -16,12 +16,10 @@ import (
...
 
 const (
-	consoleConfigYamlFile = "console-config.yaml"
-)
-
-// overridden by console config
-const (
-	defaultLogoutURL = ""
+	consoleConfigYamlFile    = "console-config.yaml"
+	defaultLogoutURL         = ""
```
benjaminapetersen: Do we need these new defaults now?

pkg/console/subresource/configmap/configmap.go
zherman0: excellent idea

pkg/console/subresource/configmap/configmap.go
zherman0: We do not.

pkg/console/subresource/configmap/configmap.go
```diff
@@ -55,12 +51,17 @@ func DefaultConfigMap(
 
 	extractedManagedConfig := extractYAML(managedConfig)
 
+	customProductName := operatorConfig.Spec.Customization.CustomProductName
```
benjaminapetersen: Lets skip the variables now:

```golang
Builder.
  CustomProductName(operatorConfig.Spec.Customization.CustomProductName) .
  CustomLogoFile(operatorConfig.Spec.Customization.CustomLogoFile.Key)
```
seems plenty clear.

benjaminapetersen:
Users will set the CustomLogoFile on the operator config, if it doesn't work, they will prob `oc get` it. So having a `.Ststus.Conditions[CustomLogoInvalid]` with an error message is prob good UX.

zherman0:
> Users will set the CustomLogoFile on the operator config, if it doesn't work, they will prob `oc get` it. So having a `.Ststus.Conditions[CustomLogoInvalid]` with an error message is prob good UX. That's assuming a bad config could cause the console pods to crash loop due to a file read air.

So here are the errors I have been able to generate:
1) User sets the operator config but does not create the configmap
Result: Console Pod stay s in `ContainerCreating` and Event message on pod says `FailedMount`
There is no obvious error in the Console deployment status
2) User sets the operator config and makes a config map but the key file is not there or invalid
Result: Console Pod has `CrashLoopBackoff`, nothing in Events but the console pod log file shows:
`cmd/main: could not read logo file: stat /var/logo/small-tractor.jpg: no such file or directory`
There is no obvious error in the Console deployment status

zherman0:
@benjaminapetersen - I added logic in the operator.go function `CheckCustomLogoImageStatus` that will see if the file exists and if it does not I set a Status Condition.  It does not work correctly but let me know if this is a reasonable way to inform user why the console pods might be crashing if there is a logo set.

enj:
Some comments about your API:

```yaml
...
spec:
  customization:
    ...
    customLogoFile:        # this really should be customLogo
      name:  myconfig      # validate using the logic stated below
      key: mypic.jpg       # there is no value is allowing a user to pick the key
 ```

```
must be no more than 253 characters

must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character
regex used for validation is:
[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*
```

File: pkg/console/operator/operator.go [hunk 34]
```diff
@@ -109,7 +112,8 @@ func NewConsoleOperator(
 		oauthClient:   oauthv1Client,
 		versionGetter: versionGetter,
 		// recorder
-		recorder: recorder,
+		recorder:       recorder,
+		resourceSyncer: resourceSyncer,
```
enj: I would expect a config informer name addition below.

pkg/console/operator/operator.go
```diff
@@ -251,3 +260,21 @@ func (c *consoleOperator) removeConsole(cr *operatorsv1.Console) error {
...
+func (c *consoleOperator) handleConfigMapSync(operatorConfig *operatorsv1.Console) error {
+	// Get Operator Config here
+	logoName := operatorConfig.Spec.Customization.CustomLogoFile.Name
+	// add syncing for the custom logo config map
+	if logoName != "" {
+		err := c.resourceSyncer.SyncConfigMap(
+			resourcesynccontroller.ResourceLocation{Namespace: api.OpenShiftConsoleNamespace, Name: logoName},
+			resourcesynccontroller.ResourceLocation{Namespace: api.OpenShiftConfigNamespace, Name: logoName},
+		)
+		if err != nil {
```
enj: Just `return c.resourceSyncer.SyncConfigMap(...`

pkg/console/operator/operator.go
```diff
@@ -251,3 +260,21 @@ func (c *consoleOperator) removeConsole(cr *operatorsv1.Console) error {
 
 	return utilerrors.FilterOut(utilerrors.NewAggregate(errs), errors.IsNotFound)
 }
+
+func (c *consoleOperator) handleConfigMapSync(operatorConfig *operatorsv1.Console) error {
+	// Get Operator Config here
+	logoName := operatorConfig.Spec.Customization.CustomLogoFile.Name
+	// add syncing for the custom logo config map
+	if logoName != "" {
```
enj: If this is empty, you need to set the destination to source to `""/""`

pkg/console/operator/operator.go
```diff
@@ -251,3 +260,21 @@ func (c *consoleOperator) removeConsole(cr *operatorsv1.Console) error {
...
+		err := c.resourceSyncer.SyncConfigMap(
+			resourcesynccontroller.ResourceLocation{Namespace: api.OpenShiftConsoleNamespace, Name: logoName},
```
enj: The destination must be static.

pkg/console/operator/operator.go
zherman0: I do not have another config informer.  The informer is part of the resourceSyncer.

pkg/console/operator/operator.go
zherman0: I am skipping the sync code completely if there is no logoName set so I don't know what good it would be to set the source to "/".  Currently, the code is working as intended.

pkg/console/operator/operator.go
zherman0: I wish I could make the destination name static but that value is set by the user.  The SyncConfigMap does not have an issue with me passing a dynamic name into the ResourceLocations.

pkg/console/operator/operator.go
```diff
@@ -251,3 +260,36 @@ func (c *consoleOperator) removeConsole(cr *operatorsv1.Console) error {
 
 	return utilerrors.FilterOut(utilerrors.NewAggregate(errs), errors.IsNotFound)
 }
+
```
benjaminapetersen: We have a number of `ConfigMaps`, this does not handle syncing them all.

I would probably name this `SyncCustomLogoConfigMap()` . Take a look in `sync_v400()`, see where we call `SyncConfigMap()` and `SyncServiceCAConfigMap()`.  Generally, if you can follow that pattern for where this func lives, that will prob keep us maintainable / less messy.

benjaminapetersen: (This means you would call it as `func SyncCustomLogoConfigMap(c *consoleOperator, operatorConfig *operatorsv1.Console)`), but I do think that `sync_v400` ought to have been a method on the `operator` to match what you have.

pkg/console/operator/operator.go
```diff
@@ -251,3 +260,36 @@ func (c *consoleOperator) removeConsole(cr *operatorsv1.Console) error {
...
+}
+
+func (c *consoleOperator) CheckCustomLogoImageStatus(operatorConfig *operatorsv1.Console) {
+	logo := operatorConfig.Spec.Customization.CustomLogoFile
+	if logo.Name != "" && logo.Key != "" {
+		//Check that configmap for custom logo exists
+		logoConfigMap, err := c.configMapClient.ConfigMaps(api.OpenShiftConfigNamespace).Get(logo.Name, metav1.GetOptions{})
+		if err != nil {
+			klog.Errorf("customLogo configmap not valid, %v", err)
```
benjaminapetersen: Do you ever see these `klog` calls?

pkg/console/operator/operator.go
```diff
@@ -201,6 +205,11 @@ func (c *consoleOperator) handleSync(operatorConfig *operatorsv1.Console, consol
 		return fmt.Errorf("console is in an unknown state: %v", operatorConfigCopy.Spec.ManagementState)
 	}
 
+	sErr := c.handleConfigMapSync(operatorConfig)
```
benjaminapetersen: Does this short circuit the loop here?  I think a couple issues could arise at this point.

pkg/console/operator/operator.go
benjaminapetersen: Meaning, should this func return an error that bubbles up? Or should the errors be exclusively in logs.

zherman0: Yes, I can see these statements in our log.

zherman0: I would prefer these stay in the log and the condition shows on the status of the operator config

pkg/console/operator/operator.go
zherman0: Why are you suggestion to pass `sync *resourceSyncer` into this?  I already have access to it through the operator.

pkg/console/operator/operator.go
zherman0: @enj - Can you clarify this further? We do have the config informer set just as it is in the cluster auth op.

pkg/console/operator/operator.go
benjaminapetersen: `""/""` meant "namespace / name"

pkg/console/operator/operator.go
benjaminapetersen: Try `oc apply -f` on this `operator config`:

```yaml
apiVersion: operator.openshift.io/v1
kind: Console
metadata:
  name: cluster
spec:
  customization:
    customProductName: Muh Clusta
    customLogoFile:
      name: console-config  # uh oh!  We gonna blow up....
      key: fake-logo.png
```

benjaminapetersen:
So I pulled this branch, built and deployed the operator. Then I did the following:

```bash 
# create a configmap with a logo
oc create configmap my-logo-file --from-file ~/Desktop/fake-logo.png -n openshift-config
# update the operator to use it 
oc apply -f console.operator.with.my-logo-file.yaml
```
My operator config file looks like this:

```yaml
apiVersion: operator.openshift.io/v1
kind: Console
metadata:
  name: cluster
spec:
  customization:
    customProductName: My Super Fancy Cluster
    customLogoFile:
      name: my-logo-file
      key: fake-logo.png
```

- [ ] I don't actually see a `customLogo` or a `customProductName` in the console, it still shows `Red Hat OpenShift Online`. 
- [ ] The console operator pod is hot looping.  It isn't looping fast, but it never settles into an idle state.  The sync loop should settle out after a couple spins, every time. 
- [ ] The operator status remains in a `Progressing` state (oc get console.operator.openshift.io -o yaml):
```
    - lastTransitionTime: 2019-06-18T18:13:04Z
      message: Changes made during sync updates, additional sync expected.
      reason: SyncLoopProgressing
      status: "True"
      type: Progressing
```
- [ ] Every time I change update the configmap, I get additional configs in the `openshift-console` namespace:
```
oc get configmap -n openshift-console
NAME            DATA   AGE
my-logo-file    0      11m
my-logo-file2   0      11m
my-logo-file3   0      11m
my-logo-file4   0      11s
my-logo-file5   0      7s
my-logo-file6   0      2s
service-ca      1      27h
# uh oh... console-config got clobbered & isn't coming back?
```
I should only have 1 configmap, just the most recent.
- [ ] If I use an existing name `console-config` or `service-ca`, it seems I permanently eliminate them. See:
```
oc get configmap -n openshift-console
NAME            DATA   AGE
my-logo-file    0      11m
my-logo-file2   0      11m
my-logo-file3   0      11m
my-logo-file4   0      64s
my-logo-file5   0      60s
my-logo-file6   0      55s
# service-ca is gone
# console-config is gone 
```
- [ ] The `operator` is in a permanent progressing state.  See `oc get console.operator.openshift.io -o yaml` or ` oc get clusteroperator console -o yaml`
- [ ] I can't undo this state, If I try to apply the following clean operator config:
```yaml
apiVersion: operator.openshift.io/v1
kind: Console
metadata:
  name: cluster
spec:
  managementState: Managed
```
- [ ] The operator remains in a hot loop (see `oc logs -f console-operator-55bffddbdf-bkswr`) and is stuck in `SyncLoopProgressing` (see `oc get console.operator.openshift.io -o yaml`)

pkg/console/operator/operator.go
benjaminapetersen: I see your latest change, that will work fine.

pkg/console/operator/operator.go
```diff
@@ -251,3 +260,40 @@ func (c *consoleOperator) removeConsole(cr *operatorsv1.Console) error {
...
+
+func (c *consoleOperator) CheckCustomLogoImageStatus(operatorConfig *operatorsv1.Console) {
+	logo := operatorConfig.Spec.Customization.CustomLogoFile
+	if logo.Name != "" && logo.Key != "" {
+		//Check that configmap for custom logo exists
+		logoConfigMap, err := c.configMapClient.ConfigMaps(api.OpenShiftConfigNamespace).Get(logo.Name, metav1.GetOptions{})
+		if err != nil {
+			klog.Errorf("customLogo configmap not valid, %v", err)
+			// Set Operator Status to CustomLogoInvalid
+			c.SetStatusCondition(operatorConfig, reasonCustomLogoInvalid, operatorsv1.ConditionFalse, reasonCustomLogoInvalid, "custom logo config map does not exist or has error")
```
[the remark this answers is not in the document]
benjaminapetersen: Aborted before a `sync` of status.

pkg/console/operator/operator.go
benjaminapetersen: This will probably clarify:
https://github.com/openshift/cluster-authentication-operator/blob/master/pkg/operator2/operator.go#L230

benjaminapetersen:
/hold

Until bugs addressed.

pkg/console/operator/operator.go
benjaminapetersen: Look around though, I believe every function in the project that makes an API call returns an error which bubbles up.

pkg/console/operator/operator.go
benjaminapetersen: Further this line:
```
operator.WithInformer(configMapInformer, operator.FilterByNames(api.OpenShiftConsoleConfigMapName, api.ServiceCAConfigMapName)),

```
Shows that right now in the`openshift-console` namespace, the `informers` only fire on changes to `console-config` & `service-ca`.  If you sync in `my-logo`, `my-logo2`, etc, these will not trigger the informer.  In addition, if `my-logo` is already sync'd into the `openshift-console` namespace, but the user changes the value of the logo within the configmap, this will also NOT be noticed.

It's possible you are seeing sync because you are triggering the loop in other ways.  By the time a typical user is getting to the point of configuring a custom logo, the system will be very idle.  There won't be other events to piggy-back on.  If you happen to be testing with some bash scripts, I'd be cautious about what they are doing that might be triggering the loop in unexpected ways.

pkg/console/operator/operator.go
benjaminapetersen: `Name` is the configmap name, which we should sync into `openshift-console` as `custom-logo`.
`Key` is the file name key.
The `<value>` of the key is the actual image.  There are a number of possible issues that could arise here, but I think we can ignore them for the most part. It could be `null`, or the `base64` encoded string may not be an image at all or a different image.  Lets not chase that rabbit unless we find reason to do so.

pkg/console/operator/operator.go
zherman0: I am glad you changed your mind on having this dynamic.  I will set the destination name to a set name and that will prevent existing configmaps from being clobbered.

pkg/console/operator/operator.go
zherman0: Still not following here, if the logoName is empty (the value has not been set on the operator config), then I would end up with `"namespace /  "`, plus this code would not get executed.

pkg/console/operator/operator.go
zherman0: So, we are assuming a "non-happy" path where the user changes the configmap directly in the `openshift-console` namespace? We should only expect a change when the user changes the configmap in `openshift-config` but I agree we should cover this just in case.  
As a side question, if the user can manipulate the `openshift-console` namespace, why did we bother putting the custom logo config map in the `openshift-config` in the first place?

zherman0: done

pkg/console/operator/operator.go
zherman0: done

benjaminapetersen: >  I am glad you changed your mind on having this dynamic. 

I dont know what this means?

pkg/console/operator/operator.go
benjaminapetersen: No, we are not saying the user changes a configmap directly in `openshift-console`.  I think you are misunderstanding the design here.  You have 2 sync loops running with the `syncer`.  

When the `resource syncer` finally decides "yes, this is a logo, copy it over"... our main sync needs to know that the new `custom-logo` configmap even exists & watches is so that it updates (and you cannot trust an accidental update because the sync loop ran for another reason).

benjaminapetersen: Make sure you understand how the sync loops communicate.  It is a very small overlap (precisely to ensure our main loop is not triggered too much).

pkg/console/operator/operator.go
benjaminapetersen: Ensure it deletes?  I was able to easily get about a dozen new configmaps in the`openshift-console` namespace.  That shouldn't happen.

pkg/console/operator/operator.go
benjaminapetersen: We haven't changed our minds.  The user SHOULD be able to do:

```yaml
    customLogoFile:
      name: console-config  # or service-ca or whatever they want to call it, we don't care...
      key: fake-logo.png
```
And it SHOULD NOT destroy `console-config`.  This should be copied into a `configmap` named `custom-logo` in `openshift-console` with no risk of destroying any console config.

benjaminapetersen:
Closing this one in favor of #245 
#245 includes this PR + some additional commits on top to help move us to completion since we have so much vacation time happening right now.  Original authorship of commits will be retained!

PR #222, opened 2019-05-02
Check if OAuth is already updated when in Removed state to prevent hotloop

jhadvig:
To prevent hot-looping in `Removed` state, we need to add condition check (added the check to the `Unmanaged` state as well)

Tested the change locally and the hot-looping is gone, also e2e are green

Bug: https://jira.coreos.com/browse/CONSOLE-1430?filter=-1

/assign @spadgett 

@benjaminapetersen fyi

### DIFF
--- a/pkg/console/operator/operator.go
+++ b/pkg/console/operator/operator.go
@@ -249,8 +249,10 @@ func (c *consoleOperator) deleteAllResources(cr *operatorsv1.Console) error {
 	// existingOAuthClient is not a delete, it is a deregister/neutralize
 	existingOAuthClient, getAuthErr := c.oauthClient.OAuthClients().Get(oauthclient.Stub().Name, metav1.GetOptions{})
 	errs = append(errs, getAuthErr)
-	_, updateAuthErr := c.oauthClient.OAuthClients().Update(oauthclient.DeRegisterConsoleFromOAuthClient(existingOAuthClient))
-	errs = append(errs, updateAuthErr)
+	if len(existingOAuthClient.RedirectURIs) != 0 {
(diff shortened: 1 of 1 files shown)


spadgett:
I think the reason we hot-loop is that we update the OAuth client on every loop, which triggers a new loop. A cleaner fix is probably to only update the OAuth client when we need to.

The drawback to this fix is if there are any errors deleting any of the resources it will never try again.

pkg/console/operator/operator.go
```diff
@@ -246,11 +252,13 @@ func (c *consoleOperator) deleteAllResources(cr *operatorsv1.Console) error {
 	errs = append(errs, c.configMapClient.ConfigMaps(api.TargetNamespace).Delete(configmap.ServiceCAStub().Name, &metav1.DeleteOptions{}))
 	// secret
 	errs = append(errs, c.secretsClient.Secrets(api.TargetNamespace).Delete(secret.Stub().Name, &metav1.DeleteOptions{}))
-	// existingOAuthClient is not a delete, it is a deregister/neutralize
-	existingOAuthClient, getAuthErr := c.oauthClient.OAuthClients().Get(oauthclient.Stub().Name, metav1.GetOptions{})
-	errs = append(errs, getAuthErr)
-	_, updateAuthErr := c.oauthClient.OAuthClients().Update(oauthclient.DeRegisterConsoleFromOAuthClient(existingOAuthClient))
-	errs = append(errs, updateAuthErr)
+	if updateOAuth {
```
spadgett: I wouldn't pass in a flag. I would get the OAuth client and check that is matches the expected value. If not, update.

jhadvig:
@spadgett updated. PTAL

PR #223, opened 2019-05-03
Update downloads deployment node selector and tolerations

spadgett:
Don't tolerate all taints, only `node-role.kubernetes.io/master`.

/cc @sjenning @benjaminapetersen @wking


sjenning:
/hold
fyi @ravisantoshgudimetla will open a PR that will supersede this.  Turns out we need to do something like this https://github.com/openshift/cloud-credential-operator/pull/64

spadgett:
closing in favor of #224

PR #229, opened 2019-05-16
Add console extension manifests

benjaminapetersen:
Adds 3 new CRDs for console extensions:
- ConsoleLink
- ConsoleCLIDownload
- ConsoleNotification (Banners)

Related to Console PRs:
- ConsoleLink https://github.com/openshift/console/pull/1360
- ConsoleCLIDownload https://github.com/openshift/console/pull/1441
- ConsoleNotification (Banners) https://github.com/openshift/console/pull/1542

### DIFF
--- a/manifests/00-crd-extension-console-cli-download.yaml
+++ b/manifests/00-crd-extension-console-cli-download.yaml
@@ -6,13 +6,12 @@ metadata:
     displayName: ConsoleCLIDownload
     description: Extension for configuring openshift web console command line interface (CLI) downloads.
 spec:
+  scope: Cluster
   group: console.openshift.io
-  version: v1
   versions:
   - name: v1
     served: true
     storage: true
-  scope: Cluster
   names:
     plural: consoleclidownloads
     singular: consoleclidownload
@@ -39,7 +38,7 @@ spec:
           required:
           - displayName
           - description
-          - link
+          - links
           properties:
             displayName:
               type: string
@@ -47,16 +46,20 @@ spec:
             description:
               type: string
               description: Description of the CLI download (can include markdown)
-            link:
-              type: object
-              description: Object that holds CLI download link details
-              required:
-              - href
-              properties:
-                text:
-                  type: string
-                  description: Text of the link
-                href:
-                  type: string
-                  description: Absolute secure URL for the link (must use https)
-                  pattern: '^https://([\w-]+.)+[\w-]+(/[\w- ./?%&=])?$'
+            links:
+              type: array
+              description: Objects that hold CLI download link details
+              nullable: true
+              items:
+                type: object
+                required:
+                - href
+                properties:
+                  text:
+                    type: string
+                    description: Text of the link
+                  href:
+                    type: string
+                    description: Absolute secure URL for the link (must use https)
+                    pattern: '^https://([\w-]+.)+[\w-]+(/[\w- ./?%&=])?$'
+
--- a/manifests/00-crd-extension-console-cli.yaml
+++ b/manifests/00-crd-extension-console-cli.yaml
@@ -6,10 +6,6 @@ metadata:
     displayName: ConsoleCLIDownload
     description: Extension for configuring openshift web console command line interface (CLI) downloads.
 spec:
-  # TODO:
-  # this relates to where the API resources live in openshift/api
-  # likely under /config/console, which might make it something like:
-  # extensions.console.config.openshift.io?
   group: console.openshift.io
   version: v1
   versions:
--- a/manifests/00-crd-extension-console-cli.yaml
+++ b/manifests/00-crd-extension-console-cli.yaml
@@ -1,0 +1,66 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: consoleclidownloads.console.openshift.io
+  annotations:
+    displayName: ConsoleCLIDownload
+    description: Extension for configuring openshift web console command line interface (CLI) downloads.
+spec:
+  # TODO:
+  # this relates to where the API resources live in openshift/api
+  # likely under /config/console, which might make it something like:
+  # extensions.console.config.openshift.io?
+  group: console.openshift.io
+  version: v1
+  versions:
+  - name: v1
+    served: true
+    storage: true
+  scope: Cluster
+  names:
+    plural: consoleclidownloads
+    singular: consoleclidownload
+    kind: ConsoleCLIDownload
+    listKind: ConsoleCLIDownloadList
+  additionalPrinterColumns:
+  - name: Display name
+    type: string
+    JSONPath: .spec.displayName
+  - name: Age
+    type: string
+    JSONPath: .metadata.creationTimestamp
+  - name: Description
+    type: string
+    JSONPath: .spec.description
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      properties:
+        spec:
+          type: object
+          description: Represents console CLI download spec
+          required:
+          - displayName
+          - description
+          - link
+          properties:
+            displayName:
+              type: string
+              description: Display name of the CLI download
+            description:
+              type: string
+              description: Description of the CLI download (can include markdown)
+            link:
+              type: object
+              description: Object that holds CLI download link details
+              required:
+              - href
+              properties:
+                text:
+                  type: string
+                  description: Text of the link
+                href:
+                  type: string
+                  description: Absolute secure URL for the link (must use https)
+                  pattern: '^https://([\w-]+.)+[\w-]+(/[\w- ./?%&=])?$'
--- a/manifests/00-crd-extension-console-link.yaml
+++ b/manifests/00-crd-extension-console-link.yaml
@@ -6,13 +6,12 @@ metadata:
     displayName: ConsoleLinks
     description: Extension for customizing OpenShift web console links
 spec:
+  scope: Cluster
   group: console.openshift.io
-  version: v1
   versions:
   - name: v1
     served: true
     storage: true
-  scope: Cluster
   names:
     plural: consolelinks
     singular: consolelink
@@ -42,16 +41,16 @@ spec:
           required:
           - text
           - href
-          - menu
+          - location
           properties:
             text:
               type: string
               description: Text of the link
             href:
               type: string
-              description: URL for the link
-              pattern: '^http(s)?://([\w-]+.)+[\w-]+(/[\w- ./?%&=])?$'
-            menu:
+              description: Absolute secure URL for the link (must use https)
+              pattern: '^https://([\w-]+.)+[\w-]+(/[\w- ./?%&=])?$'
+            location:
               type: string
-              description: Determines which dropdown menu the link is added to (help, user)
-              pattern: '\bhelp\b|\buser\b'
+              description: Determines which dropdown menu the link is added to (HelpMenu, UserMenu)
+              pattern: '\bHelpMenu\b|\bUserMenu\b'
--- a/manifests/00-crd-extension-console-link.yaml
+++ b/manifests/00-crd-extension-console-link.yaml
@@ -6,10 +6,6 @@ metadata:
     displayName: ConsoleLinks
     description: Extension for customizing OpenShift web console links
 spec:
-  # TODO:
-  # this relates to where the API resources live in openshift/api
-  # likely under /config/console, which might make it something like:
-  # extensions.console.config.openshift.io?
   group: console.openshift.io
   version: v1
   versions:
--- a/manifests/00-crd-extension-console-link.yaml
+++ b/manifests/00-crd-extension-console-link.yaml
@@ -1,0 +1,61 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: consolelinks.console.openshift.io
+  annotations:
+    displayName: ConsoleLinks
+    description: Extension for customizing OpenShift web console links
+spec:
+  # TODO:
+  # this relates to where the API resources live in openshift/api
+  # likely under /config/console, which might make it something like:
+  # extensions.console.config.openshift.io?
+  group: console.openshift.io
+  version: v1
+  versions:
+  - name: v1
+    served: true
+    storage: true
+  scope: Cluster
+  names:
+    plural: consolelinks
+    singular: consolelink
+    kind: ConsoleLink
+    listKind: ConsoleLinkList
+  additionalPrinterColumns:
+  - name: Text
+    type: string
+    JSONPath: .spec.text
+  - name: URL
+    type: string
+    JSONPath: .spec.href
+  - name: Menu
+    type: string
+    JSONPath: .spec.menu
+  - name: Age
+    type: date
+    JSONPath: .metadata.creationTimestamp
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      properties:
+        spec:
+          type: object
+          description: Represents console link customizations spec
+          required:
+          - text
+          - href
+          - menu
+          properties:
+            text:
+              type: string
+              description: Text of the link
+            href:
+              type: string
+              description: URL for the link
+              pattern: '^http(s)?://([\w-]+.)+[\w-]+(/[\w- ./?%&=])?$'
+            menu:
+              type: string
+              description: Determines which dropdown menu the link is added to (help, user)
+              pattern: '\bhelp\b|\buser\b'
--- a/manifests/00-crd-extension-console-notification-banner.yaml
+++ b/manifests/00-crd-extension-console-notification-banner.yaml
@@ -1,0 +1,65 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: consolenotificationbanners.console.openshift.io
+  annotations:
+    displayName: ConsoleNotification
+    description: Extension for configuring openshift web console notifications.
+spec:
+  group: console.openshift.io
+  version: v1
+  versions:
+  - name: v1
+    served: true
+    storage: true
+  scope: Cluster
+  names:
+    plural: consolenotifications
+    singular: consolenotification
+    kind: ConsoleNotification
+    listKind: ConsoleNotificationList
+  additionalPrinterColumns:
+  - name: Text
+    type: string
+    JSONPath: .spec.text
+  - name: Age
+    type: date
+    JSONPath: .metadata.creationTimestamp
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      properties:
+        spec:
+          type: object
+          description: Represents console notification spec
+          required:
+          - text
+          - position
+          properties:
+            text:
+              type: string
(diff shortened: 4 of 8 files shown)


benjaminapetersen:
/hold

WIP. Need to make sure we know where these will live in the API, if the group is correct, and if any will be converted into top-level config/operator config items instead of custom resources.

benjaminapetersen:
fyi @rhamilto @spadgett

manifests/00-crd-extension-console-notification-banner.yaml
```diff
@@ -0,0 +1,72 @@
...
+              properties:
+                text:
+                  type: string
+                  description: Text of the link
+                href:
+                  type: string
+                  description: Absolute URL for the link
+                  pattern: '^http(s)?://([\w-]+.)+[\w-]+(/[\w- ./?%&=])?$'
+            backgroundColor:
+              type: string
```
benjaminapetersen: Should this clarify if it needs to be a supported CSS color name or a hex value?

manifests/00-crd-extension-console-notification-banner.yaml
```diff
@@ -0,0 +1,72 @@
...
+              description: The background color for the notification
+            color:
+              type: string
```
benjaminapetersen: Should this clarify if it needs to be a supported CSS color name or a hex value?

spadgett:
@jwforres fyi

manifests/00-crd-extension-console-link.yaml
```diff
@@ -0,0 +1,61 @@
...
+  names:
+    plural: consolelinks
+    singular: consolelink
+    kind: ConsoleLink
+    listKind: ConsoleLinkList
+  additionalPrinterColumns:
+  - name: Text
+    type: string
+    JSONPath: .spec.text
+  - name: URL
```
benjaminapetersen: Not sure if we want to use name `URL` then reference `.spec.href`.  Can we choose `href` or `url`?  Seems like this could be confusing.

`Age` I get because when you `oc get <resource>` you get an `AGE` column that is based on the `creationTimestamp`, but I think that makes more sense as it is a computed value.  

We aren't computing anything, we are just swapping the label.

manifests/00-crd-extension-console-notification-banner.yaml
```diff
@@ -0,0 +1,72 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  # TODO: consolenotificationbanners.console.openshift.io
+  # will we ever have consolenotifications that act like
+  # traditional notifications?
+  name: consolenotifications.console.openshift.io
+  annotations:
+    displayName: ConsoleNotification
```
benjaminapetersen: Really think this should be `ConsoleNotificationBanner`.  This isn't how `Notification` is typically understood.

benjaminapetersen: https://github.com/openshift/api/pull/318/files#diff-d02c06759975ba6309671b3248104d7bR10

`ConsoleNotificationBanner` vs `ConsoleNotification`.

manifests/00-crd-extension-console-notification-banner.yaml
benjaminapetersen: Also, we should make it optional & defaulted.

manifests/00-crd-extension-console-notification-banner.yaml
benjaminapetersen: Also, we should make it optional & defaulted.

manifests/00-crd-extension-console-notification-banner.yaml
benjaminapetersen: https://github.com/openshift/console/pull/1542#discussion_r285668581

manifests/00-crd-extension-console-notification-banner.yaml
benjaminapetersen: https://github.com/openshift/console/pull/1542#discussion_r285668581

benjaminapetersen:
```
possibly due to local cluster configuration
May 21 19:00:40.695 E clusterversion/version changed Failing to True: UpdatePayloadResourceInvalid: Could not update customresourcedefinition "consolenotificationbanners.console.openshift.io" (221 of 353): the object is invalid, possibly due to local cluster configuration
May 21 19:09:10.695 E clusterversion/version changed Failing to True: UpdatePayloadResourceInvalid: Could not update customresourcedefinition "consolenotificationbanners.console.openshift.io" (221 of 353): the object is invalid, possibly due to local cluster configuration
May 21 19:19:25.696 E clusterversion/version changed Failing to True: UpdatePayloadResourceInvalid: Could not update customresourcedefinition "consolenotificationbanners.console.openshift.io" (221 of 353): the object is invalid, possibly due to local cluster configuration
May 21 19:29:25.693 E clusterversion/version changed Failing to True: UpdatePayloadResourceInvalid: Could not update customresourcedefinition "consolenotificationbanners.console.openshift.io" (221 of 353): the object is invalid, possibly due to local cluster configuration
May 21 19:39:40.694 E clusterversion/version changed Failing to True: UpdatePayloadResourceInvalid: Could not update customresourcedefinition "consolenotificationbanners.console.openshift.io" (221 of 353): the object is invalid, possibly due to local cluster configuration
May 21 19:49:54.657 E clusterversion/version changed Failing to True: UpdatePayloadResourceInvalid: Could not update customresourcedefinition "consolenotificationbanners.console.openshift.io" (221 of 353): the object is invalid, possibly due to local cluster configuration
```
trying to diagnose what is using these.

manifests/00-crd-extension-console-notification-banner.yaml
rhamilto: I suggest `ConsoleBanner`.

manifests/00-crd-extension-console-notification-banner.yaml
```diff
@@ -0,0 +1,65 @@
...
+                text:
+                  type: string
+                  description: Text of the link
+                href:
+                  type: string
+                  description: Absolute URL for the link
+                  pattern: '^http(s)?://([\w-]+.)+[\w-]+(/[\w- ./?%&=])?$'
+            backgroundColor:
+              type: string
+              description: The background color for the notification
```
rhamilto:
```suggestion
              description: The background color for the banner as CSS data type color
```

manifests/00-crd-extension-console-notification-banner.yaml
```diff
@@ -0,0 +1,65 @@
...
+            color:
+              type: string
+              description: The color of the text for the notification
```
rhamilto:
```suggestion
              description: The color of the text for the banner as CSS data type color
```

manifests/00-crd-extension-console-notification-banner.yaml
benjaminapetersen: Too generic?  

Though this one is still under consideration to get dropped & become `operator` config.

manifests/00-crd-extension-console-notification-banner.yaml
```diff
@@ -0,0 +1,65 @@
...
+              description: Text of the notification
+            position:
+              type: string
+              description: The location of the notification (top, bottom, topBottom)
+              pattern: '\btop\b|\bbottom\b|\btopBottom\b'
+            link:
+              type: object
+              description: Object that holds notification link details
+              required:
+              - href
```
benjaminapetersen: `- text` ought to be required as well?  can't easily have a link w/o text.

rhamilto: There is a default text (More info) in the console.

manifests/00-crd-extension-console-cli.yaml
```diff
@@ -0,0 +1,62 @@
...
+          - description
+          - link
+          properties:
+            displayName:
+              type: string
+              description: Display name of the CLI download
+            description:
+              type: string
+              description: Description of the CLI download (can include markdown)
+            link:
```
benjaminapetersen: Possibly needs to be `links [ ]`

benjaminapetersen: see https://github.com/openshift/console/pull/1441/files#r276345530

manifests/00-crd-extension-console-cli.yaml
```diff
@@ -0,0 +1,62 @@
...
+              type: object
+              description: Object that holds CLI download link details
+              required:
+              - href
```
benjaminapetersen: Should I presume `text` is defaulted here as well?
Looks like it is `Download {displayName} || link.text`

benjaminapetersen: Not sure if we need to doc that in the API resource.

benjaminapetersen: Per [original PR description](https://github.com/openshift/console/pull/1441):

image

The `Download {displayName}` is preferred, not sure if we want this to be customizable?

benjaminapetersen: I'm assuming if an operator wants to create this resource to add an item to the list, that operator is also creating a deployment with a small file server w/a route & would want to use that route to link here?

manifests/00-crd-extension-console-notification-banner.yaml
benjaminapetersen: Ok, trying to decide if we need to doc that on the api object or not.

manifests/00-crd-extension-console-cli.yaml
spadgett: It's not necessarily hosted in cluster, although it could be. I think we leave that up to the operator.

manifests/00-crd-extension-console-cli.yaml
```diff
@@ -0,0 +1,62 @@
...
+              properties:
+                text:
+                  type: string
+                  description: Text of the link
+                href:
+                  type: string
+                  description: Absolute secure URL for the link (must use https)
```
benjaminapetersen: Any reason this is the only one requiring HTTPS?

manifests/00-crd-extension-console-link.yaml
```diff
@@ -0,0 +1,60 @@
...
+          # TODO: ensure prop 'location' matches final decision in API discussion
+          - location
+          properties:
+            text:
+              type: string
+              description: Text of the link
+            href:
+              type: string
+              description: URL for the link
+              pattern: '^http(s)?://([\w-]+.)+[\w-]+(/[\w- ./?%&=])?$'
```
benjaminapetersen: HTTPS optional?

manifests/00-crd-extension-console-notification-banner.yaml
```diff
@@ -0,0 +1,66 @@
...
+              - text
+              properties:
+                text:
+                  type: string
+                  description: Text of the link
+                href:
+                  type: string
+                  description: Absolute URL for the link
+                  pattern: '^http(s)?://([\w-]+.)+[\w-]+(/[\w- ./?%&=])?$'
```
benjaminapetersen: HTTPS optional?

manifests/00-crd-extension-console-cli.yaml
spadgett: I'd suggest we make them all https. We can relax the restriction later if needed, but we can't ever make it more restrictive.

manifests/00-crd-extension-console-notification-banner.yaml
benjaminapetersen: `ConsoleNotification` seems to be the winner.

benjaminapetersen: ConsoleNotification.

manifests/00-crd-extension-console-cli.yaml
```diff
@@ -0,0 +1,62 @@
...
+              description: Description of the CLI download (can include markdown)
+            link:
+              type: object
+              description: Object that holds CLI download link details
+              required:
+              - href
+              properties:
+                text:
+                  type: string
+                  description: Text of the link
```
benjaminapetersen: `description: Text of the link.  If omitted, the text will default to "Download {displayName}".` ?
Not sure how helpful we want to make this.

manifests/04-rbac-rolebinding.yaml
```diff
@@ -65,3 +65,16 @@ subjects:
...
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: console-extensions
+roleRef:
+  kind: Role
+  name: console-extensions
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+- kind: Group
+  name: system:authenticated
```
enj: This is clearly the wrong way to implement this.

manifests/00-crd-extension-console-cli.yaml
benjaminapetersen: Any reason to further consider this?  Would there be a `link` for the binary + a `link` for docs, for example?  @rhamilto @spadgett

manifests/00-crd-extension-console-cli-download.yaml
```diff
@@ -0,0 +1,65 @@
...
+          - description
+          - links
+          properties:
+            displayName:
+              type: string
+              description: Display name of the CLI download
+            description:
+              type: string
+              description: Description of the CLI download (can include markdown)
+            links:
```
benjaminapetersen: Updated to a list of links rather than a single link.

benjaminapetersen: Will now look like:

```yaml
kind: ConsoleCLIDownload
metadata:
  creationTimestamp: '2019-06-03T18:09:15Z'
  generation: 1
  name: example
  resourceVersion: '90133'
  selfLink: /apis/console.openshift.io/v1/consoleclidownloads/example
  uid: b2d954aa-862a-11e9-8b2a-029c8d4a0f76
spec:
  description: Download this thing, its safe. Promise.
  displayName: Thing To Download
  links:
    - href: 'https://www.redhat.com'
      text: download A
    - href: 'https://www.redhatagain.com'
      text: download B

```

benjaminapetersen:
`level=fatal msg="failed to initialize the cluster: Could not update rolebinding \"console-extensions\" (243 of 378): timed out waiting for the condition"`

manifests/04-rbac-rolebinding-cluster.yaml
```diff
@@ -0,0 +1,12 @@
+kind: ClusterRoleBinding
```
benjaminapetersen: `ClusterRoleBinding`, not `RoleBinding`, should fix that test error.

benjaminapetersen:
/retest

flake details:
```
Jun 04 21:24:00.789 E clusteroperator/monitoring changed Degraded to True: Failed to rollout the stack. Error: running task Updating prometheus-adapter failed: reconciling PrometheusAdapter Deployment failed: updating deployment object failed: waiting for DeploymentRollout of prometheus-adapter: deployment prometheus-adapter is not ready. status: (replicas: 3, updated: 2, ready: 2, unavailable: 1)
Jun 04 21:25:58.536 E ns/openshift-monitoring pod/prometheus-adapter-7b8ccf5fd4-tdzv6 node/ip-10-0-139-114.ec2.internal container=prometheus-adapter container exited with code 2 (Error): 
```

manifests/04-rbac-rolebinding.yaml
spadgett: @bparees

bparees:
I spoke with @enj about the permissions aspect of this.  He indicated it is not a blocker, but I can understand his concerns about doing this which largely fall in two areas:

1) there's risk that someone extends this api resource in the future and inadvertently exposes something to all users that should not be exposed.  I definitely share this concern.  Even if it's not inadvertent you might find yourself wanting to extend this api in the future and being unable to because of the way it is shared.

2) there have been previous exploits that resulted from the platform allowing all users to do something we thought was safe that turned out not to be which has resulted in a general movement away from the use of "all users" type permissions.  I think there's less risk of that here (it is a read only resource with very limited surface area).

My view on the way forward is:

Proceed with this as is (Granting permission to all authenticated users) for now.  I don't want to see the UI team have to go implement a more complicated config controller flow right this minute, and @spadgett expressed some legitimate concerns about the quality of user experience we could provide if we did so (though i think it might be worth probing what options exist there a bit more, in terms of forcing browser refreshes or something, as suggested by @enj).  But file away a tech debt card to refactor this in the future(to use a config controller or other operator logic pattern) so we can eventually remove this permission.  Ideally if the additional 4.2 runway allows it, fix it in 4.2, but worst case, keep it on the roadmap, don't view this as the long term solution.

@enj @spadgett Does that seem reasonable?

enj:
> I spoke with @enj about the permissions aspect of this. He indicated it is not a blocker, but I can understand his concerns about doing this which largely fall in two areas:
> 
>     1. there's risk that someone extends this api resource in the future and inadvertently exposes something to all users that should not be exposed.  I definitely share this concern.  Even if it's not inadvertent you might find yourself wanting to extend this api in the future and being unable to because of the way it is shared.
> 
>     2. there have been previous exploits that resulted from the platform allowing all users to do something we thought was safe that turned out not to be which has resulted in a general movement away from the use of "all users" type permissions.

> I think there's less risk of that here (it is a read only resource with very limited surface area).

For context, the exploit resulted in full cluster compromise via _discovery_, i.e. an API that is by definition read only and "safe" (because it is just metadata about what APIs are available on the cluster).  It had nothing to do with the data itself and was a by product of the complexity of the underling wiring (and CRDs are very complex under the hood).  That complexity combined with a missing error check is what led to the CVE.

Do I think another similar CVE is likely?  No.

Would I like to assume it will happen and take precautions?  Yes.
 
> My view on the way forward is:
> 
> Proceed with this as is (Granting permission to all authenticated users) for now. I don't want to see the UI team have to go implement a more complicated config controller flow right this minute, and @spadgett expressed some legitimate concerns about the quality of user experience we could provide if we did so (though i think it might be worth probing what options exist there a bit more, in terms of forcing browser refreshes or something, as suggested by @enj). But file away a tech debt card to refactor this in the future(to use a config controller or other operator logic pattern) so we can eventually remove this permission. Ideally if the additional 4.2 runway allows it, fix it in 4.2, but worst case, keep it on the roadmap, don't view this as the long term solution.
> 
> @enj @spadgett Does that seem reasonable?

Sounds reasonable.  @spadgett @benjaminapetersen I am trusting you guys to make an honest attempt at an architecture that does not involve assigning RBAC to all authenticated users.

manifests/00-crd-extension-console-notification.yaml
```diff
@@ -0,0 +1,68 @@
...
+          - text
+          - location
+          properties:
+            text:
+              type: string
+              description: Text of the notification
+            location:
+              type: string
+              description: The location of the notification (BannerTop, BannerBottom, BannerTopBottom)
+              pattern: '\bBannerTop\b|\bBannerBottom\b|\bBannerTopBottom\b'
```
spadgett: If this is a substring match, so we'd need something like

```suggestion
              pattern: '^(BannerTop|BannerBottom|BannerTopBottom)$'
```

benjaminapetersen: Updating now.

PR #230, opened 2019-05-16
manifests: add resource requests to downloads pod

sjenning:
in service to https://github.com/openshift/origin/pull/22787

### DIFF
--- a/manifests/07-downloads-deployment.yaml
+++ b/manifests/07-downloads-deployment.yaml
@@ -38,6 +38,10 @@ spec:
           name: http
           protocol: TCP
         command: ["/bin/sh"]
+        resources:
+          requests:
+            cpu: 10m
+            memory: 50Mi
         args:
         - '-c'
         - |


manifests/07-downloads-deployment.yaml
```diff
@@ -38,6 +38,10 @@ spec:
           name: http
           protocol: TCP
         command: ["/bin/sh"]
+        resources:
+          requests:
+            cpu: 10m
+            memory: 50Mi
```
spadgett: This seems too low for what the pod actually uses :(

spadgett: Screen Shot 2019-05-16 at 4 39 34 PM

sjenning: much of that is fs cache.  i suspected that and did a `drop_caches` on the worker running the pod and it dropped to 30Mi.

PR #231, opened 2019-05-23
Bump to Go 1.12

benjaminapetersen:
- [x] Updates README to reference go 1.12
- [x] Runs go 1.12 gofmt against `/pkg`, `/cmd`. 
- [x] Updates Dockerfiles to use go1.12 
- [x] Update [openshift/release](https://github.com/openshift/release/pull/1653/files#diff-7a578435e18f1db0266de3b9be828b72R32) for `console-operator` to use `golang-1.12`.  PR: https://github.com/openshift/release/pull/3878
- [x] drops env vars no longer used (ex: `GOCACHE`)
- [x] drops flags no longer used (ex: `-shadow`)

I assume the PR to `openshift/release` must merge first, then this PR can merge to comply with the update, then we can roll from there.


spadgett:
Do we need to update the Dockefile(s) too?

jhadvig:
yes we should update both, centos and rhel one

benjaminapetersen:
Yup! Thx, updated.

spadgett:
@benjaminapetersen gofmt passed, but looks like we're passing a bad flag to govet

> flag provided but not defined: -shadow

benjaminapetersen:
Got it, in `hack/verify-govet.sh`.  Apparently `-shadow` was experimental and removed.  Looking a bit more to make sure it's fine to drop from this script.

benjaminapetersen:
(Should be fine since we did `-shadow=false` i would think)

hack/verify-govet.sh
```diff
@@ -22,7 +23,7 @@ function govet_blacklist_contains() {
 }
 
 for test_dir in $(os::util::list_go_src_dirs); do
-	if ! result="$(go tool vet -shadow=false -printfuncs=Info,Infof,Warning,Warningf "${test_dir}" 2>&1)"; then
+	if ! result="$(go vet -printfuncs=Info,Infof,Warning,Warningf "${test_dir}/..." 2>&1)"; then
```
benjaminapetersen: ./... is apparently the syntax for recursion.  No idea how this was working before, frankly.

benjaminapetersen:
Works for me, opened #232 to start that process.

PR #233, opened 2019-05-31
Bump deps

jhadvig:
Need to bump the deps because of https://github.com/openshift/api/pull/320

/assign @benjaminapetersen


benjaminapetersen:
/lgtm

With a note that you had to manually `glide update --strip-vendor`, even though `make update-deps` should have handled that.  Just keeping tabs on that for next time.

PR #234, opened 2019-06-03
Add providers with statuspage.io to operator config

jhadvig:
/assign @benjaminapetersen

### DIFF
--- a/pkg/console/subresource/configmap/configmap.go
+++ b/pkg/console/subresource/configmap/configmap.go
@@ -49,7 +49,8 @@ func DefaultConfigMap(operatorConfig *operatorv1.Console, consoleConfig *configv
 	// Build a default config that can be overwritten from other sources
 	host := rt.Spec.Host
 	apiServerURL := getApiUrl(infrastructureConfig)
-	defaultConfig := NewYamlConfig(host, defaultLogoutURL, DEFAULT_BRAND, DEFAULT_DOC_URL, apiServerURL)
+	defaultProviders := operatorv1.ConsoleProviders{}
+	defaultConfig := NewYamlConfig(host, defaultLogoutURL, DEFAULT_BRAND, DEFAULT_DOC_URL, apiServerURL, defaultProviders)
 	// Get console config from the openshift-config-managed namespace
 	extractedManagedConfig := extractYAML(managedConfig)
 
@@ -57,7 +58,8 @@ func DefaultConfigMap(operatorConfig *operatorv1.Console, consoleConfig *configv
 	logoutRedirect := consoleConfig.Spec.Authentication.LogoutRedirect
 	brand := operatorConfig.Spec.Customization.Brand
 	docURL := operatorConfig.Spec.Customization.DocumentationBaseURL
-	userDefinedConfig := NewYamlConfig(host, logoutRedirect, brand, docURL, apiServerURL)
+	providers := operatorConfig.Spec.Providers
+	userDefinedConfig := NewYamlConfig(host, logoutRedirect, brand, docURL, apiServerURL, providers)
 
 	unsupportedConfigOverride := operatorConfig.Spec.UnsupportedConfigOverrides.Raw
 
@@ -119,7 +121,7 @@ func Stub() *corev1.ConfigMap {
 	return configMap
 }
 
-func NewYamlConfig(host string, logoutRedirect string, brand operatorv1.Brand, docURL string, apiServerURL string) []byte {
+func NewYamlConfig(host string, logoutRedirect string, brand operatorv1.Brand, docURL string, apiServerURL string, providers operatorv1.ConsoleProviders) []byte {
 	conf := yaml.MapSlice{
 		{
 			Key: "kind", Value: "ConsoleConfig",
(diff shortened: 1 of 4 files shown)


pkg/console/subresource/configmap/configmap.go
```diff
@@ -155,6 +159,17 @@ func servingInfo() yaml.MapSlice {
 	}
 }
 
+func consoleProviders(providers operatorv1.ConsoleProviders) yaml.MapSlice {
+	if providers.Statuspage != nil {
```
benjaminapetersen: This should be good for now. I imagine eventually we will need to `slice := yaml.MapSlice{ }`, and then conditionally append items one-by-one, but not necessary until we have other providers.

jhadvig:
@benjaminapetersen  comments addressed. PTAL

test/e2e/providers_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/openshift/console-operator/pkg/testframework"
+)
+
+const (
+	statuspageIDField = "statuspageID"
+	providersField    = "providers"
+)
+
+func setupProvidersTestCase(t *testing.T) *testframework.Clientset {
+	client := testframework.MustNewClientset(t, nil)
```
benjaminapetersen: Mostly I'm mentioning this to ensure a `tearDown()` func puts the `config` resources back in a `vanilla` state, and that it `poll` the `operator` `config` for `Available:True, Progressing:False, Failing:False` to ensure we are ready to go on to the next test (avoid flakes).

@zherman0 was working on a `testframework` generic `poll`.

benjaminapetersen: I don't think we have to hold this PR for it, but we will have to go back and update them when the func is ready.

benjaminapetersen:
/hold
I believe this has to wait for 4.2

benjaminapetersen:
but otherwise lgtm

spadgett:
> /hold
> I believe this has to wait for 4.2

master is already 4.2 :)

benjaminapetersen:
/hold cancel

> > /hold
> > I believe this has to wait for 4.2
>
> master is already 4.2 :)

neat :D

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -316,13 +342,18 @@ func TestNewYamlConfig(t *testing.T) {
...
 				docURL:         DEFAULT_DOC_URL,
 				apiServerURL:   mockAPIServer,
+				providers: operatorv1.ConsoleProviders{
+					Statuspage: &operatorv1.StatuspageProvider{
+						PageID: mockStatuspageID,
+					},
+				},
 			},
-			want: finalCMDefaultOKD,
+			want: finalCMProviders,
```
benjaminapetersen: So we have 1 test here, called `TestNewYamlConfig`, updating the `want` from `finalCMDefaultOKD` to `finalCMProviders` is confusing, IMHO.

Our table tests should probably have several cases:

```golang
tests := []struct {
     name string
     args args
     want string
}{ 
     {
          name: "Test NewYamlConfig() with defaults",
          args: args{
              ...
              // no StatusPage.io to ensure that a null value doesn't break our funcs
          },
          want: finalDefaultOKD  // should we inline?  
     },{
          name: "Test NewYamlConfig() with Status.io provider",
          args: args{ 
             ...
             // includes a StatusPage.io value and ensure output is as expected.
          },
          want: finalCMProviders  // should we inline? 
     },
}
```
See [route.TestIsAdmitted()](https://github.com/openshift/console-operator/blob/ed50a574a7d78bd562a5da21e72c2bb5dab331b8/pkg/console/subresource/route/route_test.go#L117), I think we have a couple tests that do a good job of testing the boundaries of funcs

pkg/console/subresource/configmap/configmap.go
```diff
@@ -155,6 +159,17 @@ func servingInfo() yaml.MapSlice {
 	}
 }
 
+func consoleProviders(providers operatorv1.ConsoleProviders) yaml.MapSlice {
```
benjaminapetersen: This may not need testing if our main public func is well tested.  See my other test comment about adding more rows to the table test.

jhadvig:
@benjaminapetersen I've update the PR based on your comments.
Added extra top the `TestNewYamlConfig()` test and also added cleanup method to the e2e test.
Not sure if any additional tests are needed since there are only two cases:
- providers with statuspageID is set
- providers are not set

PTAL

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -302,27 +327,45 @@ func TestNewYamlConfig(t *testing.T) {
...
 				logoutRedirect: "",
 				brand:          DEFAULT_BRAND,
 				docURL:         DEFAULT_DOC_URL,
 				apiServerURL:   mockAPIServer,
+				providers:      operatorv1.ConsoleProviders{},
 			},
 			want: finalCMDefaultOKD,
 		},
+		{
+			name: "Test NewYamlConfig() with Statuspage.io provider",
```
benjaminapetersen: Great, thanks!

benjaminapetersen:
/lgtm

Thanks!

PR #235, opened 2019-06-06
Configmap/builder

benjaminapetersen:
Now that we've released & have decent test coverage, its worth looking at the codebase to see what we can improve.  I think our `DefaultConfigmap()` func has gotten to be quite the beast.  This refactor would impl the builder pattern to build up the several intermediate configs we create in the func. 

- Clarifies related work as a chain, example: `builder.Host(host).LogoutURL(url).Brand(brand)`
- Inlines the `yaml` directly into the unit tests.  This may seem like more duplication, but I believe best practices for testing is to keep all the inputs & outputs within the test func and not to share across tests.  

TODO:
- [x] write some tests for the Builder itself.  `TestDefaultConfigmap()` continues to pass, so that assures the internals are functioning correctly. 

/assign @jhadvig

### DIFF
--- a/pkg/console/subresource/configmap/configmap.go
+++ b/pkg/console/subresource/configmap/configmap.go
@@ -3,29 +3,20 @@ package configmap
 import (
 	"fmt"
 
-	"k8s.io/klog"
-
-	yaml2 "github.com/ghodss/yaml"
-	yaml "gopkg.in/yaml.v2"
-
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog"
 
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	routev1 "github.com/openshift/api/route/v1"
 	"github.com/openshift/console-operator/pkg/api"
+	"github.com/openshift/console-operator/pkg/console/subresource/consoleserver"
 	"github.com/openshift/console-operator/pkg/console/subresource/util"
-	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
 )
 
 const (
-	consoleConfigYamlFile   = "console-config.yaml"
-	clientSecretFilePath    = "/var/oauth-config/clientSecret"
-	oauthEndpointCAFilePath = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
-	// serving info
-	certFilePath = "/var/serving-cert/tls.crt"
-	keyFilePath  = "/var/serving-cert/tls.key"
+	consoleConfigYamlFile = "console-config.yaml"
 )
 
 // overridden by console config
@@ -40,54 +31,62 @@ func getApiUrl(infrastructureConfig *configv1.Infrastructure) string {
 	return ""
 }
 
-// DefaultConfigMap returns
-// - a new configmap,
-// - a bool indicating if config was merged (unsupportedConfigOverrides)
-// - an error
-func DefaultConfigMap(operatorConfig *operatorv1.Console, consoleConfig *configv1.Console, managedConfig *corev1.ConfigMap, infrastructureConfig *configv1.Infrastructure, rt *routev1.Route) (*corev1.ConfigMap, bool, error) {
+func statusPageId(operatorConfig *operatorv1.Console) string {
+	if operatorConfig.Spec.Providers.Statuspage != nil {
+		return operatorConfig.Spec.Providers.Statuspage.PageID
+	}
+	return ""
+}
 
-	// Build a default config that can be overwritten from other sources
-	host := rt.Spec.Host
-	apiServerURL := getApiUrl(infrastructureConfig)
-	defaultProviders := operatorv1.ConsoleProviders{}
-	defaultConfig := NewYamlConfig(host, defaultLogoutURL, DEFAULT_BRAND, DEFAULT_DOC_URL, apiServerURL, defaultProviders)
-	// Get console config from the openshift-config-managed namespace
+func DefaultConfigMap(
+	operatorConfig *operatorv1.Console,
+	consoleConfig *configv1.Console,
+	managedConfig *corev1.ConfigMap,
+	infrastructureConfig *configv1.Infrastructure,
+	rt *routev1.Route) (consoleConfigmap *corev1.ConfigMap, unsupportedOverridesHaveMerged bool, err error) {
+
+	defaultBuilder := &consoleserver.ConsoleServerCLIConfigBuilder{}
+	defaultConfig, err := defaultBuilder.Host(rt.Spec.Host).
+		LogoutURL(defaultLogoutURL).
+		Brand(DEFAULT_BRAND).
+		DocURL(DEFAULT_DOC_URL).
+		APIServerURL(getApiUrl(infrastructureConfig)).
+		ConfigYAML()
+
 	extractedManagedConfig := extractYAML(managedConfig)
 
-	// Derive a user defined config using operator config
-	logoutRedirect := consoleConfig.Spec.Authentication.LogoutRedirect
-	brand := operatorConfig.Spec.Customization.Brand
-	docURL := operatorConfig.Spec.Customization.DocumentationBaseURL
-	providers := operatorConfig.Spec.Providers
-	userDefinedConfig := NewYamlConfig(host, logoutRedirect, brand, docURL, apiServerURL, providers)
+	userDefinedBuilder := &consoleserver.ConsoleServerCLIConfigBuilder{}
+	userDefinedConfig, err := userDefinedBuilder.Host(rt.Spec.Host).
+		LogoutURL(consoleConfig.Spec.Authentication.LogoutRedirect).
+		Brand(operatorConfig.Spec.Customization.Brand).
+		DocURL(operatorConfig.Spec.Customization.DocumentationBaseURL).
+		APIServerURL(getApiUrl(infrastructureConfig)).
+		StatusPageID(statusPageId(operatorConfig)).
+		ConfigYAML()
 
 	unsupportedConfigOverride := operatorConfig.Spec.UnsupportedConfigOverrides.Raw
-
-	// merge configs with overrides, if we have them
-	mergedConfig, err := resourcemerge.MergeProcessConfig(nil, defaultConfig, extractedManagedConfig, userDefinedConfig, unsupportedConfigOverride)
-	if err != nil {
-		klog.Errorf("failed to merge configmap: %v", err)
-		return nil, false, err
+	willMergeConfigOverrides := len(unsupportedConfigOverride) != 0
+	if willMergeConfigOverrides {
+		klog.V(4).Infoln(fmt.Sprintf("with UnsupportedConfigOverrides: %v", string(unsupportedConfigOverride)))
 	}
 
-	outConfigYaml, err := yaml2.JSONToYAML(mergedConfig)
+	merger := &consoleserver.ConsoleYAMLMerger{}
+	mergedConfig, err := merger.Merge(
+		defaultConfig,
+		extractedManagedConfig,
+		userDefinedConfig,
+		unsupportedConfigOverride)
 	if err != nil {
 		klog.Errorf("failed to generate configmap: %v", err)
 		return nil, false, err
 	}
 
-	// if we actually merged config overrides, log this information
-	didMerge := len(operatorConfig.Spec.UnsupportedConfigOverrides.Raw) != 0
-	if didMerge {
-		klog.V(4).Infoln(fmt.Sprintf("with UnsupportedConfigOverrides: %v", string(unsupportedConfigOverride)))
-	}
-
 	configMap := Stub()
 	configMap.Data = map[string]string{}
-	configMap.Data[consoleConfigYamlFile] = string(outConfigYaml)
+	configMap.Data[consoleConfigYamlFile] = string(mergedConfig)
 	util.AddOwnerRef(configMap, util.OwnerRefFrom(operatorConfig))
 
-	return configMap, didMerge, nil
+	return configMap, willMergeConfigOverrides, nil
 }
 
 func DefaultPublicConfig(consoleURL string) *corev1.ConfigMap {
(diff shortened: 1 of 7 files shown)


benjaminapetersen:
@zherman0 fyi

benjaminapetersen:
I'm interested in adding a `Builder` for the `OperatorStatus` stuff as well. I believe other operators (master?) use it to capture information in a cleaner & more detailed way.  I have some work on this going as well.

pkg/console/subresource/configmap/configmap.go
```diff
@@ -40,54 +36,62 @@ func getApiUrl(infrastructureConfig *configv1.Infrastructure) string {
...
+}
+
+func DefaultConfigMap(
+	operatorConfig *operatorv1.Console,
+	consoleConfig *configv1.Console,
+	managedConfig *corev1.ConfigMap,
+	infrastructureConfig *configv1.Infrastructure,
+	rt *routev1.Route) (consoleConfigmap *corev1.ConfigMap, unsupportedConfigMerged bool, err error) {
+
+	b := &ConsoleYAMLConfigBuilder{}
```
jhadvig: nit: would rather have named variables.. here and below :)
```suggestion
	defaultBuilder := &ConsoleYAMLConfigBuilder{}
```

pkg/console/subresource/configmap/configmap.go
```diff
@@ -40,54 +36,62 @@ func getApiUrl(infrastructureConfig *configv1.Infrastructure) string {
...
-	docURL := operatorConfig.Spec.Customization.DocumentationBaseURL
-	providers := operatorConfig.Spec.Providers
-	userDefinedConfig := NewYamlConfig(host, logoutRedirect, brand, docURL, apiServerURL, providers)
+	b2 := &ConsoleYAMLConfigBuilder{}
+	userDefinedConfig, err := b2.Host(rt.Spec.Host).
+		LogoutURL(consoleConfig.Spec.Authentication.LogoutRedirect).
+		Brand(operatorConfig.Spec.Customization.Brand).
+		DocURL(operatorConfig.Spec.Customization.DocumentationBaseURL).
+		APIServerURL(getApiUrl(infrastructureConfig)).
+		StatusPageID(statusPageId(operatorConfig)).
```
jhadvig: not sure if we shouldnt have public function per config field, eg.:
- Customization (instead of Brand and DocURL)
- Providers (instead of StatusPageID)
- ...


each of this suggested function can have subfunctions that can be tested individually but also as a whole...
...just thinking out loud, so we find a proper way how to deal with this fields :)

pkg/console/subresource/configmap/configmap.go
```diff
@@ -40,54 +36,62 @@ func getApiUrl(infrastructureConfig *configv1.Infrastructure) string {
...
 	unsupportedConfigOverride := operatorConfig.Spec.UnsupportedConfigOverrides.Raw
-
-	// merge configs with overrides, if we have them
-	mergedConfig, err := resourcemerge.MergeProcessConfig(nil, defaultConfig, extractedManagedConfig, userDefinedConfig, unsupportedConfigOverride)
-	if err != nil {
-		logrus.Errorf("failed to merge configmap: %v \n", err)
-		return nil, false, err
+	willMergeConfigOverrides := len(operatorConfig.Spec.UnsupportedConfigOverrides.Raw) != 0
+	if willMergeConfigOverrides {
+		logrus.Println(fmt.Sprintf("with UnsupportedConfigOverrides: %v", string(unsupportedConfigOverride)))
```
jhadvig: We will have to replace this by `klog` right ?

pkg/console/subresource/configmap/configmap.go
```diff
@@ -40,54 +36,62 @@ func getApiUrl(infrastructureConfig *configv1.Infrastructure) string {
...
+		ConfigBytes()
 
 	unsupportedConfigOverride := operatorConfig.Spec.UnsupportedConfigOverrides.Raw
-
-	// merge configs with overrides, if we have them
-	mergedConfig, err := resourcemerge.MergeProcessConfig(nil, defaultConfig, extractedManagedConfig, userDefinedConfig, unsupportedConfigOverride)
-	if err != nil {
-		logrus.Errorf("failed to merge configmap: %v \n", err)
-		return nil, false, err
+	willMergeConfigOverrides := len(operatorConfig.Spec.UnsupportedConfigOverrides.Raw) != 0
```
jhadvig: why not use the new var ?
```suggestion
	willMergeConfigOverrides := len(unsupportedConfigOverride) != 0
```

pkg/console/subresource/configmap/configmap.go
```diff
@@ -40,54 +36,62 @@ func getApiUrl(infrastructureConfig *configv1.Infrastructure) string {
...
+		logrus.Println(fmt.Sprintf("with UnsupportedConfigOverrides: %v", string(unsupportedConfigOverride)))
 	}
 
-	outConfigYaml, err := yaml2.JSONToYAML(mergedConfig)
+	merger := &ConsoleYAMLMerger{}
+	mergedConfig, err := merger.ToYaml(
+		defaultConfig,
+		extractedManagedConfig,
+		userDefinedConfig,
+		operatorConfig.Spec.UnsupportedConfigOverrides.Raw)
```
jhadvig: why not use the new var here as well ?
```suggestion
		unsupportedConfigOverride)
```

pkg/console/subresource/configmap/configmap_builder.go
```diff
@@ -0,0 +1,156 @@
...
+)
+
+// ConsoleYamlConfigBuilder
+// Director will be DefaultConfigMap()
+//
+// b := ConsoleYamlConfigBuilder{}
+// return the default config value immediately:
+//   b.Config()
+// set all the values:
+//   b.SetHost(host).SetLogoutURL().setBrand().setDocURL().SetAPIServerURL().SetProductName().SetCustomLogoFile().Config()
```
jhadvig: The method names here are not reflecting the method names from the `YAMLConfigBuilder` interface below

pkg/console/subresource/configmap/configmap_builder.go
```diff
@@ -0,0 +1,156 @@
...
+	conf = append(conf, yaml.MapItem{Key: "apiVersion", Value: "console.openshift.io/v1"})
+	conf = append(conf, yaml.MapItem{Key: "auth", Value: b.authServer(b.logoutRedirectURL)})
+	conf = append(conf, yaml.MapItem{Key: "clusterInfo", Value: b.clusterInfo(b.host, b.apiServerURL)})
+	conf = append(conf, yaml.MapItem{Key: "customization", Value: b.customization(b.brand, b.docURL)})
+	conf = append(conf, yaml.MapItem{Key: "servingInfo", Value: b.servingInfo()})
+	conf = append(conf, yaml.MapItem{Key: "providers", Value: b.providers(b.statusPageID)})
+	return conf
+}
+
+func (b *ConsoleYAMLConfigBuilder) ConfigBytes() (consoleConfigYAML []byte, marshallError error) {
```
jhadvig: Not sure about the name here, maybe `ConfigToBytes()` ? (speaking as a not native person)...
Also we can probably drop named return values.

pkg/console/subresource/configmap/configmap.go
jhadvig: I see now that you created a new struct for the `ConsoleYAMLConfigBuilder` ... in that case I think it makes sense to have field for each property .... thinking about the downsides though ...

pkg/console/subresource/configmap/configmap.go
benjaminapetersen: correct, depending on which PR merges first.

pkg/console/subresource/configmap/configmap.go
benjaminapetersen: Yeah I want to eliminate this interim `yaml.MapSlice{ }` stuff.  Its a pain.  When we can get [config.go](https://github.com/openshift/console/blob/master/cmd/bridge/config.go) from the console:

```golang
type Config struct {
	APIVersion    string `yaml:"apiVersion"`   // yay! yaml tags! wooo
	Kind          string `yaml:"kind"`                  // yay! yaml tags! wooo
	ServingInfo   `yaml:"servingInfo"`            // yay! yaml tags! wooo
        // stuff....
	Providers     `yaml:"providers"`                // yay! yaml tags! wooo
}
// more stuff....
```
We can just take advantage of the `yaml` struct tags like for the Marshalling.

benjaminapetersen: Maybe I should just copy over that config in this PR.

pkg/console/subresource/configmap/configmap.go
benjaminapetersen: agree, I'll update that.

pkg/console/subresource/configmap/configmap.go
benjaminapetersen: Yeah I think we want to have an interface that takes `1:1` items as they are set.... and behind the scenes will generate the complicated structure & stamp out the YAML if asked.

pkg/console/subresource/configmap/configmap.go
benjaminapetersen: will do.

pkg/console/subresource/configmap/configmap_builder.go
benjaminapetersen: `ConfigYAML()`  is better than `ConfigToBytes()` yeah?

I kept the named return values only when it matters: `consoleConfigYAML []byte, marshallError error` is helpful so you know `[]byte` is YAML, otherwise the function signature isn't very clear.

pkg/console/subresource/configmap/configmap_builder.go
benjaminapetersen: Removed the interface as its not needed.

pkg/console/subresource/configmap/configmap.go
benjaminapetersen: Will do.

pkg/console/subresource/configmap/configmap.go
benjaminapetersen: Got it with the rebase.

pkg/console/subresource/configmap/configmap.go
benjaminapetersen: Update: Copied over the Config into `subresource/consoleserver/config.go`.

Was able to eliminate that extra YAML library as we can take advantage of the `yaml` tags now.

pkg/console/subresource/consoleserver/config_builder.go
```diff
@@ -0,0 +1,143 @@
...
+)
+
+// ConsoleServerCLIConfigBuilder
+// Director will be DefaultConfigMap()
+//
+// b := ConsoleYamlConfigBuilder{}
+// return the default config value immediately:
+//   b.Config()
+// set all the values:
+//   b.SetHost(host).SetLogoutURL().setBrand().setDocURL().SetAPIServerURL().SetProductName().SetCustomLogoFile().Config()
```
jhadvig: The method names in docs are not reflecting the real names.

pkg/console/subresource/consoleserver/config_builder.go
```diff
@@ -0,0 +1,143 @@
...
+		ServingInfo:   b.servingInfo(),
+		Providers:     b.providers(),
+	}
+}
+
+func (b *ConsoleServerCLIConfigBuilder) ConfigYAML() (consoleConfigYAML []byte, marshallError error) {
+	conf := b.Config()
+	yml, err := yaml.Marshal(conf)
+	if err != nil {
+		klog.V(4).Infof("Could not create config yaml %v", err)
```
jhadvig:
```suggestion
		klog.V(4).Infof("could not create config yaml %v", err)
```

pkg/console/subresource/configmap/configmap.go
jhadvig: yeah, but shouldn't we rather import the pkg from the `console` repo ?

pkg/console/subresource/configmap/configmap.go
```diff
@@ -40,54 +32,62 @@ func getApiUrl(infrastructureConfig *configv1.Infrastructure) string {
...
-	outConfigYaml, err := yaml2.JSONToYAML(mergedConfig)
+	merger := &consoleserver.ConsoleYAMLMerger{}
+	mergedConfig, err := merger.Merge(
+		defaultConfig,
+		extractedManagedConfig,
+		userDefinedConfig,
+		operatorConfig.Spec.UnsupportedConfigOverrides.Raw)
 	if err != nil {
-		klog.Errorf("failed to generate configmap: %v", err)
+		klog.Errorf("failed to generate configmap: %v \n", err)
```
jhadvig: no need to add... `klog` will deal with it for us :)
```suggestion
		klog.Errorf("failed to generate configmap: %v", err)
```

pkg/console/subresource/consoleserver/config.go
```diff
@@ -0,0 +1,65 @@
+package consoleserver
```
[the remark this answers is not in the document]
jhadvig: I've opened a [PR](https://github.com/openshift/library-go/pull/454) thats adding all this types to `library-go` so it can be shared across our repos.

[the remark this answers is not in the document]
benjaminapetersen: Yeah if we can put it somewhere that makes sense. Def realized not OpenShift/API.

pkg/console/subresource/consoleserver/config_builder.go
benjaminapetersen: I'll remove or update. Was looking into idiomatic go, the word "set" is not typically used.

pkg/console/subresource/consoleserver/config_builder.go
benjaminapetersen: Nice thx!

pkg/console/subresource/configmap/configmap.go
benjaminapetersen: Ah forgot to update.

benjaminapetersen:
Thx, the name is long... But we have so many configs that figured best be specific.

pkg/console/subresource/configmap/configmap.go
benjaminapetersen: fixeed.

pkg/console/subresource/configmap/configmap.go
```diff
@@ -40,54 +32,62 @@ func getApiUrl(infrastructureConfig *configv1.Infrastructure) string {
...
-	defaultConfig := NewYamlConfig(host, defaultLogoutURL, DEFAULT_BRAND, DEFAULT_DOC_URL, apiServerURL, defaultProviders)
-	// Get console config from the openshift-config-managed namespace
+func statusPageId(operatorConfig *operatorv1.Console) string {
+	if operatorConfig.Spec.Providers.Statuspage != nil {
+		return operatorConfig.Spec.Providers.Statuspage.PageID
+	}
+	return ""
+}
+
+func DefaultConfigMap(
```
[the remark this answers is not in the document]
benjaminapetersen: I dropped the comments in favor of named return values, since thats what the comments were really clarifying `(consoleConfigmap *corev1.ConfigMap, unsupportedOverridesHaveMerged bool, err error)`

PR #236, opened 2019-06-06
[WIP] Devconsole crd

DhritiShikhar:
To enable developer perspective in openshift

### DIFF
--- a/manifests/09-devconsole-crd.yaml
+++ b/manifests/09-devconsole-crd.yaml
@@ -1,0 +1,18 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: devconsoles.operator.openshift.io
+spec:
+  scope: Namespace
+  group: operator.openshift.io
+  names:
+    kind: Devconsole
+    listKind: DevconsoleList
+    plural: devconsoles
+    singular: devconsole
+  subresources:
+    status: {}
+  versions:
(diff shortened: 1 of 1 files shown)


spadgett:
> To enable developer perspective in openshift

This should be a flag on the existing console config object: https://github.com/openshift/api/blob/master/config/v1/types_console.go

/hold

sbose78:
Sam , using the existing console-config object might make it unusable in OSDs.
I've outlined options in https://docs.google.com/document/d/1B_ScB3uMEM1ziAC_NE5rEMLo3IK8Ug3BGAZBdENjPhk

benjaminapetersen:
It is also not clear to me why we need an additional config object.

spadgett:
> Sam , using the existing console-config object might make it unusable in OSDs.

This will be a problem for every console config flag, not just dev console. Creating a separate CRD for each feature doesn't scale and is inconsistent with how other components handle configuration. We need to address console configuration generally in OSD.

cc @bparees

bparees:
> This will be a problem for every console config flag, not just dev console. Creating a separate CRD for each feature doesn't scale and is inconsistent with how other components handle configuration. We need to address console configuration generally in OSD.

+1.  We need to talk to the OSD team and understand the concerns/risks if a dedicated-cluster-admin is able to edit the console config resource.  Or what other mechanism OSD might offer to allow the customer to control this configuration.

PR #237, opened 2019-06-10
Update example operator yaml with alt image

benjaminapetersen:
/assign @jhadvig @zherman0 

Trivial example update since we dropped that flag.

### DIFF
--- a/examples/07-operator-alt-image.yaml
+++ b/examples/07-operator-alt-image.yaml
@@ -31,8 +31,8 @@ spec:
         - console
         - operator
         args:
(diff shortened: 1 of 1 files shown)


spadgett:
Do we need this example at all? It seems like we should tell the CVO not to update the console operator deployment and just use `oc set image` on the existing operator deployment, similar to what we recommend for console here: https://github.com/openshift/console#deploying-a-custom-image-to-an-openshift-cluster

benjaminapetersen:
@spadgett we probably do not need it anymore & could update the readme.

PR #243, opened 2019-06-17
Promote sync() funcs to methods on the operator

benjaminapetersen:
Promotes the `sync()` funcs to Methods on the Operator from standalone funcs.  

/assign @jhadvig

### DIFF
--- a/pkg/console/operator/operator.go
+++ b/pkg/console/operator/operator.go
@@ -143,6 +143,12 @@ func (c *consoleOperator) Key() (metav1.Object, error) {
 	return c.operatorConfigClient.Get(api.ConfigResourceName, metav1.GetOptions{})
 }
 
+type configSet struct {
+	Console        *configv1.Console
+	Operator       *operatorsv1.Console
+	Infrastructure *configv1.Infrastructure
+}
+
 func (c *consoleOperator) Sync(obj metav1.Object) error {
 	startTime := time.Now()
 	klog.V(4).Infof("started syncing operator %q (%v)", obj.GetName(), startTime)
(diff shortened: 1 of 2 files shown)


pkg/console/operator/sync_v400.go
```diff
@@ -408,3 +398,11 @@ func getDeploymentGeneration(co *consoleOperator) int64 {
 	}
 	return deployment.Generation
 }
+
```
benjaminapetersen: This is just a move, the helpers should be collected together at the bottom.

pkg/console/operator/sync_v400.go
```diff
@@ -82,7 +80,7 @@ func sync_v400(co *consoleOperator, operatorConfig *operatorv1.Console, consoleC
 	}
 	toUpdate = toUpdate || serviceCAConfigMapChanged
 
-	sec, secChanged, secErr := SyncSecret(co, recorder, operatorConfig)
+	sec, secChanged, secErr := co.SyncSecret(operatorConfig)
```
benjaminapetersen: `recorder` is attached to the `operator` as `oc.recorder`.

pkg/console/operator/operator.go
```diff
@@ -143,6 +143,12 @@ func (c *consoleOperator) Key() (metav1.Object, error) {
 	return c.operatorConfigClient.Get(api.ConfigResourceName, metav1.GetOptions{})
 }
 
+type configSet struct {
```
benjaminapetersen: In addition, we can collect all the configs into a set & pass this, to also make it easier to pipe configs through.

benjaminapetersen: Note they are deliberately not embedded.

pkg/console/operator/operator.go
```diff
@@ -165,58 +171,64 @@ func (c *consoleOperator) Sync(obj metav1.Object) error {
...
 		return err
 	}
 
 	return nil
 }
 
-func (c *consoleOperator) handleSync(operatorConfig *operatorsv1.Console, consoleConfig *configv1.Console, infrastructureConfig *configv1.Infrastructure) error {
-	operatorConfigCopy := operatorConfig.DeepCopy()
+func (c *consoleOperator) handleSync(configs configSet) error {
+	updatedStatus := configs.Operator.DeepCopy()
```
jhadvig: not sure why `updatedStatus` ?

pkg/console/operator/sync_v400.go
```diff
@@ -169,12 +167,14 @@ func sync_v400(co *consoleOperator, operatorConfig *operatorv1.Console, consoleC
 		return err
 	}
 
-	if _, err := SyncConsoleConfig(co, consoleConfig, consoleURL); err != nil {
+	// TODO: we are mutating the original here.  Thats not great.  We should copy
+	// and update, like we do with the operator config.
```
jhadvig: Do we want to fix that here or as a followUp ?

pkg/console/operator/operator.go
jhadvig: ok, I see now after going though the whole PR, but still not sure about the name. Liked the original `operatorConfigCopy` more, since it was reflecting the object semantics... but opened for discussion

benjaminapetersen: Yeah, the tricky this is that it IS a `copy` of the whole `config`, but it is FOR status updates, only.  I think that what I'm aiming to express.

pkg/console/operator/sync_v400.go
benjaminapetersen: I'll do a `console.DeepCopy()` inside that func right before the `.UpdateStatus()` call.  That should do the trick.

pkg/console/operator/operator.go
jhadvig: Makes sense

PR #246, opened 2019-06-24
Add console extension manfiest for ConsoleExternalLogLinks CRD

rhamilto:
Resolves https://jira.coreos.com/browse/CONSOLE-1547

### DIFF
--- a/manifests/00-crd-extension-console-log-link.yaml
+++ b/manifests/00-crd-extension-console-log-link.yaml
@@ -1,0 +1,66 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: consoleexternalloglinks.console.openshift.io
+  annotations:
+    displayName: ConsoleExternalLogLinks
+    description: ConsoleExternalLogLink is an extension for customizing OpenShift web console log links.
+spec:
+  scope: Cluster
+  group: console.openshift.io
+  versions:
+  - name: v1
+    served: true
+    storage: true
+  names:
+    plural: consoleexternalloglinks
+    singular: consoleexternalloglink
+    kind: ConsoleExternalLogLink
+    listKind: ConsoleExternalLogLinkList
+  additionalPrinterColumns:
+  - name: Text
+    type: string
+    JSONPath: .spec.text
+  - name: HrefTemplate
+    type: string
+    JSONPath: .spec.hrefTemplate
+  - name: Age
+    type: date
+    JSONPath: .metadata.creationTimestamp
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      properties:
+        spec:
+          type: object
+          description: >-
+            ConsoleExternalLogLinkSpec is the desired log link configuration.
+            The log link will appear on the logs tab of the pod details page.
+          required:
+          - text
+          - hrefTemplate
+          properties:
+            text:
+              type: string
+              description: text is the display text for the link
+            hrefTemplate:
+              type: string
+              description: |-
+                hrefTemplate is an absolute secure URL (must use https) for the log link including variables to be replaced. Variables are specified in the URL with the format ${variableName}, for instance, ${containerName} and will be replaced with the corresponding values from the resource. Resource is a pod.
+                Supported variables are
+                  * ${resourceName} - name of the resource which containes the logs
+                  * ${resourceUID} - UID of the resource which contains the logs
+                    * e.g. `11111111-2222-3333-4444-555555555555`
+                  * ${containerName} - name of the resource's container that contains the logs
+                  * ${resourceNamespace} - namespace of the resource that contains the logs
+                  * ${podLabels} - JSON representation of labels matching the pod with the logs
+                    * e.g. `{"key1":"value1","key2":"value2"}`
+                e.g., https://example.com/logs?resourceName=${resourceName}&containerName=${containerName}&resourceNamespace=${resourceNamespace}&podLabels=${podLabels}
+              pattern: '^https://'
+            namespaceFilter:
+              type: string
+              description: >-
+                  namespaceFilter is a regular expression used to restrict a log link to a
+                  matching set of namespaces (e.g., `openshift-`). If not specified, links will
+                  be displayed for all the namespaces.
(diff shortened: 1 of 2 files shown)


manifests/00-crd-extension-console-pod-log-link.yaml
```diff
@@ -0,0 +1,48 @@
...
+          required:
+          - text
+          - hrefTemplate
+          properties:
+            text:
+              type: string
+              description: Text of the link
+            hrefTemplate:
+              type: string
+              description: Absolute secure URL including pod name or uid, container name, namespace, and labels variables for the pod log link (must use https).  e.g., https://example.com/logs?name=\${name}&container=\${container}&namespace=\${namespace}&labels=\${labels}
```
spadgett: We'll want to document specifically all the replacements we support and the exact format

manifests/00-crd-extension-console-pod-log-link.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: consolepodloglinks.console.openshift.io
+  annotations:
+    displayName: ConsolePodLogLinks
```
benjaminapetersen: And per our slack, considering a name change to make more generic

benjaminapetersen:
/hold 

Note that `ci/prow/e2e-aws` will fail unless we bump the `rbac` test again.  

Since we are holding off on this PR for design, We will likely have the CRD proxy for the console server done by then, so shouldn't need to do this.

rhamilto:
Updated.  @spadgett, PTAL.

benjaminapetersen:
Needs an update like this one for the tests to pass:
https://github.com/openshift/origin/pull/23231
Working on that now.

manifests/03-rbac-role-cluster-extensions.yaml
```diff
@@ -11,6 +11,7 @@ rules:
   - consolelinks
   - consoleclidownloads
   - consolenotifications
+  - consoleloglinks
```
benjaminapetersen: `consoleexternalloglinks`

https://github.com/openshift/origin/pull/23382/files#diff-7a19fbef2a463f9e30c55aa61cf6f5ffR119

manifests/00-crd-extension-console-log-link.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: consoleloglinks.console.openshift.io
```
benjaminapetersen: `ConsoleExternalLogLink` now 💩

manifests/00-crd-extension-console-log-link.yaml
```diff
@@ -0,0 +1,54 @@
...
+              type: string
+              description: Text of the link
+            hrefTemplate:
+              type: string
+              description: >-
+                Absolute secure URL (must use https) for the log link including variables to be replaced.
+                Supported variables include ${resourceName}, ${resourceUID}, ${containerName}, ${resourceNamespace}, and ${podLabels} and must match exactly.
+                e.g., https://example.com/logs?resourceName=\${resourceName}&containerName=\${containerName}&resourceNamespace=\${resourceNamespace}&podLabels=\${podLabels}
+              pattern: '^https://'
+            namespaceFilter:
```
benjaminapetersen: Trying to decide if we want a pattern on this. For example:
`^openshift-` vs `/^openshift-/` vs `/^openshift-/g` 

We are using `string.search(regex)`, so `^openshift-` will break, but the other two will work. [This stack overflow](https://stackoverflow.com/questions/172303/is-there-a-regular-expression-to-detect-a-valid-regular-expression) says that this is a perfect regex to check for a regex:

`/^((?:(?:[^?+*{}()[\]\\|]+|\\.|\[(?:\^?\\.|\^[^\\]|[^\\^])(?:[^\]\\]+|\\.)*\]|\((?:\?[:=!]|\?<[=!]|\?>)?(?1)??\)|\(\?(?:R|[+-]?\d+)\))(?:(?:[?+*]|\{\d+(?:,\d*)?\})[?+]?)?|\|)*)$/
`
I kid about including that 😄 

To make cross reference easy: https://github.com/openshift/console/pull/1892/files#r304608136

We may want to do something, however? At least check for leading `/`?  I don't think we should necessarily try to be comprehensive, that sounds like bugs.

spadgett:
@rhamilto we should match the descriptions here to what's in openshift/api since it's what's added to the OpenAPI doc (oc explain, API Explorer, etc.)

Note there is one minor change: https://github.com/openshift/api/pull/385

manifests/00-crd-extension-console-log-link.yaml
spadgett: Per discussion in https://github.com/openshift/api/pull/385, leading `/` is not necessary. I'm not sure what validation is really practical here. I think this is reasonable for now, and we can improve validation as a bug fix if needed.

benjaminapetersen: Works for me.

manifests/00-crd-extension-console-log-link.yaml
```diff
@@ -0,0 +1,54 @@
...
+        spec:
+          type: object
+          description: Represents console external log link customizations spec
+          required:
+          - text
+          - hrefTemplate
+          properties:
+            text:
+              type: string
+              description: Text of the link
```
benjaminapetersen: `Text is the display text of the link`

https://github.com/openshift/api/pull/380/files#diff-9754f97c9c90951cc7df5ac76b6eea86R20

benjaminapetersen:
@rhamilto want to retitle PR for `ConsoleExternalLogLink`? (just for future reference parity)

manifests/00-crd-extension-console-log-link.yaml
```diff
@@ -0,0 +1,54 @@
...
+            hrefTemplate:
```
benjaminapetersen: `hrefTemplate` description is pretty close, though not exact.  Do we want to just copy/paste to make maintenance over time easy?

rhamilto:
Descriptions updated to align with api.  Note I had to make some formatting adjustments to make the text yaml-friendly (removed colon after `Supported variables are` and changed `-` bullet to `*`.
@spadgett, @benjaminapetersen

spadgett:
Feel free to remove the hold when https://github.com/openshift/origin/pull/23382 lands

PR #247, opened 2019-06-25
Update link location pattern regex

rhamilto:


manifests/00-crd-extension-console-link.yaml
```diff
@@ -53,4 +53,4 @@ spec:
             location:
               type: string
               description: Determines which dropdown menu the link is added to (HelpMenu, UserMenu)
-              pattern: '\bHelpMenu\b|\bUserMenu\b'
+              pattern: '^(HelpMenu|UserMenu)$'
```
jhadvig: the quotes can be dropped
```suggestion
              pattern: ^(HelpMenu|UserMenu)$
```

PR #250, opened 2019-06-28
CONSOLE-1523: Change log level values to use klog types

zherman0:
This changes the capnslog values to accepted klog log level values.  I have just converted them as best I can but we might want to re-evaluate what each maps to.

[CONSOLE-1523](https://issues.redhat.com/browse/CONSOLE-1523)

### DIFF
--- a/pkg/console/subresource/deployment/deployment.go
+++ b/pkg/console/subresource/deployment/deployment.go
@@ -236,15 +236,15 @@ func consoleVolumeMounts(vc []volumeConfig) []corev1.VolumeMount {
 
 func GetLogLevelFlag(logLevel operatorv1.LogLevel) string {
 	flag := ""
-	// Since the console-operator logging has different logging levels then the capnslog,
-	// that we use for console server(bridge) we need to map them to each other
 	switch logLevel {
 	case operatorv1.Normal:
-		flag = "--log-level=*=NOTICE"
+		flag = "--v=2"
 	case operatorv1.Debug:
-		flag = "--log-level=*=DEBUG"
-	case operatorv1.Trace, operatorv1.TraceAll:
-		flag = "--log-level=*=TRACE"
+		flag = "--v=4"
+	case operatorv1.Trace:
+		flag = "--v=6"
+	case operatorv1.TraceAll:
+		flag = "--v=10"
 	}
 	return flag
 }


zherman0:
/hold 
hold until Console https://github.com/openshift/console/pull/1862 merges

pkg/console/subresource/deployment/deployment.go
```diff
@@ -221,11 +221,11 @@ func GetLogLevelFlag(logLevel operatorv1.LogLevel) string {
 	flag := ""
 	switch logLevel {
 	case operatorv1.Normal:
-		flag = "--log-level=*=NOTICE"
+		flag = "ERROR"
 	case operatorv1.Debug:
-		flag = "--log-level=*=DEBUG"
+		flag = "INFO"
 	case operatorv1.Trace, operatorv1.TraceAll:
-		flag = "--log-level=*=TRACE"
```
jhadvig: Based on the [klog docs](https://github.com/kubernetes/klog/blob/master/klog.go#L98-L100) I would suggest to set the severity as:

```go
	case operatorv1.Normal:
		flag = "INFO"
	case operatorv1.Debug:
		flag = "WARNING"
	case operatorv1.Trace:
                flag = "ERROR"
        case operatorv1.TraceAll:
		flag = "FATAL"
	}
```

zherman0: Those seem reasonable; the first round were mainly placeholders to discuss how to best map them in the future.  Thanks for the input.

benjaminapetersen:
@zherman0 this one has sat for a bit, finally getting the open PRs list down enough to take a look.  Is there a matching console PR? I assume both need to change at more or less the same time.  Or does the console have some logic to ensure we don't get in a bad state due to flag mismatch?

benjaminapetersen:
Button click misfire :)

zherman0:
@benjaminapetersen @spadgett - There is a a corresponding console PR: https://github.com/openshift/console/pull/1862
Currently, the changes I made required importing api machinery (I think) which makes it a XXL change and Sam was not sure he wanted all that extra weight, so we put it in `hold`.
There might be a way to re-write the klog stuff on the console side to not use the specific library but I am not sure we want to do that. Any thoughts if we should close these for now?

benjaminapetersen:
@zherman0 so the `klog` [go.mod file](https://github.com/kubernetes/klog/blob/master/go.mod) is small:

```
module k8s.io/klog/v2

go 1.12

require github.com/go-logr/logr v0.1.0
```

Curious, do you need the [kubernetes component base](https://github.com/openshift/console/pull/1862/commits/b70ecdb9941e32126166a13628f906a0322a6fe9#diff-37aff102a57d3d7b797f152915a6dc16R15)?

I'm guessing thats why the PR blew up.

pkg/console/subresource/deployment/deployment.go
benjaminapetersen: Does this work? Looking at [line 250](https://github.com/openshift/console-operator/pull/250/files#diff-8c5d48b60d928752e02d8916d831a48eR250), typically flags are as:

```
--key=value
```
So we are turning 
```
--log-level=*=NOTICE
```
into just

```
INFO
```
Seems like this will create an invalid command in our deployment func:

```
		Command: []string{
			"/opt/bridge/bin/bridge",
			"--public-dir=/opt/bridge/static",
			"--config=/var/console-config/console-config.yaml",
			"--service-ca-file=/var/service-ca/service-ca.crt",
			flag,
		},
```
Make sure you test again with a log level explicitly set, not sure this is quite right.

benjaminapetersen: IE, if `oc apply -f` a file like this:

```
apiVersion: operator.openshift.io/v1
kind: Console
metadata:
  name: cluster
spec:
  managementState: Managed
  logLevel: Trace
```
Does the console pod still run?

benjaminapetersen: I believe the new log level should be something like 

```
 -v=X
``` 
Similar to the `console-operator` config itself.  You can find this in [manifests/07-operator.yaml](https://github.com/openshift/console-operator/blob/master/manifests/07-operator.yaml)

```
command:
- console
- operator
args:
# this is the operator's klog setting
- "-v=6"
- "--config=/var/run/configmaps/config/controller-config.yaml"
```

benjaminapetersen: So, given the above, I think the `GetLogLevel()` func probably needs to be more like this:

```golang
func GetLogLevelFlag(logLevel operatorv1.LogLevel) string {
	flag := ""
	switch logLevel {
	case operatorv1.Normal:
		flag = "-v=2"
	case operatorv1.Debug:
		flag = "-v=4"
	case operatorv1.Trace:
		flag = "-v=6"
	case operatorv1.TraceAll:
		flag = "-v=10"
	}
	return flag
}
```

benjaminapetersen:
I'll get this if I run the code currently:

Screen Shot 2019-12-02 at 12 17 08 PM

benjaminapetersen:
So as a control, if I run `console-operator` as `ManagementState: Unmanaged`, so that I can pass either `TRACE` or `-v=4` (either alternative to the log level flag) I will see a `CrashLoopBackoff`:

Screen Shot 2019-12-02 at 12 17 08 PM

The `e2e` tests pass here likely because we don't actually check that pods are all happy.

pkg/console/subresource/deployment/deployment.go
```diff
@@ -238,15 +238,16 @@ func consoleVolumeMounts(vc []volumeConfig) []corev1.VolumeMount {
 
 func GetLogLevelFlag(logLevel operatorv1.LogLevel) string {
 	flag := ""
-	// Since the console-operator logging has different logging levels then the capnslog,
-	// that we use for console server(bridge) we need to map them to each other
+	// console server(bridge) expects flag log-level
 	switch logLevel {
 	case operatorv1.Normal:
-		flag = "--log-level=*=NOTICE"
+		flag = "--log-level=2"
```
benjaminapetersen: We will want the `klog` expected `flag` long term, this should be:

```
-v=2
-v=4
-v=6
-v=10
```

benjaminapetersen: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md

Nevermind, your #s are good.

zherman0:
For log level values, I am using this doc: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md

pkg/console/subresource/deployment/deployment.go
benjaminapetersen: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md

disregard, these #s are fine.

benjaminapetersen: These are still 

```
--log-level=2
```
But we need to update to 

```
-v=2
```

Wrong flag.

zherman0: I am leaving this like this for now since console can handle the old values.  We will get console portion merged first and then update this.  I will update the code but I want to make sure this does not get merged.

benjaminapetersen: This PR won't merge until the `-v` flag is updated, so you should be safe.

pkg/console/subresource/deployment/deployment.go
```diff
@@ -238,15 +238,16 @@ func consoleVolumeMounts(vc []volumeConfig) []corev1.VolumeMount {
 
 func GetLogLevelFlag(logLevel operatorv1.LogLevel) string {
 	flag := ""
-	// Since the console-operator logging has different logging levels then the capnslog,
-	// that we use for console server(bridge) we need to map them to each other
+	// console server(bridge) expects flag log-level
```
benjaminapetersen: This comment can go away, it will no longer expect `log-level` after these merge.

pkg/console/subresource/deployment/deployment.go
```diff
@@ -238,15 +238,15 @@ func consoleVolumeMounts(vc []volumeConfig) []corev1.VolumeMount {
...
+		flag = "-v=2"
 	case operatorv1.Debug:
-		flag = "--log-level=*=DEBUG"
-	case operatorv1.Trace, operatorv1.TraceAll:
-		flag = "--log-level=*=TRACE"
+		flag = "-v=3"
+	case operatorv1.Trace:
+		flag = "-v=4"
+	case operatorv1.TraceAll:
+		flag = "-v=5"
```
jhadvig, line 249:
```suggestion
		flag = "-v=10"
```

pkg/console/subresource/deployment/deployment.go
```diff
@@ -238,15 +238,15 @@ func consoleVolumeMounts(vc []volumeConfig) []corev1.VolumeMount {
...
-	// that we use for console server(bridge) we need to map them to each other
 	switch logLevel {
 	case operatorv1.Normal:
-		flag = "--log-level=*=NOTICE"
+		flag = "-v=2"
 	case operatorv1.Debug:
-		flag = "--log-level=*=DEBUG"
-	case operatorv1.Trace, operatorv1.TraceAll:
-		flag = "--log-level=*=TRACE"
+		flag = "-v=3"
```
jhadvig, line 245:
```suggestion
		flag = "-v=4"
```

pkg/console/subresource/deployment/deployment.go
```diff
@@ -238,15 +238,15 @@ func consoleVolumeMounts(vc []volumeConfig) []corev1.VolumeMount {
...
+	case operatorv1.Trace:
+		flag = "-v=4"
```
jhadvig, line 247:
```suggestion
		flag = "-v=6"
```

jhadvig (changes requested):
Aligning to the values that we are using in the Console's Switch to klog PR

spadgett:
/assign @yapei @ahardin-rh @sferich888 
for QE / docs / CEE approval

This is ready to test. @jhadvig @zherman0 I assume this should be tested with https://github.com/openshift/console/pull/1862 together using cluster-bot?

spadgett (approved):
/approve

jhadvig:
@spadgett was successfully able to launch a cluster using cluster with combination of both PRs.
`launch openshift/console-operator#250,openshift/console#1862`

Checked if the `--v=2` flag was present in the console deployment
Screenshot 2020-10-20 at 17 56 20

Also check the `console` pod if it's running:
Screenshot 2020-10-20 at 17 57 43

yapei:
Launched cluster `launch openshift/console-operator#250,openshift/console#1862` and tested following the steps in test case. The changes look good to me

sferich888:
@ahardin-rh do we have a corresponding docs PR to link this to? That shows users how to manipulate this? 
I don't seen any docs related work trackers linked from https://issues.redhat.com/browse/CONSOLE-1523 and I want to make sure we don't miss documenting a that this change (possibly in release notes) and how it can be manipulated.

jhadvig (approved):

jhadvig:
@ahardin-rh could you please respond to https://github.com/openshift/console-operator/pull/250#issuecomment-714463647 . Thanks ! :)

PR #253, opened 2019-07-09
Bug 1724977: Invalid customization values should not make operator hotloop

jhadvig:
Based on https://github.com/openshift/cluster-authentication-operator/pull/142#issue-293463578 we should handle degraded condition more carefully, meaning update only after there are changes.

/assign @benjaminapetersen @zherman0

### DIFF
--- a/pkg/console/operator/operator.go
+++ b/pkg/console/operator/operator.go
@@ -235,8 +235,6 @@ func (c *consoleOperator) handleSync(configs configSet) error {
 		return fmt.Errorf("console is in an unknown state: %v", updatedStatus.Spec.ManagementState)
 	}
 
-	// set default conditions ok first, sync can toggle if conditions are invalid
-	c.ConditionsDefault(updatedStatus)
 	err := c.sync_v400(updatedStatus, configs)
 
 	c.HandleDegraded(updatedStatus, "SyncError", func() error {


pkg/console/operator/operator.go
```diff
@@ -217,17 +217,15 @@ func (c *consoleOperator) handleSync(configs configSet) error {
 		return fmt.Errorf("console is in an unknown state: %v", updatedStatus.Spec.ManagementState)
 	}
 
-	// set default conditions ok first, sync can toggle if conditions are invalid
-	c.ConditionsDefault(updatedStatus)
```
jhadvig: Needed to remove this one to prevent the hotlooping. Other approach would be to remove the `v1helpers.SetOperatorCondition()` that is setting the `Degraded` status type inside the `ConditionsDefault()`

pkg/console/operator/operator.go
```diff
@@ -217,17 +217,15 @@ func (c *consoleOperator) handleSync(configs configSet) error {
...
 	err := c.sync_v400(updatedStatus, configs)
 	if err != nil {
-		if !customerrors.IsSyncError(err) || !customerrors.IsCustomLogoError(err) {
```
jhadvig: Here the condition felt to be vice versa...  @benjaminapetersen can you confirm ?

pkg/console/operator/status.go
```diff
@@ -200,6 +200,14 @@ func (c *consoleOperator) ConditionDegraded(operatorConfig *operatorsv1.Console,
 	return operatorConfig
 }
 
+func (c *consoleOperator) IsConditionDegraded(operatorConfig *operatorsv1.Console, reason string) bool {
+	degradedCondition := v1helpers.FindOperatorCondition(operatorConfig.Status.Conditions, operatorsv1.OperatorStatusTypeDegraded)
+	if degradedCondition.Status == operatorsv1.ConditionTrue && degradedCondition.Reason == reason {
```
zherman0: I think there can be a null pointer error here if `FindOperatorCondition` cannot find the condition.  The function returns `nil` and then you check on `degradedCondition.Status` which does not exist.
The logs show:
```
Observed a panic: "invalid memory address or nil pointer dereference"
...
/go/src/github.com/openshift/console-operator/_output/local/go/src/github.com/openshift/console-operator/pkg/console/operator/status.go:204    <-------------
/go/src/github.com/openshift/console-operator/_output/local/go/src/github.com/openshift/console-operator/pkg/console/operator/sync_v400.go:92
```

pkg/console/operator/operator.go
benjaminapetersen: Nice catch. thinking on this one with you.

jhadvig:
@benjaminapetersen can you take a look on this one please ? :)

pkg/console/operator/operator.go
benjaminapetersen: `Degraded` is the one with most potential to jitter, since its affected by other controllers we setup in `starter.go`.

benjaminapetersen: Holding on this one as we consider the approach. I think we should just go ahead and use the same helper func method as in the PR you linked  https://github.com/openshift/cluster-authentication-operator/pull/142#issue-293463578.

So we probably want something like:

```Golang
handleDegraded(operatorConfig, "CustomLogoInvalid", ...) // Creates DegradedCustomLogoInvalid
```
The old helpers can slowly be replaced by this method then.  The `NewClusterOperatorStatusController()` will then gather all of the `Degraded` variants and determine (automatically) if we are Degraded.  And the same for Available, etc.

benjaminapetersen: We can probably do these one by one in small PRs.

benjaminapetersen: `handledDegraded()` could be a good first one to pull over & update.

jhadvig:
@benjaminapetersen tested the change manually and the hotloop is gone

benjaminapetersen:
/lgtm 

This one got small :)

PR #254, opened 2019-07-10
Add proxy support

jhadvig:
Adding console support for cluster proxy. If the Proxy CRD is set, operator will read them and set proper environment variables inside the console container.
FollowUp should be setting the proxy on the console server http client [as does in.](https://github.com/openshift/console/pull/1249)

/assign @benjaminapetersen @zherman0 

Story: https://jira.coreos.com/browse/CONSOLE-1440

### DIFF
--- a/pkg/console/operator/operator.go
+++ b/pkg/console/operator/operator.go
@@ -13,6 +13,7 @@ import (
 	corev1 "k8s.io/client-go/informers/core/v1"
 	appsv1 "k8s.io/client-go/kubernetes/typed/apps/v1"
 	coreclientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog"
 
 	// openshift
@@ -32,6 +33,7 @@ import (
 
 	// informers
 	configinformer "github.com/openshift/client-go/config/informers/externalversions"
+	configinformerv1 "github.com/openshift/client-go/config/informers/externalversions/config/v1"
 	operatorinformerv1 "github.com/openshift/client-go/operator/informers/externalversions/operator/v1"
 
 	routesinformersv1 "github.com/openshift/client-go/route/informers/externalversions/route/v1"
@@ -39,6 +41,7 @@ import (
 
 	// clients
 	configclientv1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+	configlisterv1 "github.com/openshift/client-go/config/listers/config/v1"
 	operatorclientv1 "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1"
 
 	// operator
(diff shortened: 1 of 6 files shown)


pkg/console/subresource/deployment/deployment.go
```diff
@@ -54,7 +55,7 @@ type volumeConfig struct {
 	isConfigMap bool
 }
 
-func DefaultDeployment(operatorConfig *operatorv1.Console, cm *corev1.ConfigMap, serviceCAConfigMap *corev1.ConfigMap, sec *corev1.Secret, rt *routev1.Route, canMountCustomLogo bool) *appsv1.Deployment {
+func DefaultDeployment(operatorConfig *operatorv1.Console, cm *corev1.ConfigMap, serviceCAConfigMap *corev1.ConfigMap, sec *corev1.Secret, rt *routev1.Route, proxyConfig *configv1.Proxy, canMountCustomLogo bool) *appsv1.Deployment {
```
zherman0: I think this function call has gotten long enough that we should pass in a struct.  Maybe add that to tech debt?

pkg/console/subresource/deployment/deployment.go
```diff
@@ -270,6 +272,32 @@ func consoleContainer(cr *operatorv1.Console, volConfigList []volumeConfig) core
 	}
 }
 
+func setEnvironmentVariables(proxyConfig *configv1.Proxy) []corev1.EnvVar {
+	envVars := []corev1.EnvVar{}
+	if proxyConfig == nil {
+		return envVars
+	}
+	if len(proxyConfig.Status.HTTPSProxy) != 0 {
```
zherman0: You do a check on proxyConfig if it is nil, but what about proxyConfig.Status?  Is that guaranteed to be there?

spadgett: Shouldn't be needed since it's not a pointer

pkg/console/subresource/deployment/deployment.go
jhadvig: Yeah definitely would add it as techdebt 👍

pkg/console/operator/operator.go
```diff
@@ -175,10 +187,19 @@ func (c *consoleOperator) Sync(obj metav1.Object) error {
 		return err
 	}
 
+	proxyConfig, err := c.proxyCfgLister.Get("cluster")
+	if err != nil {
+		if !errors.IsNotFound(err) {
```
spadgett: 👍 The proxy config might not be there always on upgrade, and we shouldn't error. Good catch.

I believe we can remove this in 4.3.

spadgett:
@jhadvig We'll need to make sure the proxy is not used when connecting to internal services from console, but is used for the OAuth server. See https://github.com/openshift/console/pull/1249

jhadvig:
@spadgett so the proxy for oauth server is set [in](https://github.com/openshift/oauth-proxy/pull/96).
There is also a [opened PR](https://github.com/openshift/openshift-ansible/pull/11317) to pass the proxy envVars to Kibana.

PR #255, opened 2019-07-15
Add applicationMenu to ConsoleLink CRD

benjaminapetersen:
Following these changes: https://github.com/openshift/api/pull/360

@spadgett @jhadvig @rhamilto

### DIFF
--- a/manifests/00-crd-extension-console-link.yaml
+++ b/manifests/00-crd-extension-console-link.yaml
@@ -52,5 +52,17 @@ spec:
               pattern: '^https://([\w-]+.)+[\w-]+(/[\w- ./?%&=])?$'
             location:
               type: string
-              description: Determines which dropdown menu the link is added to (HelpMenu, UserMenu)
-              pattern: ^(HelpMenu|UserMenu)$
+              description: Determines which dropdown menu the link is added to (ApplicationMenu, HelpMenu, UserMenu)
+              pattern: ^(ApplicationMenu|HelpMenu|UserMenu)$
+            applicationMenu:
+              type: object
+              description: Object that holds application menu properties if the location is set to ApplicationMenu
+              required:
+                - section
+              properties:
+                section:
+                  type: string
+                  description: The section of the application menu in which the link should appear
+                imageURL:
+                  type: string
+                  description: The URL for the icon used in front of the link in the application menu.  The URL must be an HTTPS URL or a Data URI. The image should be square and will be shown at 24x24 pixels.


manifests/00-crd-extension-console-link.yaml
```diff
@@ -54,3 +54,17 @@ spec:
...
+              required:
+                - section
+                - imageURL
+              properties:
+                section:
+                  type: string
+                  description: The section of the application menu in which the link should appear
+                imageURL:
+                  type: string
+                  description: The URL for the icon used in front of the link in the application menu.  The URL must be an HTTPS URL or a Data URI. The image should be square and will be shown at 24x24 pixels.
```
benjaminapetersen: @rhamilto does this default to anything?

benjaminapetersen: If it does, we don't need to list it as required.

spadgett: @jeff-phillips-18 

I don't think we should make it required. I'd guess it just doesn't show an icon.

manifests/00-crd-extension-console-link.yaml
```diff
@@ -54,3 +54,17 @@ spec:
...
+
```
spadgett: nit: extra trailing newline

manifests/00-crd-extension-console-link.yaml
```diff
@@ -54,3 +54,17 @@ spec:
               type: string
               description: Determines which dropdown menu the link is added to (HelpMenu, UserMenu)
               pattern: ^(HelpMenu|UserMenu)$
```
spadgett:
```suggestion
              pattern: ^(ApplicationMenu|HelpMenu|UserMenu)$
```

manifests/00-crd-extension-console-link.yaml
```diff
@@ -54,3 +54,17 @@ spec:
...
+            applicationMenu:
+              type: object
+              description: Object that holds application menu defaults if the location is set to ApplicationMenu
```
spadgett: Are these really "defaults"? Maybe

```suggestion
              description: Object that holds application menu properties if the location is set to ApplicationMenu
```

manifests/00-crd-extension-console-link.yaml
jhadvig: agree, lets dont make the `imageURL` required.

benjaminapetersen: Sounds good.

manifests/00-crd-extension-console-link.yaml
benjaminapetersen: Excellent, can't miss this one.

manifests/00-crd-extension-console-link.yaml
benjaminapetersen: Agree, much better.

manifests/00-crd-extension-console-link.yaml
```diff
@@ -53,4 +53,16 @@ spec:
...
+            applicationMenu:
+              type: object
+              description: Object that holds application menu properties if the location is set to ApplicationMenu
+              required:
+                - section
+              properties:
+                section:
+                  type: string
+                  description: The section of the application menu in which the link should appear
+                imageURL:
```
spadgett: @benjaminapetersen @jhadvig do we need to add `+optional` to the property in openshift/api? I'm not aware of all the rules....

manifests/00-crd-extension-console-link.yaml
```diff
@@ -53,4 +53,16 @@ spec:
             location:
               type: string
               description: Determines which dropdown menu the link is added to (HelpMenu, UserMenu)
```
spadgett: This comment needs to be updated as well.

manifests/00-crd-extension-console-link.yaml
benjaminapetersen: Yes, we should.  I'll do a quick PR for that.

manifests/00-crd-extension-console-link.yaml
benjaminapetersen: +1

spadgett:
@jeff-phillips-18 when this merges, we should be able to make the corresponding console change to add in the links

PR #257, opened 2019-07-18
Set statuspage-id flag

jhadvig:
Thought that we are setting the flags from the console-config. Instead we have to set them directly from the operator.

/assign @benjaminapetersen

### DIFF
--- a/pkg/console/subresource/deployment/deployment.go
+++ b/pkg/console/subresource/deployment/deployment.go
@@ -1,6 +1,7 @@
 package deployment
 
 import (
+	"fmt"
 
 	// kube
 	appsv1 "k8s.io/api/apps/v1"
@@ -220,6 +221,8 @@ func consoleVolumeMounts(vc []volumeConfig) []corev1.VolumeMount {
 
 func GetLogLevelFlag(logLevel operatorv1.LogLevel) string {
 	flag := ""
+	// Since the console-operator logging has different logging levels then the capnslog,
+	// that we use for console server(bridge) we need to map them to each other
 	switch logLevel {
 	case operatorv1.Normal:
 		flag = "--log-level=*=NOTICE"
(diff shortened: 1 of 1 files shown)


pkg/console/subresource/deployment/deployment.go
```diff
@@ -231,23 +233,38 @@ func GetLogLevelFlag(logLevel operatorv1.LogLevel) string {
...
+func getStatusPageFlag(providers operatorv1.ConsoleProviders) string {
+	if providers.Statuspage != nil && len(providers.Statuspage.PageID) != 0 {
+		return fmt.Sprintf("--statuspage-id=%s", providers.Statuspage.PageID)
+	}
+	return ""
+}
+
 func consoleContainer(cr *operatorv1.Console, volConfigList []volumeConfig, proxyConfig *configv1.Proxy) corev1.Container {
 	volumeMounts := consoleVolumeMounts(volConfigList)
+	flags := []string{
```
benjaminapetersen: Want to do something like:

```go
flags := getFlags()
flags = withLogLevelFlag(cr.Spec, flags) // returns the flags unchanged if not set
flags = withStatusPageFlag(cr.Spec, flags) // returns the flags unchanged if not set
```
I think we could eliminate the extra `if len()` blocks then.

jhadvig:
@benjaminapetersen comment addressed

benjaminapetersen:
Flake on our test:

```
=== RUN   TestBrandingFromManagedConfigMap
--- FAIL: TestBrandingFromManagedConfigMap (2.46s)
    branding_test.go:99: update data for the config map in openshift-config-managed namespace with okd
    branding_test.go:99: update data for the config map in openshift-config-managed namespace with ocp
    branding_test.go:43: could not get operator config, Operation cannot be fulfilled on consoles.operator.openshift.io "cluster": the object has been modified; please apply your changes to the latest version and try again
```
I'm looking at the PR to fix that again.

benjaminapetersen:
@jhadvig just need to `gomft` I think to fix verify.

benjaminapetersen:
hmm or not:

```
2019/07/18 16:54:44 Executing test verify
2019/07/18 16:59:11 Ran for 5m13s
2019/07/18 16:59:11 Submitted failure event to sentry (id=098143a376a1491ab2b2abf4953d7302)
error: could not run steps: step verify failed: test "verify" failed: the pod ci-op-13w8jbdd/verify failed after 4m25s (failed containers: ): ContainerFailed one or more containers exited
```

that may be a flake as well.

benjaminapetersen:
/lgtm

but I dont think tests will pass yet, something broken.

pkg/console/subresource/deployment/deployment.go
```diff
@@ -231,23 +234,37 @@ func GetLogLevelFlag(logLevel operatorv1.LogLevel) string {
...
 }
 
+func withLogLevelFlag(logLevel operatorv1.LogLevel, flags []string) []string {
+	if logLevelFlag := GetLogLevelFlag(logLevel); logLevelFlag != "" {
+		return append(flags, logLevelFlag)
+	}
+	return flags
+}
+
+func withStatusPageFlag(providers operatorv1.ConsoleProviders, flags []string) []string {
```
benjaminapetersen: Just a follow up, but I'm a bit surprised we need this to direct set the flag when we already have:

https://github.com/openshift/console/blob/c0b0efd8ac532b6af16e952cb7022efd2b6aab18/pkg/serverconfig/config.go#L123

Which should have read the config file & set the flag for us.  I think that's the direction we want all of our configs to be, really. 

Something we can look at yet.

benjaminapetersen: Also, it's weird that the console reads a yaml & then sets flags imho :)

PR #258, opened 2019-07-22
Add e2e test to test if --statuspage-id flag was set on console's pod template

jhadvig:
/assign @benjaminapetersen

### DIFF
--- a/test/e2e/providers_test.go
+++ b/test/e2e/providers_test.go
@@ -1,6 +1,7 @@
 package e2e
 
 import (
+	"fmt"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


test/e2e/providers_test.go
```diff
@@ -60,6 +61,26 @@ func TestProvidersSetStatuspageID(t *testing.T) {
 	}
 }
 
+func TestProvidersSetStatuspageIDFlag(t *testing.T) {
```
benjaminapetersen: Might be good to rename the file to `providers_statuspage_test.go`.  I bet we will get more providers.

benjaminapetersen: Only thing I see, otherwise lgtm!

benjaminapetersen: This can be done later, is not a big deal.  Lets just make sure we don't end up with a future bloated test file!

PR #248, opened 2019-06-25
Various e2e test improvements

benjaminapetersen:
Most notable:
- `Pristine()` and `MustPristine()` added to request an operator config in a clean state
- `framework.StandardSetup()` and `framework.StandardCleanup()` provided to normalize setup & cleanup across tests.  
- `WaitForSettledState()` now first checks that the current `Generation` and `ObservedGeneration` match before it assesses if the operator has settled.  If the generations mismatch, we are currently watching past changes settle, which is not helpful.
- `WaitForSettledState()` will log on tests that take unusually long times.  We should be aware so that we can optimize if possible 
- `ClientSet` no longer embeds clients as there is the possibility for collisions
- Various other small fixes
- moves `pkg/testframework` into `test/framework`, since this code is entirely for e2e tests.

/assign @jhadvig @zherman0

PR #259, opened 2019-07-25
Remove LastTransitionTime from status.conditions

benjaminapetersen:
We should not directly set the `Condition. LastTransitionTime` as this can interfere with the bookkeeping of `ClusterOperatorStatusController`.

/assign @jhadvig


benjaminapetersen:
/hold

I would prefer #248 merges before this.

PR #260, opened 2019-07-25
[WIP] Migrate from status helpers to handlers

benjaminapetersen:
Replace helpers with:
- HandleDegrade()
- HandleProgressing() 
- HandleAvailable()

Operator Conditions `Degraded`, `Progressing`, `Available` should not be set directly at this point.  We should set prefixed conditions and let `ClusterOperatorStatusController()` handle setting the derived base conditions.


pkg/console/operator/status.go
```diff
@@ -61,6 +61,50 @@ const (
...
 
+// handleDegraded can be used to set a number of Degraded statuses
+// example:
+//   co.handleDegraded(operatorConfig, "RouteStatus", err)
+//   co.handleDegraded(operatorConfig, "CustomLogoInvalid", err)
+// creates condition:
+//   RouteStatusDegraded
+//   CustomLogoInvalidDegraded
+// and uses the error as its message
+func (c *consoleOperator) HandleDegraded(operatorConfig *operatorsv1.Console, prefix string, err error) {
```
jhadvig: Not really sure if it wouldn't be better to have enum instead of plane prefix., so we have all the various conditions under control. 
Here is an [example](https://github.com/openshift/cluster-openshift-apiserver-operator/blob/3697805493449a5affd78610b20ccf55cdb82eed/pkg/operator/workloadcontroller/workload_controller.go#L35) even though they have only one...

pkg/console/operator/status.go
```diff
@@ -107,33 +151,6 @@ func (c *consoleOperator) logConditions(conditions []operatorsv1.OperatorConditi
...
-	v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorsv1.OperatorCondition{
-		Type:               operatorsv1.OperatorStatusTypeDegraded,
-		Status:             operatorsv1.ConditionFalse,
-		LastTransitionTime: metav1.Now(),
-	})
-
-	return operatorConfig
-}
-
 func (c *consoleOperator) ConditionDeploymentAvailable(operatorConfig *operatorsv1.Console, message string) *operatorsv1.Console {
```
jhadvig: Wont this change also enable to remove all of these helpers ?

pkg/console/operator/sync_v400.go
```diff
@@ -59,7 +59,7 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console,
 	if svcErr != nil {
 		msg := fmt.Sprintf("%q: %v", "service", svcErr)
 		klog.V(4).Infof("incomplete sync: %v", msg)
-		co.ConditionResourceSyncProgressing(updatedOperatorConfig, msg)
+		co.HandleProgressing(updatedOperatorConfig, "ConsoleServiceSync", svcErr)
```
jhadvig: As mentioned above.. let's talk about the enum vs. planeString

pkg/console/operator/sync_v400.go
```diff
@@ -152,13 +152,7 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console,
...
 	} else if !routesub.IsAdmitted(rt) {
 		klog.V(4).Infoln("console route is not admitted")
-		co.SetStatusCondition(
-			updatedOperatorConfig,
-			operatorv1.OperatorStatusTypeAvailable,
-			operatorv1.ConditionFalse,
-			"RouteNotAdmitted",
-			"console route is not admitted",
-		)
+		co.HandleAvailable(updatedOperatorConfig, "RouteNotAdmitted", errors.New("console route is not admitted"))
```
jhadvig: nit, lets not repeat ourselves
```suggestion
                msg := "console route is not admitted"
                klog.V(4).Infoln(msg)
		co.HandleAvailable(updatedOperatorConfig, "RouteNotAdmitted", errors.New(msg))
```

benjaminapetersen:
@jhadvig fyi

pkg/console/operator/status.go
benjaminapetersen: I'm following along the lines of this:
https://github.com/openshift/cluster-authentication-operator/pull/142/files

pkg/console/operator/status.go
benjaminapetersen: I would like to remove them all, agree.  Without any structural change anywhere else, so 1:1 replacement.

pkg/console/operator/sync_v400.go
benjaminapetersen: Sure thing, lets see:
https://github.com/openshift/cluster-authentication-operator/pull/142/files

 I don't believe any of the strings should be reused.  Each one should be specific enough for a single use in one place.

pkg/console/operator/status.go
benjaminapetersen: I don't believe any of the strings should be reused. Each one should be specific enough for a single use in one place.

jhadvig: > I don't believe any of the strings should be reused. Each one should be specific enough for a single use in one place.

sounds legit 👍

pkg/console/starter/starter.go
```diff
@@ -174,6 +176,19 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		ctx.EventRecorder,
 	)
 
+	staleConditionsController := staleconditions.NewRemoveStaleConditions(
```
benjaminapetersen: For reference, we see this elsewhere:
https://github.com/openshift/cluster-authentication-operator/pull/159/commits/17a0064a2c10bbf1e2b5926f21192ed5153cb5e2#diff-c9f95324de2319f30fff1ad2fd4e7b00R160

pkg/console/operator/status.go
benjaminapetersen: I may do this in several PRs, this one can concentrate on updating `Degraded`.  That way we can move along with others like #253

pkg/console/operator/status.go
benjaminapetersen: That way we can `grep` to find the point of failure & know for certain what is going wrong.  In theory anyway...

pkg/console/operator/sync_v400.go
```diff
@@ -125,40 +125,54 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console,
 	klog.V(4).Infof("sync loop 4.0.0 resources updated: %v", toUpdate)
 	klog.V(4).Infoln("-----------------------")
 
-	// the operand is in a transitional state if any of the above resources changed
-	// or if we have not settled on the desired number of replicas or deployment is not up to date.
-	if toUpdate {
-		co.HandleProgressing(updatedOperatorConfig, "ResourceSync", errors.New("Changes made during sync updates, additional sync expected."))
-	} else {
-		version := os.Getenv("RELEASE_VERSION")
+	co.HandleProgressing(updatedOperatorConfig, "ResourceSync", func() error {
```
benjaminapetersen: @jhadvig would def like a 2nd set of eyes on this.  Trying to break out these into individual units.

Note that `HandleAvailable()`, `HandleProgressing()`, `HandleDegraded()`, will prefix `Available`, `Progressing`, `Degraded`, in the given string `Reason`.  These will be aggregated by the `NewClusterOperatorStatusController()` to decide the final state.

We should be able to trust it instead of the nested `if` `else` `else if` we were using.

benjaminapetersen:
```
I0808 22:25:58.282511       1 status_controller.go:165] clusteroperator/console diff {"status":{"conditions":[{"lastTransitionTime":"2019-08-08T21:55:54Z","reason":"AsExpected","status":"False","type":"Degraded"},{"lastTransitionTime":"2019-08-08T21:55:54Z","message":"ConsoleRouteSyncProgressing: waiting on route host","reason":"ConsoleRouteSyncProgressingConsoleRouteSync","status":"True","type":"Progressing"},{"lastTransitionTime":"2019-08-08T21:56:25Z","reason":"MultipleConditionsMatching","status":"False","type":"Available"},{"lastTransitionTime":"2019-08-08T21:55:54Z","reason":"AsExpected","status":"True","type":"Upgradeable"}]}}
```

benjaminapetersen:
Broke this PR down into several smaller to make the changes more managable.

PR #261, opened 2019-07-29
Add ManagementStateController

benjaminapetersen:
Adds the `ManagementStateController` to handle `status` messages on the operator rather than handle these manually.  This ensures we are consistent with other operators.  Note that it does not actually handle `ManagementState` in terms of sync loop.

/assign @jhadvig

### DIFF
--- a/pkg/console/operator/status.go
+++ b/pkg/console/operator/status.go
@@ -50,9 +50,6 @@ import (
 const (
 	reasonAsExpected          = "AsExpected"
 	reasonWorkloadFailing     = "WorkloadFailing"
-	reasonUnmanaged           = "ManagementStateUnmanaged"
-	reasonRemoved             = "ManagementStateRemoved"
-	reasonInvalid             = "ManagementStateInvalid"
 	reasonSyncLoopProgressing = "SyncLoopProgressing"
 	reasonSyncError           = "SynchronizationError"
 	reasonNoPodsAvailable     = "NoPodsAvailable"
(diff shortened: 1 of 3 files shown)


benjaminapetersen:
/retest

I would expect some failures, but this was a lot of flakes.

pkg/api/api.go
```diff
@@ -7,6 +7,7 @@ const (
 
 // consts to maintain existing names of various sub-resources
 const (
+	ClusterOperatorName                 = "console"
```
jhadvig: This will most probably conflict with https://github.com/openshift/console-operator/pull/262

pkg/console/starter/starter.go
```diff
@@ -5,6 +5,8 @@ import (
 	"os"
 	"time"
 
+	"github.com/openshift/library-go/pkg/operator/management"
```
jhadvig: nit.
lets put this to the rest of the openshift libs

pkg/console/operator/status.go
```diff
@@ -235,96 +232,6 @@ func (c *consoleOperator) ConditionResourceSyncDegraded(operatorConfig *operator
...
-	})
-
-	return operatorConfig
-}
-
-// If ManagementState is invalid,
-// We don't know if the operand is available bc we exit the sync loop w/o assessing resources
-// We aren't progressing because we exit the sync loop
-// We don't know if we are degraded because we exit the sync loop
-func (c *consoleOperator) ConditionsManagementStateInvalid(operatorConfig *operatorsv1.Console) *operatorsv1.Console {
```
jhadvig: Need to take care of https://github.com/openshift/console-operator/blob/master/pkg/console/operator/operator.go#L234

pkg/console/operator/status.go
```diff
@@ -235,96 +232,6 @@ func (c *consoleOperator) ConditionResourceSyncDegraded(operatorConfig *operator
...
-		Status:             operatorsv1.ConditionFalse,
-		Reason:             reasonUnmanaged,
-		Message:            "The operator is in an unmanaged state, therefore no operator actions are degraded.",
-		LastTransitionTime: metav1.Now(),
-	})
-
-	return operatorConfig
-}
-
-func (c *consoleOperator) ConditionsManagementStateRemoved(operatorConfig *operatorsv1.Console) *operatorsv1.Console {
```
jhadvig: Need to take care of https://github.com/openshift/console-operator/blob/master/pkg/console/operator/operator.go#L228

pkg/console/operator/status.go
```diff
@@ -235,96 +232,6 @@ func (c *consoleOperator) ConditionResourceSyncDegraded(operatorConfig *operator
 	return operatorConfig
 }
 
-func (c *consoleOperator) ConditionsManagementStateUnmanaged(operatorConfig *operatorsv1.Console) *operatorsv1.Console {
```
jhadvig: need to take case of https://github.com/openshift/console-operator/blob/master/pkg/console/operator/operator.go#L221

pkg/api/api.go
benjaminapetersen: Looks like its not a prob.  No rebase tag.

pkg/console/operator/status.go
benjaminapetersen: +1 got it!

pkg/console/starter/starter.go
benjaminapetersen: +1

benjaminapetersen:
@jhadvig should be ready for another pass.

PR #269, opened 2019-08-13
Add condition handlers, implement for Degraded prefixing, add StaleConditionController

benjaminapetersen:
Implements part of #260 

Degraded is a pretty significant status in terms of our telemetry, but also infrequently used in code.  Updating first for these reasons.

/assign @jhadvig 
@spadgett

PR #271, opened 2019-08-15
Update Progressing condition to use handler

benjaminapetersen:
Depends on #269 & will be rebased once it is merged.
Re [bugzilla 1740387](https://bugzilla.redhat.com/show_bug.cgi?id=1740387) - partial fix.

/assign @jhadvig 

@spadgett fyi this would update many of the ambiguous conditions we are seeing.

### DIFF
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -47,76 +47,69 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console, 
 	toUpdate := false
 
 	rt, rtChanged, rtErr := co.SyncRoute(set.Operator)
+	toUpdate = toUpdate || rtChanged
+	co.HandleProgressing(updatedOperatorConfig, "ConsoleRouteSync", rtErr)
 	if rtErr != nil {
-		msg := fmt.Sprintf("%v: %s", "route", rtErr)
-		klog.V(4).Infof("incomplete sync: %v", msg)
-		co.ConditionResourceSyncProgressing(updatedOperatorConfig, msg)
+		klog.V(4).Infof("incomplete sync: %v", fmt.Sprintf("%v: %s", "route", rtErr))
 		return rtErr
 	}
-	toUpdate = toUpdate || rtChanged
 
 	svc, svcChanged, svcErr := co.SyncService(set.Operator)
+	toUpdate = toUpdate || svcChanged
+	co.HandleProgressing(updatedOperatorConfig, "ConsoleServiceSync", svcErr)
 	if svcErr != nil {
-		msg := fmt.Sprintf("%q: %v", "service", svcErr)
-		klog.V(4).Infof("incomplete sync: %v", msg)
-		co.ConditionResourceSyncProgressing(updatedOperatorConfig, msg)
+		klog.V(4).Infof("incomplete sync: %v", fmt.Sprintf("%q: %v", "service", svcErr))
 		return svcErr
 	}
-	toUpdate = toUpdate || svcChanged
 
 	cm, cmChanged, cmErr := co.SyncConfigMap(set.Operator, set.Console, set.Infrastructure, rt)
+	toUpdate = toUpdate || cmChanged
+	co.HandleProgressing(updatedOperatorConfig, "ConsoleConfigMapSync", cmErr)
 	if cmErr != nil {
-		msg := fmt.Sprintf("%q: %v", "configmap", cmErr)
-		klog.V(4).Infof("incomplete sync: %v", msg)
-		co.ConditionResourceSyncProgressing(updatedOperatorConfig, msg)
+		klog.V(4).Infof("incomplete sync: %v", fmt.Sprintf("%q: %v", "configmap", cmErr))
 		return cmErr
 	}
-	toUpdate = toUpdate || cmChanged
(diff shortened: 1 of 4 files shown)


pkg/console/operator/sync_v400.go
```diff
@@ -47,76 +47,69 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console,
 	toUpdate := false
 
 	rt, rtChanged, rtErr := co.SyncRoute(set.Operator)
+	toUpdate = toUpdate || rtChanged
+	co.HandleProgressing(updatedOperatorConfig, "ConsoleRouteSync", rtErr)
```
benjaminapetersen: `HandleProgressing()` can't be hidden behind the `if` as we need it to toggle `true||false` based on the error so we don't end up in a permanent `Progressing` situation.  Same as `Degraded`.

Fortunately our e2e test should ensure we don't get in this state.  HOWEVER, if we ever remove a condition, this could get interesting between versions during an upgrade.

pkg/console/starter/starter.go
```diff
@@ -176,6 +177,20 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		ctx.EventRecorder,
 	)
 
+	staleConditionsController := staleconditions.NewRemoveStaleConditions(
+		[]string{
+			// in 4.1.0 we directly set this set of conditions. We should no longer do this.
+			// in 4.3.0+ we can remove this
+			"Degraded",
+			// in follow-up PRs, we should stop using the rest directly:
+			"Progressing",
```
benjaminapetersen: Adding the raw `Progressing` here.
If we ever remove any of the new conditions we added, we need to update this controller as well so an operator doesn't get stuck in an upgrade situation.

benjaminapetersen:
Tests all passed cleanly.

benjaminapetersen:
Rebased on master.

pkg/console/operator/sync_v400.go
```diff
@@ -125,21 +118,19 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console,
...
-			co.ConditionResourceSyncProgressing(updatedOperatorConfig, fmt.Sprintf("Working toward version %s", version))
-		} else {
-			if co.versionGetter.GetVersions()["operator"] != version {
-				co.versionGetter.SetVersion("operator", version)
-			}
-			co.ConditionResourceSyncNotProgressing(updatedOperatorConfig)
+			return errors.New(fmt.Sprintf("Working toward version %s", version))
 		}
-	}
+		if co.versionGetter.GetVersions()["operator"] != version {
```
benjaminapetersen: I'm not 100% on this being hidden within the `func() {}()` of `HandleProgressing()`, but it works & is consistent with the previous code.  Its just a tad "dirty" as the callback func shouldn't really have side effects, it should simply return an error, or not.

pkg/console/operator/sync_v400.go
```diff
@@ -125,21 +118,19 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console,
...
 	klog.V(4).Infoln("-----------------------")
 
-	// the operand is in a transitional state if any of the above resources changed
-	// or if we have not settled on the desired number of replicas or deployment is not up to date.
-	if toUpdate {
-		co.ConditionResourceSyncProgressing(updatedOperatorConfig, "Changes made during sync updates, additional sync expected.")
-	} else {
+	co.HandleProgressing(updatedOperatorConfig, "ResourceSyncUpdatesInProgress", func() error {
+		if toUpdate {
+			return errors.New("Changes made during sync updates, additional sync expected.")
```
jhadvig: Wont returning an error set the `Progressing` condition to `False` ?
Thought that if there needs to be a an additional sync loop, then the `Progressing` should be `True`

pkg/console/operator/sync_v400.go
```diff
@@ -125,21 +118,19 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console,
...
+		}
 		version := os.Getenv("RELEASE_VERSION")
 		if !deploymentsub.IsAvailableAndUpdated(actualDeployment) {
-			co.ConditionResourceSyncProgressing(updatedOperatorConfig, fmt.Sprintf("Working toward version %s", version))
-		} else {
-			if co.versionGetter.GetVersions()["operator"] != version {
-				co.versionGetter.SetVersion("operator", version)
-			}
-			co.ConditionResourceSyncNotProgressing(updatedOperatorConfig)
+			return errors.New(fmt.Sprintf("Working toward version %s", version))
```
jhadvig: Same as above

pkg/console/operator/sync_v400.go
benjaminapetersen: Returning an `error` will set `Progressing: true` with the error message.

BUT I don't think I did the inversion on `Available!`

Leaving a note on that PR: https://github.com/openshift/console-operator/pull/272/files#r314387565

pkg/console/operator/sync_v400.go
benjaminapetersen: I think `Available` actually needs this inversion instead.  Lemme look quick.

PR #272, opened 2019-08-15
Update Available condition to use handlers

benjaminapetersen:
Depends on #269 & #271 & will be rebased once it is merged.
Re [bugzilla 1740387](https://bugzilla.redhat.com/show_bug.cgi?id=1740387) - partial fix.

WIP until rebase.
@jhadvig I think #253 can be rebased on this change now as we no longer need the `ConditionsDefault()` as a dependency for the rest of the statuses.

/assign @jhadvig 
@spadgett fyi

This is should handle the rest of the status transition.

### DIFF
--- a/pkg/console/operator/status.go
+++ b/pkg/console/operator/status.go
@@ -88,7 +88,7 @@ func handleCondition(operatorConfig *operatorsv1.Console, conditionType string, 
 		v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions,
 			operatorsv1.OperatorCondition{
 				Type:    conditionType,
-				Status:  operatorsv1.ConditionTrue,
+				Status:  setConditionValue(conditionType, err),
 				Reason:  reason,
 				Message: err.Error(),
 			})
@@ -97,8 +97,22 @@ func handleCondition(operatorConfig *operatorsv1.Console, conditionType string, 
 	v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions,
 		operatorsv1.OperatorCondition{
 			Type:   conditionType,
-			Status: operatorsv1.ConditionFalse,
+			Status: setConditionValue(conditionType, err),
 		})
+}
+
+// Available is an inversion of the other conditions
+func setConditionValue(conditionType string, err error) operatorsv1.ConditionStatus {
+	if strings.HasSuffix(conditionType, operatorsv1.OperatorStatusTypeAvailable) {
+		if err != nil {
+			return operatorsv1.ConditionFalse
+		}
+		return operatorsv1.ConditionTrue
+	}
+	if err != nil {
+		return operatorsv1.ConditionTrue
+	}
+	return operatorsv1.ConditionFalse
 }
 
 func IsDegraded(operatorConfig *operatorsv1.Authentication) bool {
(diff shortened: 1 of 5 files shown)


pkg/console/operator/status.go
```diff
@@ -84,6 +84,9 @@ func (c *consoleOperator) HandleAvailable(operatorConfig *operatorsv1.Console, p
 
 // internal func for handling conditions
 func handleCondition(operatorConfig *operatorsv1.Console, conditionType string, reason string, err error) {
+	if conditionType == "" {
```
benjaminapetersen: I think a `panic` here is probably correct, we shouldn't allow empty string.

pkg/console/operator/status.go
```diff
@@ -157,86 +160,11 @@ func (c *consoleOperator) logConditions(conditions []operatorsv1.OperatorConditi
...
-		Message: "As expected",
-	})
-	v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorsv1.OperatorCondition{
-		Type:    operatorsv1.OperatorStatusTypeProgressing,
-		Status:  operatorsv1.ConditionFalse,
-		Reason:  reasonAsExpected,
-		Message: "As expected",
-	})
-
+	c.HandleAvailable(operatorConfig, "Default", nil)
```
benjaminapetersen: `HandleAvailable` when it receives `nil`, should set `Available:true`.  I think this will set it to false, it needs an inversion of the `Degraded` & `Progressing` logic.

pkg/console/operator/status.go
jhadvig: Wont this cause program to crash unless `recover()` func is called ?

pkg/console/operator/sync_v400.go
```diff
@@ -125,45 +118,43 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console,
...
+	co.HandleAvailable(updatedOperatorConfig, "DeploymentIsReady", func() error {
+		if !deploymentsub.IsReady(actualDeployment) {
+			msg := fmt.Sprintf("%v pods available for console deployment", actualDeployment.Status.ReadyReplicas)
+			klog.V(4).Infoln(msg)
+			return errors.New(msg)
+		}
+		return nil
+	}())
+	co.HandleAvailable(updatedOperatorConfig, "DeploymentIsUpdated", func() error {
+		if deploymentsub.IsReadyAndUpdated(actualDeployment) {
```
jhadvig:
```suggestion
		if !deploymentsub.IsReadyAndUpdated(actualDeployment) {
```

pkg/console/operator/sync_v400.go
```diff
@@ -125,45 +118,43 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console,
...
+			msg := fmt.Sprintf("%v replicas ready at version %s", actualDeployment.Status.ReadyReplicas, os.Getenv("RELEASE_VERSION"))
+			return errors.New(msg)
```
jhadvig: would add also:
```
klog.V(4).Infoln(msg)
```

pkg/console/operator/status.go
benjaminapetersen: It's really for development.  We shouldn't do this, so it should crash.  Not sure if there is a better way to ensure this?

pkg/console/operator/sync_v400.go
benjaminapetersen: +1

pkg/console/operator/sync_v400.go
benjaminapetersen: +1
Def need to push these into a helper.

pkg/console/operator/status.go
```diff
@@ -79,11 +79,33 @@ func (c *consoleOperator) HandleProgressing(operatorConfig *operatorsv1.Console,
 
 func (c *consoleOperator) HandleAvailable(operatorConfig *operatorsv1.Console, prefix string, err error) {
 	conditionType := prefix + operatorsv1.OperatorStatusTypeAvailable
-	handleCondition(operatorConfig, conditionType, prefix, err)
+	if conditionType == "" {
```
benjaminapetersen: Hate to duplicate all of this, but such is life.

pkg/console/operator/status.go
```diff
@@ -79,11 +79,33 @@ func (c *consoleOperator) HandleProgressing(operatorConfig *operatorsv1.Console,
...
+		panic("Condition type cannot be empty string")
+	}
+	if err != nil {
+		v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions,
+			operatorsv1.OperatorCondition{
+				Type: conditionType,
+				// Available is inversion of Progressing,Degraded
+				Status:  operatorsv1.ConditionFalse,
```
jhadvig: maybe

```
...
Status:  setConditionValue(conditionType,err),
...

func setConditionValue(conditionType string, err error) (operatorsv1.ConditionStatus) {
  if strings.HasSuffix(conditionType, operatorsv1.OperatorStatusTypeAvailable) {
    if err != nil {
      return operatorsv1.ConditionFalse
    }
    return operatorsv1.ConditionTrue
  }
  if err != nil {
    return operatorsv1.ConditionTrue
  }
  return operatorsv1.ConditionFalse
}
```

benjaminapetersen: yeah lets do this.

pkg/console/operator/status.go
benjaminapetersen: eh, i will eliminate.  its obvious "dont do that"

pkg/console/operator/status.go
benjaminapetersen: ill eliminate.

benjaminapetersen:
/retest 

bunch of flakes.

PR #273, opened 2019-08-16
[WIP] Remove SyncError from Degraded reporting

benjaminapetersen:
/assign @jhadvig @spadgett 

items:
- Small bug fixes around `Degraded`
- Pointing out that our handling of `Degraded` is not great.  We have only 1 Degraded at this point.  Need to clarify this and ensure we are capturing what we need.

### DIFF
--- a/pkg/console/operator/operator.go
+++ b/pkg/console/operator/operator.go
@@ -239,15 +239,9 @@ func (c *consoleOperator) handleSync(configs configSet) error {
 	c.ConditionsDefault(updatedStatus)
 	err := c.sync_v400(updatedStatus, configs)
 
-	c.HandleDegraded(updatedStatus, "SyncError", func() error {
-		if !customerrors.IsSyncError(err) {
-			return err
-		}
-		return nil
-	}())
-
+	// anything wrong with custom logo should be reported as degraded
 	c.HandleDegraded(updatedStatus, "CustomLogo", func() error {
-		if !customerrors.IsCustomLogoError(err) {
+		if customerrors.IsCustomLogoError(err) {
 			return err
 		}
 		return nil


pkg/console/operator/operator.go
```diff
@@ -239,15 +239,9 @@ func (c *consoleOperator) handleSync(configs configSet) error {
...
-		if !customerrors.IsSyncError(err) {
-			return err
-		}
-		return nil
-	}())
-
+	// anything wrong with custom logo should be reported as degraded
 	c.HandleDegraded(updatedStatus, "CustomLogo", func() error {
-		if !customerrors.IsCustomLogoError(err) {
+		if customerrors.IsCustomLogoError(err) {
```
benjaminapetersen: Logic flip from the restructuring of last PR.  If error && is correct type of error, return it.

pkg/console/operator/operator.go
```diff
@@ -239,15 +239,9 @@ func (c *consoleOperator) handleSync(configs configSet) error {
 	c.ConditionsDefault(updatedStatus)
 	err := c.sync_v400(updatedStatus, configs)
 
-	c.HandleDegraded(updatedStatus, "SyncError", func() error {
```
benjaminapetersen: `SyncErrors` are not real errors, from the beginning, we have used this as a way to just kick the sync loop.  They really shouldn't be reported at all.

The problem here now is that `CustomLogo` is the ONLY degraded status we report.  Everything else is `Progressing`. 

We probably need to reasses what we assign to `Progressing` and decide if that is actually `Degraded`.

benjaminapetersen: We used to say `if !IsSyncError && !IsCustomLogoError` then we are `Degraded` with the error as a `message`.  The problem is that we report too many `GenericSyncErrors` and they don't mean anything (since the message doesn't make it into Telemetry from Monitoring).  Not sure at the moment how to fix this.  Its not meaningful to track an error if we don't know what the error is.  BUT we shouldn't ignore them and let it look like everything is fine.

benjaminapetersen:
Closing this as #274 will handle it better.

PR #280, opened 2019-08-29
Bug 1746925: Custom logo file cant be found when SVG

jhadvig:
When creating a SVG type custom logo, the created configMap will contain the data in the `data` field not in the `binaryData` as in case of PNG, JPG,.. which will cause the console pod to CrashLoop.
Since we [recommend the use of the SVG format](https://github.com/openshift/api/blob/master/operator/v1/types_console.go#L79), we need to check for the data in the appropriate field.

/assign @benjaminapetersen 
@spadgett FYI

Will add additional e2e for this case.


benjaminapetersen:
Good catch, I always used a .png when testing.

jhadvig:
@spadgett @benjaminapetersen comment addressed. Thanks !

PR #284, opened 2019-09-03
[WIP] bug 1748280: custom logo crashloop

benjaminapetersen:
Fix ValidateCustomLogo() to ensure image from configmap is not mounted unless it is properly sync'd.

### DIFF
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -457,12 +457,30 @@ func (c *consoleOperator) ValidateCustomLogo(operatorConfig *operatorsv1.Console
 		klog.V(4).Infoln("no custom logo configured")
 		return false, "", nil
 	}
-	logoConfigMap, err := c.configMapClient.ConfigMaps(api.OpenShiftConfigNamespace).Get(logoConfigMapName, metav1.GetOptions{})
+
+	// original user-defined logo in openshift-config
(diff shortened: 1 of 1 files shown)


pkg/console/operator/sync_v400.go
```diff
@@ -457,12 +457,30 @@ func (c *consoleOperator) ValidateCustomLogo(operatorConfig *operatorsv1.Console
...
+		msg := fmt.Sprintf("source custom logo file %v in openshift-config not found", logoConfigMapName)
+		klog.V(4).Infof(msg)
+		return false, "FailedGetSource", customerrors.NewCustomLogoError(msg)
+	}
+	if err != nil {
+		return false, "SourceError", customerrors.NewCustomLogoError(fmt.Sprintf("custom logo: %v\n", err))
+	}
+
+	// sync'd logo configmap into openshift-console namespace
+	logoConfigMap, err := c.configMapClient.ConfigMaps(api.OpenShiftConsoleNamespace).Get(api.OpenShiftCustomLogoConfigMapName, metav1.GetOptions{})
```
benjaminapetersen: This is what we did not do previously.  We need the source configmap, but also the destination confimap validated.

jhadvig:
We can probably close this PR in favour of https://github.com/openshift/library-go/pull/520
Tracked down the root cause to the library-go where the sync is not counting with the `binaryData` field.
Once the library-go fix is merged will update the dependencies, which should solve the issue.

PR #286, opened 2019-09-04
Bug 1748280: Pod CrashLoop cause by library-go + Bump

jhadvig:
/assign @benjaminapetersen 

The issue was traced down to the library-go where the sync is not couting with the `binaryData` field. It was fixed in https://github.com/openshift/library-go/pull/520
Bump is carrying the fix.

Tested the fix manually and its solving the issue.


spadgett:
Looks like some files in vendor are failing gofmt. Can we exclude those directories?

jhadvig:
@spadgett fixed the issue, needed to clean the cache. Should be fine now

spadgett:
> The issue was traced down to the library-go where the sync is not couting with the `binaryData` field. It was fixed in [openshift/library-go#520](https://github.com/openshift/library-go/pull/520)

Nice job tracking this down!

PR #289, opened 2019-09-04
Make sure configMap with custom logo is created before mount

jhadvig:
... before we forget 😄 

/assign @benjaminapetersen

### DIFF
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -412,9 +412,9 @@ func (co *consoleOperator) SyncRoute(operatorConfig *operatorv1.Console) (consol
 
 func (c *consoleOperator) SyncCustomLogoConfigMap(operatorConfig *operatorsv1.Console) (okToMount bool, reason string, err error) {
 	// validate first, to avoid a broken volume mount & a crashlooping console
-	okToMount, reason, err = c.ValidateCustomLogo(operatorConfig)
+	okToSync, okToMount, reason, err := c.ValidateCustomLogo(operatorConfig)
 
-	if okToMount || configmapsub.IsRemoved(operatorConfig) {
+	if okToSync || okToMount || configmapsub.IsRemoved(operatorConfig) {
 		if err := c.UpdateCustomLogoSyncSource(operatorConfig); err != nil {
 			return false, "FailedSyncSource", customerrors.NewCustomLogoError("custom logo sync source update error")
 		}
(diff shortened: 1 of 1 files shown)


pkg/console/operator/sync_v400.go
```diff
@@ -458,10 +458,14 @@ func (c *consoleOperator) ValidateCustomLogo(operatorConfig *operatorsv1.Console
 		return false, "", nil
 	}
 	logoConfigMap, err := c.configMapClient.ConfigMaps(api.OpenShiftConfigNamespace).Get(logoConfigMapName, metav1.GetOptions{})
-	// If we 404, the logo file may not have been created yet.
+	// We should not mount until the configMap with the custom logo is created by user
```
benjaminapetersen: So, we can't mount until the `Destination` configmap exists.  We need both!  Source to even test, but destination is the one we are mounting.

```golang
// source exists, great, we can continue, but not yet mount. 
c.configMapClient.ConfigMaps(api.OpenShiftConfigNamespace).Get(logoConfigMapName, metav1.GetOptions{})

// destination CM in openshift-console namespace, great, now we can mount!
logoConfigMap, err := c.configMapClient.ConfigMaps(api.OpenShiftConsoleNamespace).Get(api.OpenShiftCustomLogoConfigMapName, metav1.GetOptions{})
```

spadgett:
Will this prevent the operator from reporting degraded if the user never creates the config map?

pkg/console/operator/sync_v400.go
benjaminapetersen: Ah, per this: 
https://github.com/openshift/console-operator/pull/284/files#diff-283b3e79dd03261002107f949ad39c38R480

jhadvig:
@spadgett yes it should, since we are returning a custom sync error.

pkg/console/operator/sync_v400.go
jhadvig: so the problem is when the user defines the custom logo and also creates the ConfigMap in the `openshift-config` NS for the first time. With the check of the synced CM in the `openshift-console` NS, it wont get ever synced, since we will return `okToMount` with `false`. Will need to updated the logic in https://github.com/openshift/console-operator/blob/master/pkg/console/operator/sync_v400.go#L417 and add that case.

benjaminapetersen: Perhaps we need an `okToSync` bool?

pkg/console/operator/sync_v400.go
```diff
@@ -444,24 +444,40 @@ func (c *consoleOperator) UpdateCustomLogoSyncSource(operatorConfig *operatorsv1
...
+		return false, false, "SourceError", customerrors.NewCustomLogoError(fmt.Sprintf("source custom logo error: %v\n", err))
+	}
+
+	logoConfigMap, err := c.configMapClient.ConfigMaps(api.OpenShiftConsoleNamespace).Get(api.OpenShiftCustomLogoConfigMapName, metav1.GetOptions{})
+	// We should not mount until the 'custom-logo' configMap is not synced in the
+	// 'openshift-console' namespace. If the configMap is missing we need to sync it.
+	if apierrors.IsNotFound(err) {
+		msg := fmt.Sprintf("destination custom logo file %v in openshift-console not found", api.OpenShiftCustomLogoConfigMapName)
+		klog.V(4).Infof(msg)
+		return true, false, "FailedGetDestinationConfigMap", customerrors.NewCustomLogoError(msg)
```
benjaminapetersen: Nit, you use `FailedGetDestination` as the `Type` will be `CustomLogoSync`. so:

`CustomLogoSync` -> `FailedGetDestination` will be what is seen in monitoring.

pkg/console/operator/sync_v400.go
```diff
@@ -444,24 +444,40 @@ func (c *consoleOperator) UpdateCustomLogoSyncSource(operatorConfig *operatorsv1
...
 		klog.V(4).Infoln("no custom logo configured")
-		return false, "", nil
+		return false, false, "", nil
+	}
+	_, err = c.configMapClient.ConfigMaps(api.OpenShiftConfigNamespace).Get(logoConfigMapName, metav1.GetOptions{})
+	// We should not mount until the configMap with the custom logo is created by user
+	if apierrors.IsNotFound(err) {
+		msg := fmt.Sprintf("source custom logo file %v in openshift-config not found", logoConfigMapName)
+		klog.V(4).Infof(msg)
+		return false, false, "FailedGetSourceConfigMap", customerrors.NewCustomLogoError(msg)
```
benjaminapetersen: Nit, you use `FailedGetSource` as the `Type` will be `CustomLogoSync`. so:

`CustomLogoSync` -> `FailedGetSource` will be what is seen in monitoring.

pkg/console/operator/sync_v400.go
```diff
@@ -444,24 +444,40 @@ func (c *consoleOperator) UpdateCustomLogoSyncSource(operatorConfig *operatorsv1
...
+	}
+	if err != nil {
+		return false, false, "SourceError", customerrors.NewCustomLogoError(fmt.Sprintf("source custom logo error: %v\n", err))
+	}
+
+	logoConfigMap, err := c.configMapClient.ConfigMaps(api.OpenShiftConsoleNamespace).Get(api.OpenShiftCustomLogoConfigMapName, metav1.GetOptions{})
+	// We should not mount until the 'custom-logo' configMap is not synced in the
```
benjaminapetersen: nit: remove 2nd`not` -`We should not mount until the 'custom-logo' configMap is synced in the...`

pkg/console/operator/sync_v400.go
```diff
@@ -444,37 +444,54 @@ func (c *consoleOperator) UpdateCustomLogoSyncSource(operatorConfig *operatorsv1
...
+	}
+	if err != nil {
+		return false, false, "SourceError", customerrors.NewCustomLogoError(fmt.Sprintf("source custom logo error: %v\n", err))
+	}
+
+	logoConfigMap, err := c.configMapClient.ConfigMaps(api.OpenShiftConsoleNamespace).Get(api.OpenShiftCustomLogoConfigMapName, metav1.GetOptions{})
+	// We should not mount until the 'custom-logo' configMap is synced in the
+	// 'openshift-console' namespace. If the configMap is missing we need to sync it.
+	if apierrors.IsNotFound(err) {
+		msg := fmt.Sprintf("destination custom logo file %v in openshift-console not found", api.OpenShiftCustomLogoConfigMapName)
```
benjaminapetersen: nit: i suppose we should use the `const` `TargetNamespace` instead of `openshift-console`.
I bet there is a bit of that scattered about :)

benjaminapetersen:
/lgtm

Thx @jhadvig!

PR #296, opened 2019-09-17
Proxy e2e test

jhadvig:
Test scenario: 
- e2e test for setting `NoProxy` & `HTTPProxy` & `HTTPSProxy` to the Proxy CR and checking the new rolledout console deployment pod for corresponding environment variables (`NO_PROXY` & `HTTP_PROXY` & `HTTPS_PROXY`) 

/assign @benjaminapetersen

### DIFF
--- a/test/e2e/proxy_test.go
+++ b/test/e2e/proxy_test.go
@@ -1,0 +1,86 @@
+package e2e
+
+import (
+	"testing"
+	"time"
+
+	apps1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	configv1 "github.com/openshift/api/config/v1"
+	operatorsv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/console-operator/test/e2e/framework"
+)
+
+func setupProxyTest(t *testing.T) (*framework.ClientSet, *operatorsv1.Console) {
+	clientSet, operatorConfig := framework.StandardSetup(t)
+	return clientSet, operatorConfig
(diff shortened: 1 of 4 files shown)


test/e2e/proxy_test.go
```diff
@@ -0,0 +1,67 @@
...
+		},
+		"clusterVars": {
+			{Name: "NO_PROXY", Value: clusterProxyConfig.NoProxy, ValueFrom: nil},
+			{Name: "HTTP_PROXY", Value: clusterProxyConfig.HTTPProxy, ValueFrom: nil},
+			{Name: "HTTPS_PROXY", Value: clusterProxyConfig.HTTPSProxy, ValueFrom: nil},
+		},
+	}
+
+	consoleDeployment, err := framework.GetConsoleDeployment(client)
+	if err != nil {
```
benjaminapetersen: So can this run without the Proxy CR?  This is a sufficient test?

benjaminapetersen: I could see on our end that passing the envs to the deployment is sufficient.

But do we then need a test in `console` validating that it responds through the proxy?

jhadvig: We probably should. Lets talk about that :)

jhadvig:
@benjaminapetersen updated the PR, cause I noticed it was flaking. Should be fixed now. PTAL

jhadvig:
@benjaminapetersen thats the flake I've seen locally every N-th run of the test.. but with changes now I was not able to reproduce...investigating

benjaminapetersen:
/lgtm

I think this does a good job proving that we do what we are supposed to do when given a proxy config.  

Now, I'm still curious about a proxy test that validates that the cluster works and the console/components is/are responsive through the proxy.  Was that still something being looked into?

PR #298, opened 2019-09-20
Update operator for project link API changes

jhadvig:
Adding the `NamespaceDashboard` location

Story: https://jira.coreos.com/browse/CONSOLE-1710

/assign @spadgett 

@rhamilto FYI

### DIFF
--- a/manifests/00-crd-extension-console-link.yaml
+++ b/manifests/00-crd-extension-console-link.yaml
@@ -52,13 +52,13 @@ spec:
               pattern: '^https://([\w-]+.)+[\w-]+(/[\w- ./?%&=])?$'
             location:
               type: string
-              description: Determines which dropdown menu the link is added to (ApplicationMenu, HelpMenu, UserMenu)
-              pattern: ^(ApplicationMenu|HelpMenu|UserMenu)$
+              description: Determines which dropdown menu the link is added to (ApplicationMenu, HelpMenu, UserMenu, NamespaceDashboard)
+              pattern: ^(ApplicationMenu|HelpMenu|UserMenu|NamespaceDashboard)$
             applicationMenu:
               type: object
               description: Object that holds application menu properties if the location is set to ApplicationMenu
               required:
-                - section
+              - section
               properties:
                 section:
                   type: string
@@ -66,3 +66,14 @@ spec:
                 imageURL:
                   type: string
                   description: The URL for the icon used in front of the link in the application menu.  The URL must be an HTTPS URL or a Data URI. The image should be square and will be shown at 24x24 pixels.
+            namespaceDashboard:
+              type: object
+              description: Object that holds information about namespaces in which the dashboard link should appear, and it is applicable only when location is set to NamespaceDashboard. If not specified, the link will appear in all namespaces.
+              required:
+              - namespaces
+              properties:
+                namespaces:
+                  type: array
+                  description: Namespaces is an array of namespace names in which the dashboard link should appear.
+                  items:
+                    type: string


spadgett:
thanks @jhadvig !

/hold
for 4.3

manifests/00-crd-extension-console-link.yaml
```diff
@@ -66,3 +66,12 @@ spec:
...
                   type: string
                   description: The URL for the icon used in front of the link in the application menu.  The URL must be an HTTPS URL or a Data URI. The image should be square and will be shown at 24x24 pixels.
+            namespaceDashboard:
+              type: object
+              description: Object that holds information about namespaces in which the dashboard link should appear, and it is applicable only when location is set to NamespaceDashboard. If not specified, the link will appear in all namespaces.
+              required:
+                - namespaces
+              properties:
+                namespaces:
+                  type: array
```
spadgett: I believe need can use `items` to specify it's an array of string.

https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#specifying-a-structural-schema

spadgett:
```suggestion
                  type: array
                  items:
                    type: string
```

manifests/00-crd-extension-console-link.yaml
```diff
@@ -52,8 +52,8 @@ spec:
               pattern: '^https://([\w-]+.)+[\w-]+(/[\w- ./?%&=])?$'
             location:
               type: string
-              description: Determines which dropdown menu the link is added to (ApplicationMenu, HelpMenu, UserMenu)
-              pattern: ^(ApplicationMenu|HelpMenu|UserMenu)$
+              description: Determines which dropdown menu the link is added to (ApplicationMenu, HelpMenu, UserMenu, NamespaceDashboard)
```
rhamilto: As mentioned in planning yesterday, do we want to reconsider the location name given we intend to add these links to project and namespace kebab dropdowns, too?

benjaminapetersen: `NamespaceDashboard`... does sound vague to me.  I'm not familiar with this, but I tend to think of an entire page as a dashboard.

spadgett: We're adding it to the existing namespace dashboard page. I think it's consistent with names for other locations. Although to @rhamilto's point, if it's also in the actions, maybe we need a more generic name

jhadvig: @spadgett in that case, more generic name would make sense. 
@rhamilto maybe: 
- `NamespaceLink`
- `NamespaceHelper`
- `NamespaceActionLink`
- `NamespaceDashboardAndAction`   😄
- `NamespaceCommonLink`
...
..
.

rhamilto: `NamespaceLauncher`?

jhadvig: @rhamilto `NamespaceLauncher` doesn't sound related to the intend this specific console link should be doing.
Most generic is `NamespaceLink` or maybe `NamespaceSharedLink`.

@spadgett thoughts ?

spadgett: My vote would be for either `location: Namspace` or `location: NamspaceDashboard`. The other locations don't have `Link` on the end, so we should be consistent in our naming.

jhadvig: Then I would go with the `NamspaceDashboard` as well

manifests/00-crd-extension-console-link.yaml
```diff
@@ -66,3 +66,14 @@ spec:
                 imageURL:
                   type: string
                   description: The URL for the icon used in front of the link in the application menu.  The URL must be an HTTPS URL or a Data URI. The image should be square and will be shown at 24x24 pixels.
+            namespaceDashboard:
+              type: object
+              description: Object that holds information about namespaces in which the dashboard link should appear, and it is applicable only when location is set to NamespaceDashboard. If not specified, the link will appear in all namespaces.
+              required:
+                - namespaces
```
benjaminapetersen: Tiny `nit`, our other `CRD`s do list spacing like:

```
required:
- text   # no indent required
- hrefTemplate # no ident required
```
yay yamls :)

jhadvig:
@spadgett @benjaminapetersen comment addressed. PTAL

PR #301, opened 2019-09-25
Add URL for downloading CLI to the console-config configmap

jhadvig:
PR for adding the URL for downloading CLI to the `console-config` configmap, in `clusterInfo.cliDownloadURL` field.

Screen Shot 2019-09-25 at 21 27 02

Story: https://jira.coreos.com/browse/CONSOLE-1423

/assign @benjaminapetersen

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -1,6 +1,13 @@
 package api
 
 const (
+	CliArch            = "amd64"
+	LinuxPlatformKey   = "linux"
+	MacPlatformKey     = "mac"
+	WindowsPlatformKey = "windows"
+)
+
+const (
 	TargetNamespace    = "openshift-console"
 	ConfigResourceName = "cluster"
 )
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -1,13 +1,6 @@
 package api
 
 const (
-	CliArch            = "amd64"
-	LinuxPlatformKey   = "linux"
-	MacPlatformKey     = "mac"
-	WindowsPlatformKey = "windows"
-)
-
-const (
 	TargetNamespace    = "openshift-console"
 	ConfigResourceName = "cluster"
 )
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -316,7 +316,12 @@ func (co *consoleOperator) SyncConfigMap(
 		return nil, false, "FailedManagedConfig", mcErr
 	}
 
-	defaultConfigmap, _, err := configmapsub.DefaultConfigMap(operatorConfig, consoleConfig, managedConfig, infrastructureConfig, rt)
+	downloadsRoute, err := co.routeClient.Routes(api.TargetNamespace).Get("downloads", metav1.GetOptions{})
+	if err != nil {
+		return nil, false, "FailedGetDownloadsRoute", err
+	}
+
+	defaultConfigmap, _, err := configmapsub.DefaultConfigMap(operatorConfig, consoleConfig, managedConfig, infrastructureConfig, rt, downloadsRoute)
 	if err != nil {
 		return nil, false, "FailedConsoleConfigBuilder", err
 	}
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -309,7 +309,7 @@ func (co *consoleOperator) SyncConfigMap(
 	operatorConfig *operatorv1.Console,
 	consoleConfig *configv1.Console,
 	infrastructureConfig *configv1.Infrastructure,
-	rt *routev1.Route) (consoleConfigMap *corev1.ConfigMap, changed bool, reason string, err error) {
+	consoleRoute *routev1.Route) (consoleConfigMap *corev1.ConfigMap, changed bool, reason string, err error) {
 
 	managedConfig, mcErr := co.configMapClient.ConfigMaps(api.OpenShiftConfigManagedNamespace).Get(api.OpenShiftConsoleConfigMapName, metav1.GetOptions{})
 	if mcErr != nil && !apierrors.IsNotFound(mcErr) {
@@ -321,7 +321,7 @@ func (co *consoleOperator) SyncConfigMap(
 		return nil, false, "FailedGetDownloadsRoute", err
 	}
 
-	defaultConfigmap, _, err := configmapsub.DefaultConfigMap(operatorConfig, consoleConfig, managedConfig, infrastructureConfig, rt, downloadsRoute)
+	defaultConfigmap, _, err := configmapsub.DefaultConfigMap(operatorConfig, consoleConfig, managedConfig, infrastructureConfig, consoleRoute, downloadsRoute)
 	if err != nil {
 		return nil, false, "FailedConsoleConfigBuilder", err
 	}
--- a/pkg/console/subresource/configmap/configmap.go
+++ b/pkg/console/subresource/configmap/configmap.go
@@ -27,6 +27,13 @@ func getApiUrl(infrastructureConfig *configv1.Infrastructure) string {
 	return ""
 }
 
+func getCliDownloadUrl(downloadsRoute *routev1.Route) string {
+	if downloadsRoute != nil {
+		return downloadsRoute.Spec.Host
+	}
+	return ""
+}
+
 func statusPageId(operatorConfig *operatorv1.Console) string {
 	if operatorConfig.Spec.Providers.Statuspage != nil {
 		return operatorConfig.Spec.Providers.Statuspage.PageID
@@ -39,7 +46,8 @@ func DefaultConfigMap(
 	consoleConfig *configv1.Console,
 	managedConfig *corev1.ConfigMap,
 	infrastructureConfig *configv1.Infrastructure,
-	rt *routev1.Route) (consoleConfigmap *corev1.ConfigMap, unsupportedOverridesHaveMerged bool, err error) {
+	rt *routev1.Route,
+	downloadsRoute *routev1.Route) (consoleConfigmap *corev1.ConfigMap, unsupportedOverridesHaveMerged bool, err error) {
 
 	defaultBuilder := &consoleserver.ConsoleServerCLIConfigBuilder{}
 	defaultConfig, err := defaultBuilder.Host(rt.Spec.Host).
@@ -47,6 +55,7 @@ func DefaultConfigMap(
 		Brand(DEFAULT_BRAND).
 		DocURL(DEFAULT_DOC_URL).
 		APIServerURL(getApiUrl(infrastructureConfig)).
+		CLIDownloadURL(getCliDownloadUrl(downloadsRoute)).
 		ConfigYAML()
 
 	extractedManagedConfig := extractYAML(managedConfig)
@@ -56,6 +65,7 @@ func DefaultConfigMap(
 		Brand(operatorConfig.Spec.Customization.Brand).
 		DocURL(operatorConfig.Spec.Customization.DocumentationBaseURL).
 		APIServerURL(getApiUrl(infrastructureConfig)).
+		CLIDownloadURL(getCliDownloadUrl(downloadsRoute)).
 		CustomLogoFile(operatorConfig.Spec.Customization.CustomLogoFile.Key).
 		CustomProductName(operatorConfig.Spec.Customization.CustomProductName).
 		StatusPageID(statusPageId(operatorConfig)).
--- a/pkg/console/subresource/configmap/configmap.go
+++ b/pkg/console/subresource/configmap/configmap.go
@@ -46,11 +46,11 @@ func DefaultConfigMap(
 	consoleConfig *configv1.Console,
 	managedConfig *corev1.ConfigMap,
 	infrastructureConfig *configv1.Infrastructure,
-	rt *routev1.Route,
+	consoleRoute *routev1.Route,
 	downloadsRoute *routev1.Route) (consoleConfigmap *corev1.ConfigMap, unsupportedOverridesHaveMerged bool, err error) {
 
 	defaultBuilder := &consoleserver.ConsoleServerCLIConfigBuilder{}
-	defaultConfig, err := defaultBuilder.Host(rt.Spec.Host).
+	defaultConfig, err := defaultBuilder.Host(consoleRoute.Spec.Host).
 		LogoutURL(defaultLogoutURL).
 		Brand(DEFAULT_BRAND).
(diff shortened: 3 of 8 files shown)


spadgett:
We'll a way to specify different URLs for different platforms

jhadvig:
@spadgett the downloads pod contains binary for linux,osx, widnows... was in impression that it is the route to the pod that we want to wire to the console.

jhadvig:
Or we want a direct link for each platform ?

spadgett:
> Or we want a direct link for each platform in the console's "Command Line Tools" page ?

Yes, IMO this is a much better experience than dumping the user in a directory listing.

jhadvig:
@spadgett also, for each platform there are is a: `oc`, `oc.tar`, `oc.zip`
I suppose we want to provide the plane `oc` binary since its a single file ?

spadgett:
Hm, maybe linux .tar.gz and mac and windows use .zip ?

jhadvig:
@spadgett @benjaminapetersen  I've update the PR
Now there will be a URL for each platform we build binary for.

PTAL

Screen Shot 2019-09-26 at 17 34 35

pkg/console/subresource/configmap/configmap.go
```diff
@@ -39,14 +46,16 @@ func DefaultConfigMap(
 	consoleConfig *configv1.Console,
 	managedConfig *corev1.ConfigMap,
 	infrastructureConfig *configv1.Infrastructure,
-	rt *routev1.Route) (consoleConfigmap *corev1.ConfigMap, unsupportedOverridesHaveMerged bool, err error) {
+	rt *routev1.Route,
```
benjaminapetersen: Can probably update to 

```golang
consoleRoute *routev1.Route,
downloadsRoute *routev1.Route) (consoleConfigmap *corev1.ConfigMap, unsupportedOverridesHaveMerged bool, err error) {
```
Mean name both routes now that there are two.

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -19,6 +19,10 @@ import (
 
 const (
 	host               = "localhost"
+	downloadHost       = "https://downloads-openshift-console.apps.some.cluster.openshift.com"
```
benjaminapetersen: To be honest, for the test I'd rather inline these.  Its a tad repetitious, but I think it makes a better test if the `input` and `output` are specific to each test and the `output` (ie, the want) is hard-coded in the specific test.

benjaminapetersen: It may be good to add one test w/o the downloads route (`nil`) and make sure the config is still properly generated:

```golang
{
     name: "Test default configmap, no customization overrides",
     // downloadsRoute: &routev1.Route{ ... }
}, {
     name: "Test default configmap, without downloads",
     // downloadsRoute: nil
}
```

pkg/console/subresource/consoleserver/types.go
```diff
@@ -39,9 +39,17 @@ type ServingInfo struct {
...
-	ConsoleBasePath    string `yaml:"consoleBasePath,omitempty"`
-	MasterPublicURL    string `yaml:"masterPublicURL,omitempty"`
+	ConsoleBaseAddress string           `yaml:"consoleBaseAddress,omitempty"`
+	ConsoleBasePath    string           `yaml:"consoleBasePath,omitempty"`
+	MasterPublicURL    string           `yaml:"masterPublicURL,omitempty"`
+	CLIDownloadURLs    *CLIDownloadURLs `yaml:"cliDownloadURLs,omitempty"`
+}
+
+// CLIDownloadURLs contains download URLs for each of the platforms we provide CLI binary.
+type CLIDownloadURLs struct {
```
benjaminapetersen: Maybe simplify this a bit:

```golang
type CLIDownloadURLs struct {
	Linux   string `yaml:"linuxDownloadURL"`
	Mac    string `yaml:"macDownloadURL"`
	Windows string `yaml:"windowsDownloadURL"`
```
Since its already a heading `CLIDownloadURLs` we don't need to repeat the `DownloadURL` part in the field names.

pkg/console/subresource/util/util.go
```diff
@@ -106,3 +106,9 @@ func HTTPS(host string) string {
 	secured := fmt.Sprintf("%s%s", protocol, host)
 	return secured
 }
+
```
benjaminapetersen: I think you can keep this func in the `configmap_builder.go` file.  It isn't shared elsewhere.

benjaminapetersen: Also, should we give it a unit test?

pkg/api/api.go
```diff
@@ -1,5 +1,12 @@
 package api
 
+const (
```
benjaminapetersen: I'm inclined to move these into the `configmap_builder.go` file, since they are only used in one place.

pkg/console/operator/sync_v400.go
```diff
@@ -316,7 +316,12 @@ func (co *consoleOperator) SyncConfigMap(
 		return nil, false, "FailedManagedConfig", mcErr
 	}
 
-	defaultConfigmap, _, err := configmapsub.DefaultConfigMap(operatorConfig, consoleConfig, managedConfig, infrastructureConfig, rt)
+	downloadsRoute, err := co.routeClient.Routes(api.TargetNamespace).Get("downloads", metav1.GetOptions{})
+	if err != nil {
+		return nil, false, "FailedGetDownloadsRoute", err
+	}
+
+	defaultConfigmap, _, err := configmapsub.DefaultConfigMap(operatorConfig, consoleConfig, managedConfig, infrastructureConfig, rt, downloadsRoute)
```
benjaminapetersen: want to update the name to `consoleRoute, downloadsRoute`?

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -332,14 +382,137 @@ servingInfo:
...
+  certFile: /var/serving-cert/tls.crt
+  keyFile: /var/serving-cert/tls.key
+providers: 
+  statuspageID: id-1234
+`,
+				},
+			},
+		},
+		{
+			name: "Test openshift-console namespace without downloads route",
```
benjaminapetersen: +1 for a test `without` the downloads route.
`Test openshift-console namespace without downloads route` - `namespace`?  I'm a bit confused by the test name, however.

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -332,14 +382,137 @@ servingInfo:
   keyFile: /var/serving-cert/tls.key
 providers: 
   statuspageID: id-1234
+`,
+				},
+			},
+		},
+		{
+			name: "Test operator config with Statuspage pageID",
```
benjaminapetersen: I believe we already had this test, trying to figure out why its showing up as new.

pkg/console/subresource/consoleserver/config_builder.go
```diff
@@ -58,6 +59,18 @@ func (b *ConsoleServerCLIConfigBuilder) APIServerURL(apiServerURL string) *Conso
 	b.apiServerURL = apiServerURL
 	return b
 }
+func (b *ConsoleServerCLIConfigBuilder) CLIDownloadURL(cliDownloadURL string) *ConsoleServerCLIConfigBuilder {
+	if cliDownloadURL != "" {
+		b.cliDownloadURLs = &CLIDownloadURLs{
```
benjaminapetersen: See my other comment about dropping the `DownloadURL` redundancy. I think we can clean this up just a bit.

pkg/console/subresource/consoleserver/types.go
benjaminapetersen: To be honest, we may want to do CLIDownloadURLs[arch][platform] just like the downloads deployment. Only amd64 now.

pkg/console/subresource/util/util.go
jhadvig: will add !

pkg/console/subresource/configmap/configmap_test.go
jhadvig: Wasnt really sure how to name the test, other then this since the setup is `nil` download route. But really should be something like : `Test getting URLs for downloading CLI, without downloads route.` ... maybe ?

pkg/console/subresource/consoleserver/config_builder.go
jhadvig: makes sense !

pkg/console/subresource/configmap/configmap_test.go
jhadvig: > It may be good to add one test w/o the downloads route (nil) and make sure the config is still properly generated:

All other TCs have the download route set :)

pkg/console/subresource/consoleserver/types.go
jhadvig: @benjaminapetersen how about? 

```go
type ClusterInfo struct {
	ConsoleBaseAddress string          `yaml:"consoleBaseAddress,omitempty"`
	ConsoleBasePath    string          `yaml:"consoleBasePath,omitempty"`
	MasterPublicURL    string          `yaml:"masterPublicURL,omitempty"`
	CLIDownloadURLs    CLIDownloadURLs `yaml:"cliDownloadURLs,omitempty"`
}

// CLIDownloadURLs contains all the architertures that we are building CLI binaries for.
type CLIDownloadURLs struct {
	Amd64 ArchPlatformsURL `yaml:"amd64,omitempty"`
}

// ArchPlatformsURL contains URLs for each of the platforms we provide CLI binary.
type ArchPlatformsURL struct {
	Linux   string `yaml:"linux"`
	Mac     string `yaml:"mac"`
	Windows string `yaml:"windows"`
}
```

Although Im not really sure about the `CLIDownloadURLs`, but its pretty generic and with the aditional architecture specified it kinda dioes make sense?

pkg/console/subresource/configmap/configmap_test.go
benjaminapetersen: The expectation is that there would be no downloads then, if there is no route, correct?  I'm good with long tests names that clearly indicate everything, including expected output.

Perhaps `Validate that there is no download CLI links when download route is `nil``?

Out test naming convention isn't super strong, so if we have to break it for clarity, thats fine.  We could always go back and improve the names of the other tests if you have something that works better.

pkg/console/subresource/consoleserver/types.go
benjaminapetersen: Yup!  Though I wonder if `ArchPlatformsURL` vs `ArchPlatformURLs`.  As "one platform has many urls", and if you had many archs, you would `[]ArchPlatformURLs`,

```golang
type CLIDownloadURLs struct {
   Amd64 ArchPlatformURLs `yaml:"amd64,omitempty"`
   Foo32 ArchPlatformURLs `yaml:"Foo32,omitempty"`
   Bar64 ArchPlatformURLs `yaml:"Bar64,omitempty"`
}
```
Since we expect 3+ URLs for archs added.

jhadvig: But if we tan to chain it like you mentioned in your previous comment it cant be an array.

jhadvig: +1 on `ArchPlatformURLs`

pkg/console/subresource/configmap/configmap_test.go
jhadvig: Also the we have the `host` const here

jhadvig:
@benjaminapetersen comments addressed. PTAL

jhadvig:
@benjaminapetersen so was looking into the console side of this story and figured out that it might be better to create [ConsoleCLIDownload](https://github.com/openshift/console-operator/blob/master/manifests/00-crd-extension-console-cli-download.yaml) CR that will represent the per architecture binaries for all the different platforms.

Eg. I've created following CR :

```
yaml
kind: ConsoleCLIDownload
metadata:
  name: download-cli-amd64
spec:
  description: >
    With the OpenShift command line interface, you can create applications and
    manage OpenShift projects from a terminal.


    The oc binary offers the same capabilities as the kubectl binary, but it is
    further extended to natively support OpenShift Container Platform features.
  displayName: oc - OpenShift Command Line Interface (CLI)  - amd64
  links:
    - href: 'https://www.example.com'
      text: Download oc for Linux
    - href: 'https://www.example.com'
      text: Download oc for Mac
    - href: 'https://www.example.com'
      text: Download oc for Windows
```

which will produce the third section in the below screenshot:

Screen Shot 2019-09-30 at 13 02 42


With this I was thinking that instead of adding logic to operator and console, let the operator create the CR for each arch type, that will contain the binaries for all the platforms.

cc'ing @spadgett for opinion

benjaminapetersen:
Ha, I like this. leverages what we already built, no point in all the additional wiring.

pkg/console/subresource/configmap/configmap.go
```diff
@@ -27,6 +27,13 @@ func getApiUrl(infrastructureConfig *configv1.Infrastructure) string {
 	return ""
 }
 
+func getCliDownloadUrl(downloadsRoute *routev1.Route) string {
+	if downloadsRoute != nil {
+		return downloadsRoute.Spec.Host
```
benjaminapetersen: `route.Spec.Host` will actually be removed, we have a `GetCanonicalHost()` func that could be used instead.

spadgett:
If we're going to do this for `oc`, we might want to do it for `odo` as well. I'd suggest special handling to make sure `oc` is moved to the top of the list. We'll need to handle the case where you are running against native k8s and this resource doesn't exist.

Only potential downside I see is we can't give special handling to the different platforms (e.g., highlight the macos link if we detect you are on a mac).

benjaminapetersen:
@spadgett  @jhadvig agree we discussed the ordering problem a little bit.  `ConsoleCLIDownload` as a type is pretty specific, We could add an optional `platform` tag to each link, which would give us the opportunity to have the filtering logic in the `console`.

benjaminapetersen:
Something like this could be worth discussing:
```yaml
kind: ConsoleCLIDownload
metadata:
  name: example-oc
spec:
  displayName: example-oc
  description: >
    This is a link with platforms
  links:
    - href: 'https://www.example.com'
      name: oc for linux
      platform: linux
    - href: 'https://www.example.com'
      name: oc for mac
      platform: mac
    - href: 'https://www.example.com'
      name: oc for windows
      platform: windows
```

benjaminapetersen:
/hold

As we investigate this other direction.

jhadvig:
Closing in favour of https://github.com/openshift/console-operator/pull/306

PR #305, opened 2019-10-02
NewServiceSyncController

benjaminapetersen:
Split SyncService() func into separate NewServiceSyncController() from main operator control loop.  

Rather than one mega controller doing all the things, we should have many smaller controllers doing specific things.  The first step will be to split off by resources as this is fairly simple.  However, some resources, such as the Deployment, may be split into several if it makes sense.

This `controller` does not make use of `monis.app/go/openshift/operator` boilerplate, thus it implements all of the following:

```golang
func sync() {}
func Run() {}
func runWorker() {}
func processNextWorkItem() {}
func newEventHandler() {}
``` 
which, for the most part, is pretty generic.  

@jhadvig FYI, this may be helpful to look at for the CRDs control loop.

### DIFF
--- a/pkg/console/controllers/service/controller.go
+++ b/pkg/console/controllers/service/controller.go
@@ -1,0 +1,162 @@
+package service
+
+import (
+	"fmt"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	coreinformersv1 "k8s.io/client-go/informers/core/v1"
+	coreclientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog"
+
+	operatorsv1 "github.com/openshift/api/operator/v1"
+	operatorclientv1 "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1"
+	"github.com/openshift/console-operator/pkg/api"
+	"github.com/openshift/console-operator/pkg/console/status"
+	"github.com/openshift/console-operator/pkg/console/subresource/service"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+)
+
+const (
+	// key is basically irrelevant
+	controllerWorkQueueKey = "service-sync-work-queue-key"
+	controllerName         = "ConsoleServiceSyncController"
+)
+
+// ctrl just needs the clients so it can make requests
+// the informers will automatically notify it of changes
+// and kick the sync loop
+type ServiceSyncController struct {
+	operatorConfigClient operatorclientv1.ConsoleInterface
+	// live clients, we dont need listers w/caches
+	serviceClient coreclientv1.ServicesGetter
+	// names
+	targetNamespace string
+	serviceName     string
+	// events
+	cachesToSync []cache.InformerSynced
+	queue        workqueue.RateLimitingInterface
+	recorder     events.Recorder
+}
+
+// factory func needs clients and informers
+// informers to start them up, clients to pass
+func NewServiceSyncController(
+	operatorConfigClient operatorclientv1.ConsoleInterface,
+	corev1Client coreclientv1.CoreV1Interface,
+	serviceInformer coreinformersv1.ServiceInformer,
+	// names
+	targetNamespace string,
+	serviceName string,
+	// events
+	recorder events.Recorder,
+) *ServiceSyncController {
+
+	corev1Client.Services(targetNamespace)
+
+	ctrl := &ServiceSyncController{
+		operatorConfigClient: operatorConfigClient,
+		serviceClient:        corev1Client,
+		// names
+		targetNamespace: targetNamespace,
+		serviceName:     serviceName,
+		// events
+		recorder:     recorder,
+		cachesToSync: nil,
+		queue:        workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), controllerName),
+	}
+
+	serviceInformer.Informer().AddEventHandler(ctrl.newEventHandler())
+	ctrl.cachesToSync = append(ctrl.cachesToSync,
+		serviceInformer.Informer().HasSynced,
+	)
+
+	return ctrl
+}
+
+func (c *ServiceSyncController) sync() error {
+	startTime := time.Now()
+	klog.V(4).Infof("started syncing service %q (%v)", c.serviceName, startTime)
+	defer klog.V(4).Infof("finished syncing service %q (%v)", c.serviceName, time.Since(startTime))
+
+	operatorConfig, err := c.operatorConfigClient.Get(api.ConfigResourceName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	switch operatorConfig.Spec.ManagementState {
+	case operatorsv1.Managed:
+		klog.V(4).Infoln("console is in a managed state: syncing service")
+	case operatorsv1.Unmanaged:
+		klog.V(4).Infoln("console is in an unmanaged state: skipping service sync")
+		return nil
+	case operatorsv1.Removed:
+		klog.V(4).Infoln("console is in a removed state: deleting service")
+		return c.removeService()
+	default:
+		return fmt.Errorf("unknown state: %v", operatorConfig.Spec.ManagementState)
+	}
+
+	updatedOperatorConfig := operatorConfig.DeepCopy()
+	_, _, svcErr := resourceapply.ApplyService(c.serviceClient, c.recorder, service.DefaultService(nil))
+	status.HandleProgressingOrDegraded(updatedOperatorConfig, "ServiceSync", "FailedApply", svcErr)
+	status.SyncStatus(c.operatorConfigClient, updatedOperatorConfig)
+
+	return svcErr
+}
+
+func (c *ServiceSyncController) removeService() error {
+	klog.V(2).Info("deleting console service")
+	defer klog.V(2).Info("finished deleting console service")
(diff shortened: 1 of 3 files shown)


pkg/console/controllers/service/controller.go
```diff
@@ -0,0 +1,148 @@
...
+	spec, _, _, err := c.operatorConfigClient.GetOperatorState()
+	if err != nil {
+		return err
+	}
+	// only run if we are managed
+	if !management.IsOperatorManaged(spec.ManagementState) {
+		return nil
+	}
+
+	// if err := SyncService(); err != nil {
```
benjaminapetersen: TODO: here is where the current `service` sync logic will be wired in once removed from the original sync loop.

pkg/console/operator/sync_v400.go
```diff
@@ -68,6 +68,9 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console,
 		return rtErr
 	}
 
+	// TODO: remove this from this sync loop and treat it as its own sync loop.
```
benjaminapetersen: Extracting this next.

pkg/console/starter/starter.go
```diff
@@ -152,6 +154,15 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		resourceSyncer,
 	)
 
+	service.NewServiceSyncController(
```
benjaminapetersen: We will likely have a bunch of controllers starting here, just all sharing the informers, clients, etc that we have already created.

pkg/console/starter/starter.go
```diff
@@ -152,6 +154,15 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
...
+		operatorClient,
+		kubeClient,
+		kubeInformersNamespaced,
+		api.OpenShiftConsoleNamespace,
```
benjaminapetersen: Passing in names to our controllers should make them a bit more flexible.

jhadvig: yeah I would rather remove this since the value will be always the same. Not sure why would we reuse this one ?

pkg/console/controllers/service/controller.go
```diff
@@ -0,0 +1,170 @@
...
+		// names
+		targetNamespace: targetNamespace,
+		serviceName:     serviceName,
+		// events
+		recorder:     recorder,
+		cachesToSync: nil,
+		queue:        workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), controllerName),
+	}
+
+	serviceInformer.Informer().AddEventHandler(ctrl.newEventHandler())
```
jhadvig: shouldn't we add `operatorConfigInformer` as well, to trigger a sync loop after config gets updated ?

pkg/console/controllers/service/controller.go
```diff
@@ -0,0 +1,170 @@
...
+		klog.V(4).Infoln("console is in an unmanaged state: skipping service sync")
+		return nil
+	case operatorsv1.Removed:
+		klog.V(4).Infoln("console is in an removed state: deleting service")
+		return c.removeService()
+	default:
+		return fmt.Errorf("unknown state: %v", operatorConfig.Spec.ManagementState)
+	}
+
+	if !management.IsOperatorManaged(operatorConfig.Spec.ManagementState) {
```
jhadvig: this is not needed anymore since we have `switch`

pkg/console/controllers/service/controller.go
```diff
@@ -0,0 +1,170 @@
...
+		return nil
+	}
+
+	updatedOperatorConfig := operatorConfig.DeepCopy()
+	_, _, svcErr := c.SyncService()
```
jhadvig: lets the `resourceapply.ApplyService( )` directly here :)

pkg/console/controllers/service/controller.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+// boilerplate, since this controller is not making use of monis.app/go boilerplate
+func (c *ServiceSyncController) Run(workers int, stopCh <-chan struct{}) {
+	defer runtime.HandleCrash()
+	defer c.queue.ShutDown()
+	klog.Infof("starting %v", controllerName)
+	defer klog.Infof("shutting down %v", controllerName)
+	if !cache.WaitForCacheSync(stopCh, c.cachesToSync...) {
+		return
```
jhadvig: lets log this cache error 
```suggestion
		klog.Infoln("caches did not sync")
		runtime.HandleError(fmt.Errorf("caches did not sync"))
		return
```

pkg/console/operator/sync_v400.go
```diff
@@ -69,12 +68,15 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console,
 		return rtErr
 	}
 
-	svc, svcChanged, svcErrReason, svcErr := co.SyncService(set.Operator)
-	toUpdate = toUpdate || svcChanged
-	status.HandleProgressingOrDegraded(updatedOperatorConfig, "ServiceSync", svcErrReason, svcErr)
-	if svcErr != nil {
-		return svcErr
-	}
+	//// TODO: remove!
```
jhadvig: Shall we leave this here ?

pkg/console/operator/sync_v400.go
```diff
@@ -186,9 +188,10 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console,
 
 	defer func() {
 		klog.V(4).Infof("sync loop 4.0.0 complete")
-		if svcChanged {
-			klog.V(4).Infof("\t service changed: %v", svc.GetResourceVersion())
-		}
+		// TODO: remove
```
jhadvig: same ?

pkg/console/operator/sync_v400.go
```diff
@@ -396,15 +399,16 @@ func (co *consoleOperator) SyncTrustedCAConfigMap(operatorConfig *operatorv1.Con
 	return actual, true, "", err
 }
 
+// TODO: remove!
```
jhadvig: same ?

pkg/console/starter/starter.go
```diff
@@ -215,7 +230,9 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		informer.Start(ctx.Done())
 	}
 
+	go consoleServiceController.Run(1, ctx.Done())
 	go consoleOperator.Run(ctx.Done())
+
```
jhadvig: nit: lets remove this one :)

pkg/console/controllers/service/controller.go
benjaminapetersen: We do not want informers for anything but `service` for this one.  
That was a mistake on my part to suggest adding them to your `Downloads` controller, we can follow & remove.  
The theory is that it can hide bugs by having unrelated resource kick your sync when it shouldn't sync.

pkg/console/controllers/service/controller.go
benjaminapetersen: Ah, thats right, I was pondering this a bit more since this `func` does a few extra things, but I agree, we don't need.

pkg/console/controllers/service/controller.go
benjaminapetersen: Yeah I'm good with that.

pkg/console/controllers/service/controller.go
benjaminapetersen: Sounds good.

pkg/console/operator/sync_v400.go
benjaminapetersen: Nope :)

pkg/console/operator/sync_v400.go
benjaminapetersen: will do.

pkg/console/starter/starter.go
benjaminapetersen: I think its good practice for controller patterns, perhaps we will have another operator in the future and would want to copy a controller across operators but not have to refactor a bunch of hard-coded values.

pkg/console/starter/starter.go
benjaminapetersen: Will do.
I left the space between "our controllers" and "lib go controllers" but doesn't really do us much.

benjaminapetersen:
@jhadvig most items addressed & updated, 2 with comments. Thx!

benjaminapetersen:
/retest

`level=fatal msg="failed to initialize the cluster: Some cluster operators are still updating: machine-api, monitoring"`

benjaminapetersen:
/retest

```
level=fatal msg="failed to initialize the cluster: Working towards 0.0.1-2019-10-17-184159: 96% complete"
```

pkg/console/starter/starter.go
jhadvig: ok :)

pkg/console/controllers/service/controller.go
```diff
@@ -0,0 +1,162 @@
...
+	defer klog.V(4).Infof("finished syncing service %q (%v)", c.serviceName, time.Since(startTime))
+
+	operatorConfig, err := c.operatorConfigClient.Get(api.ConfigResourceName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	switch operatorConfig.Spec.ManagementState {
+	case operatorsv1.Managed:
+		klog.V(4).Infoln("console is in an managed state: syncing service")
```
jhadvig:
```suggestion
		klog.V(4).Infoln("console is in a managed state: syncing service")
```

pkg/console/controllers/service/controller.go
```diff
@@ -0,0 +1,162 @@
...
+	case operatorsv1.Unmanaged:
+		klog.V(4).Infoln("console is in an unmanaged state: skipping service sync")
+		return nil
+	case operatorsv1.Removed:
+		klog.V(4).Infoln("console is in an removed state: deleting service")
```
jhadvig:
```suggestion
		klog.V(4).Infoln("console is in a removed state: deleting service")
```

benjaminapetersen:
rebased.
@jhadvig

benjaminapetersen:
/retest

```
level=fatal msg="Terraform destroy: failed to destroy using Terraform"
```

PR #306, opened 2019-10-03
Add ConsoleCliDownloads controller

jhadvig:
https://jira.coreos.com/browse/CONSOLE-1423

Adding ConsoleCliDownloads controller that will take care of ConsoleCliDownloads `oc` binaries for different platforms.

/assign @benjaminapetersen

### DIFF
--- a/manifests/03-rbac-role-cluster-extensions.yaml
+++ b/manifests/03-rbac-role-cluster-extensions.yaml
@@ -9,9 +9,9 @@ rules:
   - console.openshift.io
   resources:
   - consolelinks
-  - consoleclidownloads
   - consolenotifications
   - consoleexternalloglinks
+  - consoleclidownloads
   verbs:
   - get
   - list
--- a/pkg/console/controllers/clidownloads/controller.go
+++ b/pkg/console/controllers/clidownloads/controller.go
@@ -1,0 +1,291 @@
+package clidownloads
+
+import (
+	// standard lib
+	"fmt"
+	"time"
+
+	// kube
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog"
+
+	// openshift
+	v1 "github.com/openshift/api/console/v1"
+	operatorsv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+
+	// informers
+	consoleinformersv1 "github.com/openshift/client-go/console/informers/externalversions/console/v1"
+	operatorinformersv1 "github.com/openshift/client-go/operator/informers/externalversions/operator/v1"
+	routesinformersv1 "github.com/openshift/client-go/route/informers/externalversions/route/v1"
+
+	// clients
+	consoleclientv1 "github.com/openshift/client-go/console/clientset/versioned/typed/console/v1"
+	operatorclientv1 "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1"
+	routeclientv1 "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
+
+	// operator
+	"github.com/openshift/console-operator/pkg/api"
+	"github.com/openshift/console-operator/pkg/console/status"
+	routesub "github.com/openshift/console-operator/pkg/console/subresource/route"
+	"github.com/openshift/console-operator/pkg/console/subresource/util"
+)
+
+const (
+	controllerWorkQueueKey = "clidownloads-sync-work-queue-key"
+	controllerName         = "ConsoleCLIDownloadsSyncController"
+
+	ocCLIDownloadsCustomResourceName  = "oc-cli-downloads"
+	odoCLIDownloadsCustomResourceName = "odo-cli-downloads"
+)
+
+type CLIDownloadsSyncController struct {
+	// clients
+	consoleCliDownloadsClient consoleclientv1.ConsoleCLIDownloadInterface
+	routeClient               routeclientv1.RoutesGetter
+	operatorConfigClient      operatorclientv1.ConsoleInterface
+	// events
+	cachesToSync []cache.InformerSynced
+	queue        workqueue.RateLimitingInterface
+	recorder     events.Recorder
+}
+
+func NewCLIDownloadsSyncController(
+	// clients
+	operatorClient v1helpers.OperatorClient,
+	operatorConfigClient operatorclientv1.OperatorV1Interface,
+	cliDownloadsInterface consoleclientv1.ConsoleCLIDownloadInterface,
+	routeClient routeclientv1.RoutesGetter,
+	// informers
+	operatorConfigInformer operatorinformersv1.ConsoleInformer,
+	consoleCLIDownloadsInformers consoleinformersv1.ConsoleCLIDownloadInformer,
+	routesInformers routesinformersv1.RouteInformer,
+	// recorder
+	recorder events.Recorder,
+) *CLIDownloadsSyncController {
+
+	ctrl := &CLIDownloadsSyncController{
+		// clients
+		consoleCliDownloadsClient: cliDownloadsInterface,
+		routeClient:               routeClient,
+		operatorConfigClient:      operatorConfigClient.Consoles(),
+		// events
+		recorder: recorder,
+		queue:    workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "ConsoleCliDownloadsSyncer"),
+	}
+
+	operatorClient.Informer().AddEventHandler(ctrl.newEventHandler())
+	operatorConfigInformer.Informer().AddEventHandler(ctrl.newEventHandler())
+	consoleCLIDownloadsInformers.Informer().AddEventHandler(ctrl.newEventHandler())
+	routesInformers.Informer().AddEventHandler(ctrl.newEventHandler())
+
+	ctrl.cachesToSync = append(ctrl.cachesToSync,
+		operatorClient.Informer().HasSynced,
+		operatorConfigInformer.Informer().HasSynced,
+		consoleCLIDownloadsInformers.Informer().HasSynced,
+		routesInformers.Informer().HasSynced,
+	)
+
+	return ctrl
+}
+
+func (c *CLIDownloadsSyncController) sync() error {
+	operatorConfig, err := c.operatorConfigClient.Get(api.ConfigResourceName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	updatedOperatorConfig := operatorConfig.DeepCopy()
+
+	switch updatedOperatorConfig.Spec.ManagementState {
+	case operatorsv1.Managed:
+		klog.V(4).Infoln("console is in a managed state: syncing ConsoleCliDownloads custom resources")
+	case operatorsv1.Unmanaged:
+		klog.V(4).Infoln("console is in an unmanaged state: skipping ConsoleCliDownloads custom resources sync")
+		return nil
+	case operatorsv1.Removed:
+		klog.V(4).Infoln("console is in a removed state: deleting ConsoleCliDownloads custom resources")
+		return c.removeCLIDownloads()
+	default:
+		return fmt.Errorf("console is in an unknown state: %v", updatedOperatorConfig.Spec.ManagementState)
+	}
+
+	consoleRoute, err := c.routeClient.Routes(api.TargetNamespace).Get(api.OpenShiftConsoleDownloadsRouteName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	host := routesub.GetCanonicalHost(consoleRoute)
+	ocConsoleCLIDownloads := PlatformBasedOCConsoleCLIDownloads(host, "amd64", ocCLIDownloadsCustomResourceName)
+	_, ocCLIDownloadsErrReason, ocCLIDownloadsErr := ApplyCLIDownloads(c.consoleCliDownloadsClient, ocConsoleCLIDownloads)
+	status.HandleDegraded(updatedOperatorConfig, "OCDownloadsSync", ocCLIDownloadsErrReason, ocCLIDownloadsErr)
+	if ocCLIDownloadsErr != nil {
+		return ocCLIDownloadsErr
+	}
+
+	_, odoCLIDownloadsErrReason, odoCLIDownloadsErr := ApplyCLIDownloads(c.consoleCliDownloadsClient, ODOConsoleCLIDownloads())
+	status.HandleDegraded(updatedOperatorConfig, "ODODownloadsSync", odoCLIDownloadsErrReason, odoCLIDownloadsErr)
+	if odoCLIDownloadsErr != nil {
+		return odoCLIDownloadsErr
+	}
+
+	return nil
+}
+
+func (c *CLIDownloadsSyncController) removeCLIDownloads() error {
+	defer klog.V(4).Info("finished deleting ConsoleCliDownloads custom resources")
+	var errs []error
+	errs = append(errs, c.consoleCliDownloadsClient.Delete(ocCLIDownloadsCustomResourceName, &metav1.DeleteOptions{}))
+	errs = append(errs, c.consoleCliDownloadsClient.Delete(odoCLIDownloadsCustomResourceName, &metav1.DeleteOptions{}))
+	return utilerrors.FilterOut(utilerrors.NewAggregate(errs), errors.IsNotFound)
+}
+
+func GetPlatformURL(baseURL, platform, archiveType string) string {
+	return fmt.Sprintf("%s/%s/%s", baseURL, platform, archiveType)
+}
+
+func PlatformBasedOCConsoleCLIDownloads(host, arch, cliDownloadsName string) *v1.ConsoleCLIDownload {
+	baseURL := fmt.Sprintf("%s/%s", util.HTTPS(host), arch)
+	platforms := []struct {
+		label    string
+		key      string
+		archType string
+	}{
+		{"Linux", "linux", "oc.tar"},
+		{"Mac", "mac", "oc.zip"},
+		{"Windows", "windows", "oc.zip"},
+	}
+
+	links := []v1.Link{}
+	for _, platform := range platforms {
+		links = append(links, v1.Link{
+			Href: GetPlatformURL(baseURL, platform.key, platform.archType),
+			Text: fmt.Sprintf("Download oc for %s", platform.label),
+		})
+	}
+
+	return &v1.ConsoleCLIDownload{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: cliDownloadsName,
+		},
+		Spec: v1.ConsoleCLIDownloadSpec{
+			Description: `With the OpenShift command line interface, you can create applications and manage OpenShift projects from a terminal.
+
+The oc binary offers the same capabilities as the kubectl binary, but it is further extended to natively support OpenShift Container Platform features.
+`,
+			DisplayName: "oc - OpenShift Command Line Interface (CLI)",
+			Links:       links,
+		},
+	}
+}
+
+func ODOConsoleCLIDownloads() *v1.ConsoleCLIDownload {
+	return &v1.ConsoleCLIDownload{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: odoCLIDownloadsCustomResourceName,
+		},
+		Spec: v1.ConsoleCLIDownloadSpec{
+			Description: `OpenShift Do (odo) is a fast, iterative, and straightforward CLI tool for developers who write, build, and deploy applications on OpenShift.
+
+odo abstracts away complex Kubernetes and OpenShift concepts, thus allowing developers to focus on what is most important to them: code.
+`,
+			DisplayName: "odo - Developer-focused CLI for OpenShift",
+			Links: []v1.Link{
+				{
+					Href: "https://mirror.openshift.com/pub/openshift-v4/clients/odo/latest/",
+					Text: "Download odo",
+				},
+			},
+		},
+	}
+}
+
+// TODO: All the custom `Apply*` functions should be at some point be placed into:
+// openshift/library-go/pkg/console/resource/resourceapply/core.go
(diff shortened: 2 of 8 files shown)


pkg/console/controllers/clidownloads/controller.go
```diff
@@ -0,0 +1,197 @@
...
+	controllerName         = "ConsoleCliDownloadsSyncController"
+
+	ocCliDownloadsCustomResourceName  = "oc-cli-downloads"
+	odoCliDownloadsCustomResourceName = "odo-cli-downloads"
+
+	LinuxPlatformKey   = "linux"
+	MacPlatformKey     = "mac"
+	WindowsPlatformKey = "windows"
+)
+
```
benjaminapetersen: nit: Probably we should caps `CLI` as `CLIDownloadsSyncController `.

pkg/console/controllers/clidownloads/controller.go
```diff
@@ -0,0 +1,197 @@
...
+	consoleCliDownloadsClient consolev1client.ConsoleCLIDownloadInterface
+	routeClient               routeclientv1.RoutesGetter
+	// listers
+	cachesToSync []cache.InformerSynced
+	queue        workqueue.RateLimitingInterface
+	recorder     events.Recorder
+}
+
+// factory func needs clients and informers
+func NewCliDownloadsSyncController(
```
benjaminapetersen: nit: caps `CLI`.

pkg/console/controllers/clidownloads/controller.go
```diff
@@ -0,0 +1,197 @@
...
+
+	consoleRoute, err := c.routeClient.Routes(api.TargetNamespace).Get(api.OpenShiftConsoleRouteName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	host := routesub.GetCanonicalHost(consoleRoute)
+	requiredOCCliDownloads := platformBasedCliDownloads(host, "amd64", ocCliDownloadsCustomResourceName)
+
+	_, _, cliDownloadsErr := GetOrCreate(c.consoleCliDownloadsClient, requiredOCCliDownloads)
```
benjaminapetersen: Looks good so far, I bet we want to have a `HandleDegraded()` call here, with some potential reasons.

pkg/console/controllers/clidownloads/controller.go
```diff
@@ -0,0 +1,197 @@
...
+				{
+					Href: getPlatformURL(WindowsPlatformKey, "oc.tar"),
+					Text: "Download oc for Windows",
+				},
+			},
+		},
+	}
+}
+
+func GetOrCreate(consoleClient consolev1client.ConsoleCLIDownloadInterface, required *v1.ConsoleCLIDownload) (*v1.ConsoleCLIDownload, bool, error) {
```
benjaminapetersen: Contemplating, do we want a full `Apply` for this?  If it exists, but someone tampers with it, we probably want to stomp it back to the correct values.

pkg/console/controllers/clidownloads/controller.go
```diff
@@ -0,0 +1,197 @@
...
+	routev1Client routeclientv1.RoutesGetter,
+	consoleInformer consolev1informers.ConsoleCLIDownloadInformer,
+	recorder events.Recorder,
+) *CliDownloadsSyncController {
+
+	ctrl := &CliDownloadsSyncController{
+		// clients
+		operatorConfigClient:      operatorClient,
+		consoleCliDownloadsClient: consoleClient,
+		routeClient:               routev1Client,
```
benjaminapetersen: `consoleClient` as `cliDownloadsInterface`?

pkg/console/starter/starter.go
```diff
@@ -118,6 +128,11 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		oauthinformers.WithTweakListOptions(tweakListOptionsForOAuth),
 	)
 
+	consoleInformer := consoleinformers.NewSharedInformerFactory(
```
benjaminapetersen: Prob plural like the others `consoleInformers`.

benjaminapetersen: Then need to ensure we start the informers, this one is new:

```
	for _, informer := range []interface {
		Start(stopCh <-chan struct{})
	}{
                // etc.
		resourceSyncerInformers,
		operatorConfigInformers,
                // add here, and should be good!
		consoleInformers,
		configInformers,
                 // etc.
	} {
		informer.Start(ctx.Done())
	}
```

benjaminapetersen: Yup, start the informer and you will get this:

Screen Shot 2019-10-07 at 12 37 38 PM

pkg/console/controllers/clidownloads/controller.go
jhadvig: unfortunatelly `HandleDegraded()` is a `consoleOperator` struct method, so can't use it here. We need either to refact the `status.go` methods to be able to call on different struct, or have a separate function for the `CLIDownloadsSyncController` struct

jhadvig:
@benjaminapetersen updated the PR. The controller will make sure the `ConsoleCliDownload` CR will be created for `oc` and `odo` (@spadgett FYI)
Already started to work on the unit tests but need to align on the format of the controller.

Also I've seen https://github.com/openshift/console-operator/pull/309 ... after it will land will update the PR with `HandleDegraded` as we spoke.

PTAL

manifests/03-rbac-role-cluster-extensions.yaml
```diff
@@ -9,10 +9,20 @@ rules:
...
   resources:
   - consolelinks
-  - consoleclidownloads
   - consolenotifications
   - consoleexternalloglinks
   verbs:
   - get
   - list
   - watch
+- apiGroups:
```
benjaminapetersen: Good call, this is better.

pkg/console/controllers/clidownloads/controller.go
```diff
@@ -0,0 +1,260 @@
...
+		},
+		Spec: v1.ConsoleCLIDownloadSpec{
+			Description: `OpenShift Do (odo) is a fast, iterative, and straightforward CLI tool for developers who write, build, and deploy applications on OpenShift.
+
+odo abstracts away complex Kubernetes and OpenShift concepts, thus allowing developers to focus on what is most important to them: code.
+`,
+			DisplayName: "odo - Developer-focused CLI for OpenShift",
+			Links: []v1.Link{
+				{
+					Href: "https://mirror.openshift.com/pub/openshift-v4/clients/odo/latest/",
```
benjaminapetersen: Are we good with this `href` being hard-coded?  I imagine it doesn't really change?  Any concerns about versions?

pkg/console/starter/starter.go
```diff
@@ -118,6 +128,11 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		oauthinformers.WithTweakListOptions(tweakListOptionsForOAuth),
 	)
 
+	consoleInformers := consoleinformers.NewSharedInformerFactory(
```
benjaminapetersen: I'm considering calling this `consoleExtensionInformers`, just for clarity.  Its hard to keep track of all the different configs.

pkg/console/starter/starter.go
```diff
@@ -152,6 +167,15 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		resourceSyncer,
 	)
 
+	cliDownloadsController := clidownloads.NewCLIDownloadsSyncController(
```
benjaminapetersen: Lets organize these with `client`/`informer` next to each other, `client` first like https://github.com/openshift/console-operator/pull/310/files#diff-69ecbb024dab972c8729678abdf9c0faR79.

I think it will help, granted .  don't think there will be much added to this.

```golang
cliDownloadsController := clidownloads.NewCLIDownloadsSyncController(
  // operator is to set status, correct?
  operatorConfigClient.OperatorV1(),
  operatorClient,
  // client, informer pair for ConsoleCLIDownloads
  consoleClient.ConsoleV1().ConsoleCLIDownloads(),
  consoleInformers.Console().V1().ConsoleCLIDownloads(),
  // route just to get route... but should we be informed of a route change?
  routesClient.RouteV1(),
  recorder,
)
```

benjaminapetersen: I think we prob want the `routeInformer`.  If the `Host` changes on the route, then we need to update the `CLIDownloads`.

pkg/console/controllers/clidownloads/controller.go
jhadvig: good question, @spadgett do you know if there is any chance that this shall change ?
If so can add and entry to the operator config and read it from there ?

pkg/console/starter/starter.go
jhadvig: `consoleCLIDownloadsInformers`... per our conversation

pkg/console/controllers/clidownloads/controller.go
spadgett: I'm not sure about odo API compatibility, but at some point, we might want to lock to a specific odo version for a specific openshift version. It is hard-coded to latest today.

@mohammedzee1000 @kadel We now have an extension point in console for adding to the command-line download page. Is there another operator that should own creating the CR for the odo download? If console owns it, what version should we link to?

pkg/console/controllers/clidownloads/controller.go
```diff
@@ -0,0 +1,261 @@
...
+	return nil
+}
+
+func GetPlatformURL(baseURL, platform, archiveType string) string {
+	return fmt.Sprintf("%s/%s/%s", baseURL, platform, archiveType)
+}
+
+func PlatformBasedOCConsoleCLIDownloads(host, arch, cliDownloadsName string) *v1.ConsoleCLIDownload {
+	baseURL := fmt.Sprintf("%s/%s", util.HTTPS(host), arch)
+
```
benjaminapetersen: We could potentially have a map of strings here:

```golang
platforms := [string]string{
  "Linux": "linux",
   "Mac": "mac",
  "Windows": "windows",
}
```

and a loop to make the `Links`:

```golang
links := []v1.Link{}
for label, platform range platforms {
   links := append(links, Link{
      Href: GetPlatformURL(baseURL, platform, "oc.tar"),
      Text: fmt.Sprintf("Download oc for %v", Label)
   })
}
```
That or maybe just inline the `*PlatformKey`s since they are tied to the specifc "Download for *Platform*` text.

pkg/console/controllers/clidownloads/controller.go
```diff
@@ -0,0 +1,261 @@
...
+			Links: []v1.Link{
+				{
+					Href: "https://mirror.openshift.com/pub/openshift-v4/clients/odo/latest/",
+					Text: "Download odo",
+				},
+			},
+		},
+	}
+}
+
```
benjaminapetersen: Pondering if we should make this 

```golang
ApplyConsoleCLIDownload(....)
``` 
We can probably look at a func in `library-go` like:

```golang
func ApplyService(client coreclientv1.ServicesGetter, recorder events.Recorder, required *corev1.Service) (*corev1.Service, bool, error) {
  // does things
}
```
and copy it.  I imagine we want the `eventRecorder` in there to create events if things go awry. 

So something like:

```golang
func ApplyConsoleCLIDownload(cliDownloadsClient CliDownloadsGetter, recorder events.Recorder, required *v1.ConsoleCLIDownload) (v1.ConsoleCLIDownload, string, error) {
  // more or less copy the ApplyService func, but tweak it to fit this resource.
}
```

jhadvig: `CreateOrUpdate` already has similar logic... need to add the recored and rename though :)

jhadvig: so looked at adding the recorder but noticed that the methods that the recorder is passed into are private (`reportCreateEvent` && `reportUpdateEvent`)

Also our custom apply [CustomApplyOAuth](https://github.com/openshift/console-operator/blob/master/pkg/console/subresource/oauthclient/oauthclient.go#L21) method is not using the recored as well, supposedly due to the fact the methods are private.

Would ship it for now and add this as a techDebt ?

jhadvig:
@benjaminapetersen Ive updated the PR based on comments also added comments.

 Added a `removeCLIDownloads()` method for the `Removed` operator state.

Only thing that is missing are the e2e tests that I've started to work on, but I think it would be better to address them as a separate PR, since this is already quite beefy 😄 

Other thing is that we need to figure out https://github.com/openshift/console-operator/pull/306#discussion_r333060441

PTAL

pkg/console/controllers/clidownloads/controller.go
kadel: Currently, the odo is aiming to be compatible with 3.9 and up. odo is trying to be compatible with every supported OCP version. Using the latest should be fine.

There is no odo operator if the console can own this that would be great.

jhadvig: @spadgett I guess than our operator will managed `odo` as well

spadgett: @kadel Should we point to odo `latest`? Or a version like `v0.0.20`?

pkg/console/controllers/clidownloads/controller.go
benjaminapetersen: Nice!

pkg/console/controllers/clidownloads/controller.go
benjaminapetersen: I'm fine managing it if we can point to `latest`.  If we get to a point where we need to keep up with version numbers, would we want to revisit the mechanism for that?

pkg/console/controllers/clidownloads/controller.go
```diff
@@ -0,0 +1,285 @@
...
+		return c.removeCLIDownloads()
+	default:
+		return fmt.Errorf("console is in an unknown state: %v", updatedOperatorConfig.Spec.ManagementState)
+	}
+
+	consoleRoute, err := c.routeClient.Routes(api.TargetNamespace).Get(api.OpenShiftConsoleDownloadsRouteName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
```
benjaminapetersen: Pondering if we want 2 different `types`:

```golang
status.HandleDegraded(client, "OCDownloadsSync", ....)
status.HandleDegraded(client, "ODODownloadsSync",....)
```
Since we `return err` immediately after, we can't really stomp over the `oc` sync error with an `odo` sync error. But, in general, I think we don't want to reuse `type` to avoid the potential for stomping on an err (say if we refactor down the road and the shape here changes).

pkg/console/controllers/clidownloads/controller.go
benjaminapetersen: Also, just validating, `odo` is not built for several `platforms`?  `linux` only, I assume?  Should this be clarifiedin the text?

pkg/console/controllers/clidownloads/controller.go
benjaminapetersen: Agree!

Do you want to throw a couple `TODO` comments in there quick?  This would be helpful if we promote it to `openshift/library-go/pkg/console/resource/resourceapply/core.go` (does not yet exist, but if others are going to use these in their operators, we could ensure the appropriate `apply` funcs exist for each of our custom resources.)

pkg/console/starter/starter.go
benjaminapetersen: 👍

pkg/console/controllers/clidownloads/controller.go
jhadvig: > Also, just validating, odo is not built for several platforms? linux only, I assume? Should this be clarifiedin the text?

we probably should

pkg/console/controllers/clidownloads/controller.go
jhadvig: Makes sense to me !

pkg/console/controllers/clidownloads/controller.go
kadel: > Also, just validating, `odo` is not built for several `platforms`? `linux` only, I assume? Should this be clarifiedin the text?

odo is built for all 3 platforms (windows, mac, linux) all 3 of them are available at https://mirror.openshift.com/pub/openshift-v4/clients/odo/latest/

kadel: > I'm fine managing it if we can point to `latest`. If we get to a point where we need to keep up with version numbers, would we want to revisit the mechanism for that?

Right now the only time where I see that we would want to keep up with the versions is when there are some big changes in core openshfit or k8s apis. Wouldn't that also result in OpenShift v5?

pkg/console/starter/starter.go
benjaminapetersen: Check that, you'll need both for `route`:

```golang
routeClient // so we can GET the downloads route & pluck out the canonical host.
routeInformer // notify us of any route changes
```

pkg/console/controllers/clidownloads/controller.go
benjaminapetersen: Should this then mirror the links we generate from console, by listing the 3 links separately?

```
odo-linux-amd64.tar.gz  
odo-darwin-amd64.tar.gz  
odo-windows-amd64.exe
```

benjaminapetersen: Screen Shot 2019-10-07 at 12 37 38 PM

benjaminapetersen: This would imply that the file names are stable, but is likely a better experience if we can confirm.

pkg/console/controllers/clidownloads/controller.go
```diff
@@ -0,0 +1,295 @@
...
+	odoCLIDownloadsCustomResourceName = "odo-cli-downloads"
+)
+
+type CLIDownloadsSyncController struct {
+	// clients
+	operatorClient            v1helpers.OperatorClient
+	consoleCliDownloadsClient consoleclientv1.ConsoleCLIDownloadInterface
+	routeClient               routeclientv1.RoutesGetter
+	// listers
+	routesLister         routeslisterv1.RouteLister
```
benjaminapetersen: So you can drop the `routesLister` now, we aren't using it.

Recalling that:
- A client handles CRUD in our get,list,watch,create,update,delete, etc
- a Lister is a `cached` version of `Get` and `List`.  We don't really need it (and we aren't actually implementing it anyway here) since there are not a lot of these resources.

pkg/console/controllers/clidownloads/controller.go
```diff
@@ -0,0 +1,295 @@
...
+		operatorConfigLister: operatorConfigInformer.Lister(),
+		// events
+		recorder: recorder,
+		queue:    workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "ConsoleCliDownloadsSyncer"),
+	}
+
+	operatorClient.Informer().AddEventHandler(ctrl.newEventHandler())
+	operatorConfigInformer.Informer().AddEventHandler(ctrl.newEventHandler())
+	consoleCLIDownloadsInformers.Informer().AddEventHandler(ctrl.newEventHandler())
+	routesInformers.Informer().AddEventHandler(ctrl.newEventHandler())
```
jhadvig: > So you can drop the routesLister now, we aren't using it.

shouldn't we keep it at least here ?

pkg/console/controllers/clidownloads/controller.go
kadel: I think that it might be safer to go with a link to the latest directory for now.
We are just preparing the 1.0 release, and it might change a little bit. After the release it will be stable, but I can't guarantee the exact filenames right now.

jhadvig:
@benjaminapetersen comments addressed. Left the `odo` link to the latest directory, instead of particular platform binary, per [discussion](https://github.com/openshift/console-operator/pull/306#discussion_r334367942) with @kadel 

PTAL

pkg/console/controllers/clidownloads/controller.go
benjaminapetersen: ok that works.

benjaminapetersen: I would probably recommend the individual files as a better experience, once you are set.

pkg/console/controllers/clidownloads/controller.go
benjaminapetersen: So, you do need the `informer`, yes!
And the `client`.
But we don't need the `lister`.

pkg/console/controllers/clidownloads/controller.go
```diff
@@ -0,0 +1,294 @@
...
+	routesInformers routesinformersv1.RouteInformer,
+	// recorder
+	recorder events.Recorder,
+) *CLIDownloadsSyncController {
+
+	ctrl := &CLIDownloadsSyncController{
+		// clients
+		consoleCliDownloadsClient: cliDownloadsInterface,
+		routeClient:               routeClient,
+		// listers
```
benjaminapetersen: You don't need the `operatorconfiglister` here.  We only need the `informer` below to `Inform()` & kick our watches.  And we only need a `client` for when we need to do a `.Get()`.

benjaminapetersen: After a little more research and a few more conversations, confirmed we only need the `Informer` and `client` for the actual `CLIDownloads` resource.  We shouldn't watch operator config or anything else.

benjaminapetersen:
/retest

Terraform failed.

benjaminapetersen:
And here is our test failure:

```
fail [github.com/openshift/origin/test/extended/authorization/rbac/groups_default_rules.go:221]: Oct 15 19:10:38.376: system:authenticated has extra permissions in namespace "":
{APIGroups:["console.openshift.io"], Resources:["consoleclidownloads"], Verbs:["create" "update" "delete"]}
```

@spadgett

benjaminapetersen:
/retest

https://github.com/openshift/origin/pull/23975 has merged so test should be unblocked.

manifests/03-rbac-role-cluster-extensions.yaml
```diff
@@ -9,10 +9,20 @@ rules:
...
+  - console.openshift.io
+  resources:
+  - consoleclidownloads
+  verbs:
+  - get
+  - list
+  - watch
+  - create
```
spadgett: This should only be get/list/watch

benjaminapetersen: I think he needs full `CRUD` for this one, since the `operator` is creating 2 of this resource, or updating if tampered with.

benjaminapetersen: I updated the test incorrectly, added this resource to the same list as the rest of our resources.  My mistake, I forgot this resource is more actively managed than our others.

benjaminapetersen: So you probably need to pull the `consoleclidownloads` out of `ClusterRole` `console-extensions-reader` file and put it in its own file. Name the role something like `console-clidownloads-extension-manager` and then create a `ClusterRoleBinding` for `console-operator`.

Then we can update the `origin` test to allow this.

spadgett: No, this is wrong. This is the role that gets assigned to `system:authenticated`. We *cannot* let all users edit these. The operator role is a different one.

spadgett: > I updated the test incorrectly, added this resource to the same list as the rest of our resources. My mistake, I forgot this resource is more actively managed than our others.

Your origin test updates are correct.

spadgett:
/lgtm cancel
RBAC is incorrect

manifests/03-rbac-role-cluster-extensions.yaml
benjaminapetersen: Right, intended to mean "console operator only" definitely not, "everyone".

jhadvig:
@benjaminapetersen PR updated. PTAL

PR #307, opened 2019-10-03
simplify download server

rphillips:
Instead of maintaining a custom threaded http server, this PR migrates the downloader to use Python's included threaded http server.

This PR also captures sigterm, which was already merged into master.

### DIFF
--- a/manifests/07-downloads-deployment.yaml
+++ b/manifests/07-downloads-deployment.yaml
@@ -59,29 +59,15 @@ spec:
         - '-c'
         - |
           cat <<EOF >>/tmp/serve.py
-          import BaseHTTPServer, os, re, signal, SimpleHTTPServer, socket, sys, tarfile, tempfile, threading, time, zipfile
+          from BaseHTTPServer import HTTPServer, BaseHTTPRequestHandler
+          from SocketServer import ThreadingMixIn
+          import SimpleHTTPServer, os, re, signal, sys, tarfile, tempfile, zipfile
+          
+          class ThreadedHTTPServer(ThreadingMixIn, HTTPServer):
+              pass
 
           signal.signal(signal.SIGTERM, lambda signum, frame: sys.exit(0))
-
-          # Launch multiple listeners as threads
-          class Thread(threading.Thread):
-              def __init__(self, i, socket):
-                  threading.Thread.__init__(self)
-                  self.i = i
-                  self.socket = socket
-                  self.daemon = True
-                  self.start()
-
-              def run(self):
-                  httpd = BaseHTTPServer.HTTPServer(addr, SimpleHTTPServer.SimpleHTTPRequestHandler, False)
-
-                  # Prevent the HTTP server from re-binding every handler.
-                  # https://stackoverflow.com/questions/46210672/
-                  httpd.socket = self.socket
-                  httpd.server_bind = self.server_close = lambda self: None
-
-                  httpd.serve_forever()
-
+          
           temp_dir = tempfile.mkdtemp()
           print('serving from {}'.format(temp_dir))
           os.chdir(temp_dir)
@@ -89,7 +75,6 @@ spec:
               os.mkdir(arch)
               for operating_system in ['linux', 'mac', 'windows']:
                   os.mkdir(os.path.join(arch, operating_system))
-
           for arch, operating_system, path in [
                   ('amd64', 'linux', '/usr/bin/oc'),
                   ('amd64', 'mac', '/usr/share/openshift/mac/oc'),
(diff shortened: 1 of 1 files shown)


manifests/07-downloads-deployment.yaml
```diff
@@ -59,59 +59,42 @@ spec:
...
+                      tar.add(path, basename)
+                  with zipfile.ZipFile('{}.zip'.format(archive_path_root), 'w') as zip:
+                      zip.write(path, basename)
+          
+              addr = ('', 8080)
+              server = ThreadedHTTPServer(addr, SimpleHTTPServer.SimpleHTTPRequestHandler)
+              print('Starting server, use <Ctrl-C> to stop')
+              server.serve_forever()
+              EOF
+          python2 /tmp/serve.py  # the cli image only has Python 2.7
```
wking: You're dropping `exec` from #275.  I think we want to keep that to avoid a dangling shell process.

manifests/07-downloads-deployment.yaml
```diff
@@ -59,59 +59,42 @@ spec:
...
-          sock.bind(addr)
-          sock.listen(5)
-
-          [Thread(i, socket=sock) for i in range(100)]
-          time.sleep(9e9)
-          EOF
-          exec python2 /tmp/serve.py  # the cli image only has Python 2.7
+          signal.signal(signal.SIGTERM, lambda signum, frame: sys.exit(0))
+          
+          if __name__ == '__main__':
```
wking: Do we care about this?  We're going to execute the script immediately below; it will never be a library.  I'd rather drop this line and avoid the churny indent change.

benjaminapetersen:
/hold

for review comments

manifests/07-downloads-deployment.yaml
rphillips: updated and rebased

manifests/07-downloads-deployment.yaml
rphillips: updated.

benjaminapetersen:
@rphillips please also clarify the what & why of the PR in the description.  Thanks!

manifests/07-downloads-deployment.yaml
```diff
@@ -104,16 +89,11 @@ spec:
...
           addr = ('', 8080)
-          sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-          sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-          sock.bind(addr)
-          sock.listen(5)
-
-          [Thread(i, socket=sock) for i in range(100)]
-          time.sleep(9e9)
+          server = ThreadedHTTPServer(addr, SimpleHTTPServer.SimpleHTTPRequestHandler)
+          print('Starting server, use <Ctrl-C> to stop')
```
wking: I don't think we need this line; nobody's at a teminal to ctrl-c it.

manifests/07-downloads-deployment.yaml
```diff
@@ -59,37 +59,22 @@ spec:
         - '-c'
         - |
           cat <<EOF >>/tmp/serve.py
-          import BaseHTTPServer, os, re, signal, SimpleHTTPServer, socket, sys, tarfile, tempfile, threading, time, zipfile
+          from BaseHTTPServer import HTTPServer, BaseHTTPRequestHandler
+          from SocketServer import ThreadingMixIn
+          import SimpleHTTPServer, os, re, signal, sys, tarfile, tempfile, zipfile
+          
+          class ThreadedHTTPServer(ThreadingMixIn, HTTPServer):
```
wking: I added the bulk of this script in #177.  As discussed there, it was largely a copy-paste from ci-operator.  @smarterclayton added that code on 2018-06-25 in openshift/ci-operator#35.  I'm not familiar enough with its source to know what it's all about, but he links to https://stackoverflow.com/questions/46210672/ below and that says:

> No, I don't want a hack such as ThreadingMixIn that merely collects up the response and returns it as a unit.

So I expect he was aware of this approach and has consciously decided against it.  But maybe there have been stdlib improvements in 2.7 since that SO post?  I haven't hunted deeply enough to know.

manifests/07-downloads-deployment.yaml
rphillips: Fixed

manifests/07-downloads-deployment.yaml
rphillips: I don't understand the original motivation... The main thread accepts the client and uses a thread to handle the I/O of the client. There is not a streaming interface with the current code.

benjaminapetersen:
/close

I'm going to close this as I'm not sure we have a good reason for the change.  Feel free to reopen if we want to pick up the conversation again!

PR #312, opened 2019-10-15
[WIP] Migrate/operator boilerplate

benjaminapetersen:
Since https://github.com/enj/go has been forked as https://github.com/openshift/operator-boilerplate, we need to update our dependencies to import it properly.

Depends on:
- [ ] https://github.com/openshift/operator-boilerplate/pull/1 (fix internal import paths)


pkg/console/operator/operator.go
```diff
@@ -22,13 +22,15 @@ import (
 	oauthinformersv1 "github.com/openshift/client-go/oauth/informers/externalversions/oauth/v1"
 	routeclientv1 "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
 	"github.com/openshift/console-operator/pkg/api"
+	"github.com/openshift/console-operator/vendor.bak/github.com/openshift/operator-boilerplate/openshift/operator"
```
jhadvig: Is this iomport correct ? Looks wrong, but maybe I'm wrong 😄

benjaminapetersen: This is incorrect.  I will fix when that repo settles down.  Not sure if it will, or if we need to inline the files.  Discussion in progress.

PR #314, opened 2019-10-16
Temporarily return extension CRDs from cluster-config-operator to our Manifests dir

benjaminapetersen:
They shall return to us!
To make @deads2k happy 😄 
/cc @spadgett @jhadvig 

TODO as follow-up:
- [x] revert https://github.com/openshift/cluster-config-operator/pull/95 once this PR has merged. 
- [x] investigate mechanism to generate these from `openshift/api` type definitions to avoid drift 
- [ ] Reopen #316 and import the CRDs instead.

### DIFF
--- a/manifests/00-crd-consoleexternalloglinks.yaml
+++ b/manifests/00-crd-consoleexternalloglinks.yaml
@@ -1,0 +1,89 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: consoleexternalloglinks.console.openshift.io
+  annotations:
+    displayName: ConsoleExternalLogLinks
+    description: ConsoleExternalLogLink is an extension for customizing OpenShift
+      web console log links.
+spec:
+  scope: Cluster
+  group: console.openshift.io
+  versions:
+    - name: v1
+      served: true
+      storage: true
+  names:
+    plural: consoleexternalloglinks
+    singular: consoleexternalloglink
+    kind: ConsoleExternalLogLink
+    listKind: ConsoleExternalLogLinkList
+  additionalPrinterColumns:
+    - name: Text
+      type: string
+      JSONPath: .spec.text
+    - name: HrefTemplate
+      type: string
+      JSONPath: .spec.hrefTemplate
+    - name: Age
+      type: date
+      JSONPath: .metadata.creationTimestamp
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      description: ConsoleExternalLogLink is an extension for customizing OpenShift
+        web console log links.
+      type: object
+      required:
+        - spec
+      properties:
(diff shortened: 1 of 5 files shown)


manifests/00-crd-consoleexternalloglinks.yaml
```diff
@@ -0,0 +1,85 @@
...
+            hrefTemplate:
+              description: "hrefTemplate is an absolute secure URL (must use https)
+                for the log link including variables to be replaced. Variables are
+                specified in the URL with the format ${variableName}, for instance,
+                ${containerName} and will be replaced with the corresponding values
+                from the resource. Resource is a pod. Supported variables are: - ${resourceName}
+                - name of the resource which containes the logs - ${resourceUID} -
+                UID of the resource which contains the logs               - e.g. `11111111-2222-3333-4444-555555555555`
+                - ${containerName} - name of the resource's container that contains
+                the logs - ${resourceNamespace} - namespace of the resource that contains
```
spadgett: This is missing `resourceNamespaceUID` from https://github.com/openshift/api/pull/466

It would be good to fix the formatting here

benjaminapetersen: Yikes the `generate` definitely did not format this well.

benjaminapetersen: Bah, and we have a `/hold` on the updates with the `patterns`, so I'll have to reconcile by hand.  😢  
(https://github.com/openshift/cluster-config-operator/pull/101/files)

benjaminapetersen: Updated.

manifests/00-crd-consoleexternalloglinks.yaml
```diff
@@ -0,0 +1,89 @@
...
+                ${containerName} and will be replaced with the corresponding values
+                from the resource. Resource is a pod. Supported variables are:
+                - ${resourceName} - name of the resource which containes the logs
+                - ${resourceUID} - UID of the resource which contains the logs
+                  - e.g. `11111111-2222-3333-4444-555555555555`
+                - ${containerName} - name of the resource's container that contains the logs
+                - ${resourceNamespace} - namespace of the resource that contains the logs
+                - ${resourceNamespaceUID} - namespace UID of the resource that contains the logs
+                - ${podLabels} - JSON representation of labels matching the pod with the logs
+                  - e.g. `{\"key1\":\"value1\",\"key2\":\"value2\"}`
```
benjaminapetersen: Strangely the embedded `"` within the overall `"` seems to work fine, here.

benjaminapetersen:
/close
#315 imports the generator so we don't have to do this by hand.

benjaminapetersen:
/retest

@spadgett lets run with this one temporarily so we don't have a broken merge queue.

PR #315, opened 2019-10-16
Makefile CRD generation

benjaminapetersen:
Imports `./vendor/github.com/openshift/library-go/alpha-build-machinery/make/` so we can leverage the CRD generator commands & regenerates our CRDs in `/manifests`.

TODO as follow-up:
- [ ] Revert https://github.com/openshift/cluster-config-operator/pull/95 once this PR has merged. 
   - [ ] Opened https://github.com/openshift/cluster-config-operator/pull/103
- [x] Investigate mechanism to generate these from `openshift/api` type definitions to avoid drift 

/cc @spadgett @jhadvig


benjaminapetersen:
/close

in favor of #316

PR #317, opened 2019-10-21
remove trailing slash for odo href

jhadvig:
Regexp that we have on the `spec.links.href` [field](https://github.com/openshift/console-operator/blob/master/manifests/00-crd-consoleclidownloads.yaml#L82) refuses trailing slashes on the `href` field. For that cause the `odo` CR wont be created.

/assign @benjaminapetersen

### DIFF
--- a/pkg/console/controllers/clidownloads/controller.go
+++ b/pkg/console/controllers/clidownloads/controller.go
@@ -201,7 +201,7 @@ odo abstracts away complex Kubernetes and OpenShift concepts, thus allowing deve
 			DisplayName: "odo - Developer-focused CLI for OpenShift",
 			Links: []v1.Link{
 				{
-					Href: "https://mirror.openshift.com/pub/openshift-v4/clients/odo/latest/",
+					Href: "https://mirror.openshift.com/pub/openshift-v4/clients/odo/latest",
 					Text: "Download odo",
 				},
 			},


spadgett:
> Regexp that we have on the `spec.links.href` [field](https://github.com/openshift/console-operator/blob/master/manifests/00-crd-consoleclidownloads.yaml#L82) refuses trailing slashes on the `href` field. For that cause the `odo` CR wont be created.
> 
> /assign @benjaminapetersen

It looks to me like the regex was copied from the documentation URL and is wrong. We should loosen the regex. Documentation URL has to have a trailing slash. None of these other links do. 

We should consider a backport.

/hold

spadgett:
@jhadvig Can you create a Bugzilla for the incorrect regex? We'll want to backport this. There's no reason to have this restriction on any of the link extensions (except documentation URL, which is a special case).

pkg/console/controllers/clidownloads/controller.go
```diff
@@ -201,7 +201,7 @@ odo abstracts away complex Kubernetes and OpenShift concepts, thus allowing deve
 			DisplayName: "odo - Developer-focused CLI for OpenShift",
 			Links: []v1.Link{
 				{
-					Href: "https://mirror.openshift.com/pub/openshift-v4/clients/odo/latest/",
+					Href: "https://mirror.openshift.com/pub/openshift-v4/clients/odo/latest",
```
benjaminapetersen: This will redirect to `https://mirror.openshift.com/pub/openshift-v4/clients/odo/latest/`, so it does work.  We just need to decide if we need to support the trailling `/` and relax the `regex`.

jhadvig: already submitted the [PR to fix the validation](https://github.com/openshift/api/pull/485) ... Will probably close this one since the updated link will redirect to the one with the trailing space anyway... @benjaminapetersen thoughts ?

benjaminapetersen: Agree!

jhadvig:
Closing after discussion with @benjaminapetersen

PR #318, opened 2019-10-21
Add e2e test for oc and odo ConsoleCLIDownloads CRs

jhadvig:
Adding the TCs to the Managed, Unmanaged and Removed test suites, since the testing scenario is pretty much the same as other resources.

Due to the wrong regexp on the ConsoleCLIDownloads CRD the added e2e will fail until it will be fixed.

/hold

/assign @benjaminapetersen

### DIFF
--- a/pkg/console/controllers/clidownloads/controller.go
+++ b/pkg/console/controllers/clidownloads/controller.go
@@ -44,9 +44,6 @@ import (
 const (
 	controllerWorkQueueKey = "clidownloads-sync-work-queue-key"
 	controllerName         = "ConsoleCLIDownloadsSyncController"
-
-	ocCLIDownloadsCustomResourceName  = "oc-cli-downloads"
-	odoCLIDownloadsCustomResourceName = "odo-cli-downloads"
 )
 
 type CLIDownloadsSyncController struct {
@@ -125,7 +122,7 @@ func (c *CLIDownloadsSyncController) sync() error {
 	}
 
 	host := routesub.GetCanonicalHost(consoleRoute)
-	ocConsoleCLIDownloads := PlatformBasedOCConsoleCLIDownloads(host, "amd64", ocCLIDownloadsCustomResourceName)
+	ocConsoleCLIDownloads := PlatformBasedOCConsoleCLIDownloads(host, "amd64", api.OCCLIDownloadsCustomResourceName)
 	_, ocCLIDownloadsErrReason, ocCLIDownloadsErr := ApplyCLIDownloads(c.consoleCliDownloadsClient, ocConsoleCLIDownloads)
 	status.HandleDegraded(updatedOperatorConfig, "OCDownloadsSync", ocCLIDownloadsErrReason, ocCLIDownloadsErr)
 	if ocCLIDownloadsErr != nil {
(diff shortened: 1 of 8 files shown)


test/e2e/framework/framework.go
```diff
@@ -69,6 +70,10 @@ func GetResource(client *ClientSet, resource string) (runtime.Object, error) {
 		res, err = GetConsoleService(client)
 	case "Route":
 		res, err = GetConsoleRoute(client)
+	case "OCCLIDownloads":
```
benjaminapetersen: This is where I wish we had taken `name` as a param as well, instead of creating `OCCLIDownloads` and `ODOCLIDownloads` as new magic strings to imply 2 different resources of the same kind.

test/e2e/framework/framework.go
```diff
@@ -120,6 +128,10 @@ func deleteResource(client *ClientSet, resource string) error {
 		err = client.Core.Services(consoleapi.OpenShiftConsoleNamespace).Delete(consoleapi.OpenShiftConsoleServiceName, &metav1.DeleteOptions{})
 	case "Route":
 		err = client.Routes.Routes(consoleapi.OpenShiftConsoleNamespace).Delete(consoleapi.OpenShiftConsoleRouteName, &metav1.DeleteOptions{})
+	case "OCCLIDownloads":
```
benjaminapetersen: Same here, wish we didn't need these two internal custom strings `OCCLIDownloads` and `ODOCLIDownloads`.

pkg/console/controllers/clidownloads/controller.go
```diff
@@ -144,8 +141,8 @@ func (c *CLIDownloadsSyncController) sync() error {
 func (c *CLIDownloadsSyncController) removeCLIDownloads() error {
 	defer klog.V(4).Info("finished deleting ConsoleCliDownloads custom resources")
 	var errs []error
-	errs = append(errs, c.consoleCliDownloadsClient.Delete(ocCLIDownloadsCustomResourceName, &metav1.DeleteOptions{}))
-	errs = append(errs, c.consoleCliDownloadsClient.Delete(odoCLIDownloadsCustomResourceName, &metav1.DeleteOptions{}))
+	errs = append(errs, c.consoleCliDownloadsClient.Delete(api.OCCLIDownloadsCustomResourceName, &metav1.DeleteOptions{}))
```
benjaminapetersen: Makes sense!

test/e2e/framework/framework.go
```diff
@@ -40,7 +41,7 @@ func ResetClusterProxyConfig(client *ClientSet) error {
 }
 
 func DeleteAll(t *testing.T, client *ClientSet) {
-	resources := []string{"Deployment", "Service", "Route", "ConfigMap"}
+	resources := []string{"Deployment", "Service", "Route", "ConfigMap", "OCCLIDownloads", "ODOCLIDownloads"}
```
benjaminapetersen: Pondering this.

test/e2e/framework/framework.go
```diff
@@ -107,8 +112,11 @@ func GetConsoleRoute(client *ClientSet) (*routev1.Route, error) {
 }
 
 func GetConsoleDeployment(client *ClientSet) (*appv1.Deployment, error) {
-	deployment, err := client.Apps.Deployments(consoleapi.OpenShiftConsoleNamespace).Get(consoleapi.OpenShiftConsoleDeploymentName, metav1.GetOptions{})
-	return deployment, err
+	return client.Apps.Deployments(consoleapi.OpenShiftConsoleNamespace).Get(consoleapi.OpenShiftConsoleDeploymentName, metav1.GetOptions{})
```
benjaminapetersen: Good cleanup.

test/e2e/framework/framework.go
benjaminapetersen: A fix could be, if we don't want to restructure the functions:

```golang
// use the actual names & run the switch against that, so we don't have a string representing another string:
resources := []string{
   "Deployment",   // might want to update to "console-deployment"
   "Service",  // might want to update to "console-service"
   // etc
   consoleapi.OCCLIDownloadsCustomResourceName,
   consoleapi.ODOCLIDownloadsCustomResourceName,
```
And maybe just leave a `TODO: ideally use Name + Kind` as something we could work on later if it became an issue.

jhadvig:
@benjaminapetersen fixed last failing test. Should be all god now.
PTAL

test/e2e/framework/framework.go
```diff
@@ -52,39 +65,31 @@ func DeleteAll(t *testing.T, client *ClientSet) {
...
 	case "Route":
-		res, err = GetConsoleRoute(client)
+		res, err = client.Routes.Routes(resource.namespace).Get(resource.name, metav1.GetOptions{})
+	case "ConsoleCLIDownloads":
+		res, err = client.ConsoleCliDownloads.Get(resource.name, metav1.GetOptions{})
 	case "Deployment":
 		fallthrough
 	default:
-		res, err = GetConsoleDeployment(client)
+		res, err = client.Apps.Deployments(resource.namespace).Get(resource.name, metav1.GetOptions{})
```
benjaminapetersen: Should we `fallthrough` or should we `errror` with "not a valid resource"?

jhadvig:
@benjaminapetersen also experiencing a lot of `timed out waiting for the condition` even though locally all the tests are passing just fine...

PR #316, opened 2019-10-16
Bug 1764224: Use extension manifests from openshift/api

benjaminapetersen:
3rd pass on this PR today :)

We can just copy the generated CRDs from `openshift/api` in `/vendor` in our `Dockerfile`s.

TODO as follow-up:
- [x] delete `/manifests` CRDs for extensions, since we are importing from `/vendor`.
- [x] Revert https://github.com/openshift/cluster-config-operator/pull/95 once this PR has merged. 
   - [x] Opened https://github.com/openshift/cluster-config-operator/pull/103
- [x] Investigate mechanism to generate these from `openshift/api` type definitions to avoid drift 

/cc @spadgett @jhadvig

PR #321, opened 2019-10-22
Update CRD manifests due to the Href field validation update

jhadvig:
Copied the validation [change](https://github.com/openshift/api/pull/485/files) change. Although not sure why there are so many changes.

/assign @benjaminapetersen


benjaminapetersen:
I think we probably want this one instead:
https://github.com/openshift/console-operator/pull/316

Perhaps with a fresh `bump deps`.

benjaminapetersen:
/close

Lets consume these now from `openshift/api` via #321

PR #324, opened 2019-10-23
Bug 1764227: Incorrect validation pattern on ConsoleCLIDownloads Href field

jhadvig:
Updated just the validation pattern since if I've would copied the CRD yamls from openshift/api master, there would be more changes then necessary.

/assign @benjaminapetersen 
cc'ing @spadgett


benjaminapetersen:
And note to self, don't `cherry-pick-approved`, we have process for that 😄

PR #325, opened 2019-10-24
Consume operator config from openshift/api

benjaminapetersen:
Depends on a re-bump with:
- https://github.com/openshift/api/pull/492
to get everything we need.

/hold


benjaminapetersen:
/hold remove

https://github.com/openshift/api/pull/492 merged, bumped, should be good to go.

benjaminapetersen:
/hold cancel

always forget which one :)

smarterclayton:
/hold

This isn’t passing and is in a retry loop

spadgett:
/lgtm cancel

@benjaminapetersen the Dockerfile COPY command needs to be updated since the operator CRD is in a different folder

benjaminapetersen:
/hold remove

This should fix, updated the `COPY` in the `Dockerfile`.

Dockerfile.rhel7
```diff
@@ -14,7 +14,8 @@ COPY --from=builder /tmp/build/console /usr/bin/console
 COPY manifests /manifests/
 
 # extensions manifests generated from openshift/api types
-COPY vendor/github.com/openshift/api/console/v1/*.yaml /manifests
+COPY vendor/github.com/openshift/api/console/v1/*.yaml /manifests/
+COPY vendor/github.com/openshift/api/operator/v1/0000_70_console-operator.crd.yaml /manifests/
```
spadgett: Is Dockerfile.rhel7 still used? I'm wondering if we can remove it and only maintain one Dockerfile.

benjaminapetersen: Good Q.  The `machine-`* operators, `cluster-version` operator, and `api-server` operators still have a `.rhel7` file.  I'm assuming its still in use.

benjaminapetersen:
/retest

`timed out waiting for the condition`.

benjaminapetersen:
@spadgett will do.

spadgett:
@jhadvig FYI, this has the label selector change you added

benjaminapetersen:
I think our new flake is a 100% merge blocker 
@jhadvig race to fix. 😢 😭 

error:
`timed out waiting for the condition`

spadgett:
> I think our new flake is a 100% merge blocker

We should verify it's not something in this PR

PR #326, opened 2019-10-29
Dockerfile destinations must end in trailing slash when COPY multiple files

benjaminapetersen:
Fixes:

```
Step 9/12 : COPY manifests /manifests/
 ---> e755cdd503bb
Step 10/12 : COPY vendor/github.com/openshift/api/console/v1/*.yaml /manifests
When using COPY with more than one source file, the destination must be a directory and end with a /
pushing image...
```

/assign @jhadvig


benjaminapetersen:
/close

Gonna roll this up into #325 since they both have to meddle with the `Dockerfile` `COPY`.

PR #299, opened 2019-09-21
Bug 1764313: 4.3 console_url is no longer reported on metrics endpoint (fix, with e2e test)

benjaminapetersen:
An e2e test to validate that the `/metrics` endpoint on the `operator` responds with an expected output.

PR #327, opened 2019-10-30
Delay registry of console_url metric until starter.go

benjaminapetersen:


benjaminapetersen:
combining this into #299

PR #310, opened 2019-10-08
Reorganize args to NewConsoleOperator func

benjaminapetersen:
Reorganize args to NewConsoleOperator() to fit client/informer sets together.

This will make it easier to extract out components into separate controllers. 

/assign @jhadvig

PR #328, opened 2019-10-31
Bug 1764704: Sync router-ca to the console namespace

jhadvig:
Syncing the `router-ca` from `openshift-config-managed` namespace to `openshift-console` namespace and pasing to the `console` deployment.

Will update the unit tests after the first round of review to avoid unnecessary rewrites.

/assign @benjaminapetersen

### DIFF
--- a/hack/test-unit.sh
+++ b/hack/test-unit.sh
@@ -10,6 +10,7 @@ PACKAGES_TO_TEST=(
     "github.com/openshift/console-operator/pkg/console/operator"
     "github.com/openshift/console-operator/pkg/console/starter"
     "github.com/openshift/console-operator/pkg/console/subresource/configmap"
+    "github.com/openshift/console-operator/pkg/console/subresource/consoleserver"
     "github.com/openshift/console-operator/pkg/console/subresource/deployment"
     "github.com/openshift/console-operator/pkg/console/subresource/oauthclient"
     "github.com/openshift/console-operator/pkg/console/subresource/route"
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -95,6 +95,9 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console, 
 	// The sync loop may not settle, we are unable to honor it in current state.
 	status.HandleProgressingOrDegraded(updatedOperatorConfig, "CustomLogoSync", customLogoErrReason, customLogoError)
 
+	routerCAConfigMap, routerCAErrReason, routerCAError := co.ValidateRouterCAConfigMap()
+	status.HandleProgressingOrDegraded(updatedOperatorConfig, "RouterCAValidation", routerCAErrReason, routerCAError)
+
 	sec, secChanged, secErr := co.SyncSecret(set.Operator)
 	toUpdate = toUpdate || secChanged
 	status.HandleProgressingOrDegraded(updatedOperatorConfig, "OAuthClientSecretSync", "FailedApply", secErr)
@@ -109,7 +112,7 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console, 
 		return oauthErr
 	}
 
-	actualDeployment, depChanged, depErrReason, depErr := co.SyncDeployment(set.Operator, cm, serviceCAConfigMap, trustedCAConfigMap, sec, rt, set.Proxy, customLogoCanMount)
+	actualDeployment, depChanged, depErrReason, depErr := co.SyncDeployment(set.Operator, cm, serviceCAConfigMap, routerCAConfigMap, trustedCAConfigMap, sec, rt, set.Proxy, customLogoCanMount)
 	toUpdate = toUpdate || depChanged
 	status.HandleProgressingOrDegraded(updatedOperatorConfig, "DeploymentSync", depErrReason, depErr)
 	if depErr != nil {
@@ -231,13 +234,14 @@ func (co *consoleOperator) SyncDeployment(
 	operatorConfig *operatorv1.Console,
 	cm *corev1.ConfigMap,
 	serviceCAConfigMap *corev1.ConfigMap,
+	routerCAConfigMap *corev1.ConfigMap,
 	trustedCAConfigMap *corev1.ConfigMap,
 	sec *corev1.Secret,
 	rt *routev1.Route,
 	proxyConfig *configv1.Proxy,
 	canMountCustomLogo bool) (consoleDeployment *appsv1.Deployment, changed bool, reason string, err error) {
 
-	requiredDeployment := deploymentsub.DefaultDeployment(operatorConfig, cm, serviceCAConfigMap, trustedCAConfigMap, sec, rt, proxyConfig, canMountCustomLogo)
+	requiredDeployment := deploymentsub.DefaultDeployment(operatorConfig, cm, serviceCAConfigMap, routerCAConfigMap, trustedCAConfigMap, sec, rt, proxyConfig, canMountCustomLogo)
 	expectedGeneration := getDeploymentGeneration(co)
 	genChanged := operatorConfig.ObjectMeta.Generation != operatorConfig.Status.ObservedGeneration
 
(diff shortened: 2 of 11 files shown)


spadgett:
cc @csrwng @sjenning

pkg/console/subresource/deployment/deployment.go
```diff
@@ -403,6 +406,12 @@ func defaultVolumeConfig() []volumeConfig {
 			path:        "/var/service-ca",
 			isConfigMap: true,
 		},
+		{
+			name:        api.RouterCAConfigMapName,
+			readOnly:    true,
+			path:        "/var/router-ca",
```
spadgett: We need to set this as the `oauthEndpointCAFile` in the console config map.

pkg/console/controllers/router/controller.go
```diff
@@ -0,0 +1,133 @@
...
+	return err
+}
+
+func (c *RouterCASyncController) removeRouterCAConfigMap() error {
+	klog.V(2).Info("deleting router-ca configmap")
+	defer klog.V(2).Info("finished deleting router-ca configmap")
+	return c.configMapClient.ConfigMaps(api.OpenShiftConsoleNamespace).Delete(api.RouterCAConfigMapName, &metav1.DeleteOptions{})
+}
+
+func (c *RouterCASyncController) Run(workers int, stopCh <-chan struct{}) {
```
benjaminapetersen: Looks good. We can choose to use the `boilerplate` package to eliminate some of this repetitive stuff, but I have no qualms with this right now.

pkg/console/controllers/router/controller.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	switch operatorConfig.Spec.ManagementState {
+	case operatorsv1.Managed:
+		klog.V(4).Infoln("console is in a managed state: syncing router-ca configmap")
+	case operatorsv1.Unmanaged:
+		klog.V(4).Infoln("console is in an unmanaged state: skipping router-ca configmap sync")
+		return nil
+	case operatorsv1.Removed:
+		klog.V(4).Infoln("console is in a removed state: deleting router-ca configmap")
```
benjaminapetersen: All this controller to only handle removed 😂

pkg/console/starter/starter.go
```diff
@@ -145,6 +146,8 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 
 	resourceSyncerInformers, resourceSyncer := getResourceSyncer(ctx, clientwrapper.WithoutSecret(kubeClient), operatorClient)
 
+	err = startResourceSyncing(resourceSyncer)
+
```
benjaminapetersen: prob should handle right here:
```golang
if err != nil {
	return err
}
```

jhadvig:
@benjaminapetersen @spadgett comments addressed

jhadvig:
```53) Modal Annotations
A Jasmine spec timed out. Resetting the WebDriver Control Flow.```

/retest

**Comment by jhadvig** [comment]
@benjaminapetersen updated the PR, with couple of nits I've found... PTAL

**Comment by jhadvig** [comment]
@benjaminapetersen updated the PR with additional commit that is adding the fallback to the `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` if the `router-ca` is not synced in the `openshift-console` namespace.
Will update also the unit tests after the review... PTAL

File: pkg/console/operator/sync_v400.go [hunk 5]
```diff
@@ -311,7 +311,18 @@ func (co *consoleOperator) SyncConfigMap(
...
 
-	defaultConfigmap, _, err := configmapsub.DefaultConfigMap(operatorConfig, consoleConfig, managedConfig, infrastructureConfig, consoleRoute)
+	useDefaultRouterCA := true
+	// We are syncing the `router-ca` configmap from `openshift-config-managed` to `openshift-console`.
+	// `router-ca` is only published in `openshift-config-managed` if an operator-generated default certificate is used.
+	// It will not exist if all ingresscontrollers user admin-provided default certificates.
+	// If the `router-ca` configmap in `openshift-console` exist we should mount that to the console container,
+	// otherwise default to `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`
+	_, rcaErr := co.configMapClient.ConfigMaps(api.OpenShiftConsoleNamespace).Get(api.RouterCAConfigMapName, metav1.GetOptions{})
+	if rcaErr != nil && apierrors.IsNotFound(rcaErr) {
```
benjaminapetersen: what about other errors? this only covers 404.

benjaminapetersen: is there anything else that would matter?

pkg/console/subresource/consoleserver/config_builder.go
```diff
@@ -9,8 +9,9 @@ import (
 )
 
 const (
-	clientSecretFilePath    = "/var/oauth-config/clientSecret"
-	oauthEndpointCAFilePath = "/var/router-ca/ca-bundle.crt"
+	clientSecretFilePath           = "/var/oauth-config/clientSecret"
+	oauthEndpointCAFilePath        = "/var/router-ca/ca-bundle.crt"
```
benjaminapetersen: Lets fix the name here as well, `routerCAFIlePath` vs `oauthEndpointCAFilePath`.  
`oauthEndpointCAFilePath` and `defaultOAuthEndpointCAFilePath` sound like the same thing.

pkg/console/subresource/consoleserver/config_builder.go
```diff
@@ -36,6 +37,7 @@ type ConsoleServerCLIConfigBuilder struct {
 	statusPageID      string
 	customProductName string
 	customLogoFile    string
+	routerCA          string
```
benjaminapetersen: This property should prob be `CAFile`, as it may be the router-CA, or it may not.

pkg/console/starter/starter.go
```diff
@@ -193,15 +194,15 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		recorder,
 	)
 
-	// resourceSyncController contains additional locig for all the secrets and
-	// configmaps that we resourceSyncer is taking care of
-	resourceSyncController := resourcesync.NewResourceSyncController(
-		// clients
+	// ResourceSyncDestinationController contains additional locig for all the
```
benjaminapetersen: `locig` -> `logic`

pkg/console/starter/starter.go
```diff
@@ -188,6 +193,19 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		recorder,
 	)
 
+	// resourceSyncController contains additional locig for all the secrets and
+	// configmaps that we resourceSyncer is taking care of
+	resourceSyncController := resourcesync.NewResourceSyncController(
```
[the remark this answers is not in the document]
benjaminapetersen: I updated this recently:
https://github.com/openshift/console-operator/blob/master/pkg/console/starter/starter.go#L149 in #310.

pkg/console/operator/sync_v400.go
jhadvig: was thinking about it and wasn't really sure if any other errors should matter in this context... saying that not really sure what should be the `reason` and `type` of the created condition, if there will be other error type... also if it makes sense to even  create one in this case...

pkg/console/subresource/consoleserver/config_builder.go
jhadvig: yeah, probably it will be better to give the true names with an additional comment

hack/test-unit.sh
```diff
@@ -10,6 +10,7 @@ PACKAGES_TO_TEST=(
     "github.com/openshift/console-operator/pkg/console/operator"
     "github.com/openshift/console-operator/pkg/console/starter"
     "github.com/openshift/console-operator/pkg/console/subresource/configmap"
+    "github.com/openshift/console-operator/pkg/console/subresource/consoleserver"
```
jhadvig: It appears we have been missing this pkg in our unit tests

pkg/console/subresource/consoleserver/config_builder.go
```diff
@@ -119,10 +130,15 @@ func (b *ConsoleServerCLIConfigBuilder) clusterInfo() ClusterInfo {
 }
 
 func (b *ConsoleServerCLIConfigBuilder) authServer() Auth {
+	// we need this fallback due to the way our unit test are structured,
+	// where the ConsoleServerCLIConfigBuilder object is being instantiated empty
```
jhadvig: not a bad thing I thing... It's good to have a fallback here...

hack/test-unit.sh
benjaminapetersen: Not ideal to have to manually add these eh?  Prob should update this script to loop the directory instead, but looks good for now.

pkg/console/operator/sync_v400.go
benjaminapetersen: Maybe we should just add a log and call it good:

```golang
if rcaErr != nil {
  klog.V(4).Infof("router-ca GET error: %s",rcaErr)
}
```
Agree with you, probably doesn't matter, unless there is some funny state we aren't thinking about.  But a `4+` log level dumping some detail may be helpful.

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -68,7 +70,7 @@ apiVersion: console.openshift.io/v1
 auth:
   clientID: console
   clientSecretFile: /var/oauth-config/clientSecret
-  oauthEndpointCAFile: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+  oauthEndpointCAFile: /var/router-ca/ca-bundle.crt
```
benjaminapetersen: I think we may want to add 2 new tests here:

```
name: "Test configmap with router-ca",
name: "Test configmap without router-ca",
```
Probably can duplicate the test `name: "Test default configmap, no customization",` and just swap the bool & output.  This would make it explicit.

pkg/console/subresource/consoleserver/config_builder.go
benjaminapetersen: It can't function correctly without one or the other, correct?  Is it just because of unit test structure, or is the config broken without an entry?

PR #333, opened 2019-10-31
Skip Unmanaged and Removed TCs

jhadvig:
/assign @benjaminapetersen


benjaminapetersen:
/lgtm

so we can merge what we need tomorrow, then we can turn these back on.

PR #334, opened 2019-11-01
Add status handling for RouteHealth

benjaminapetersen:
Most of the "Console isn't happy" questions we get stem from the console route being unhealthy at this point.  This adds a `RouteHealthDegraded` condition, with several checks to see if the route is in a good place.

It is similar to the check done by the authentication operator, which seems to have made debugging much easier:
https://github.com/openshift/cluster-authentication-operator/blob/e7d5e3d45f5188e9c0631e93552e5b7ac48a06e4/pkg/operator2/operator.go#L445

/assign @spadgett @jhadvig

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -31,3 +31,8 @@ const (
 	OCCLIDownloadsCustomResourceName    = "oc-cli-downloads"
 	ODOCLIDownloadsCustomResourceName   = "odo-cli-downloads"
 )
+
+const (
+	RouterCAFilePath        = "/var/router-ca/ca-bundle.crt"
+	OAuthEndpointCAFilePath = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+)
--- a/pkg/console/operator/health_deployment.go
+++ b/pkg/console/operator/health_deployment.go
@@ -1,0 +1,41 @@
+package operator
+
+import (
+	"errors"
+	"fmt"
+	"os"
+
+	deploymentv1 "k8s.io/api/apps/v1"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	"github.com/openshift/console-operator/pkg/console/status"
+	deploymentsub "github.com/openshift/console-operator/pkg/console/subresource/deployment"
+)
+
+func (co *consoleOperator) CheckDeploymentHealth(opConfig *operatorv1.Console, deployment *deploymentv1.Deployment, toUpdate bool) {
+	status.HandleAvailable(func() (conf *operatorv1.Console, prefix string, reason string, err error) {
+		prefix = "Deployment"
+		if !deploymentsub.IsReady(deployment) {
+			return opConfig, prefix, "InsufficientReplicas", errors.New(fmt.Sprintf("%v pods available for console deployment", deployment.Status.ReadyReplicas))
+		}
+		if !deploymentsub.IsReadyAndUpdated(deployment) {
+			return opConfig, prefix, "FailedUpdate", errors.New(fmt.Sprintf("%v replicas ready at version %s", deployment.Status.ReadyReplicas, os.Getenv("RELEASE_VERSION")))
+		}
+		return opConfig, prefix, "", nil
+	}())
+
+	status.HandleProgressing(opConfig, "SyncLoopRefresh", "InProgress", func() error {
+		if toUpdate {
+			return errors.New("Changes made during sync updates, additional sync expected.")
+		}
+		version := os.Getenv("RELEASE_VERSION")
+		if !deploymentsub.IsAvailableAndUpdated(deployment) {
+			return errors.New(fmt.Sprintf("Working toward version %s", version))
+		}
+		if co.versionGetter.GetVersions()["operator"] != version {
+			co.versionGetter.SetVersion("operator", version)
+		}
+		return nil
+	}())
+}
--- a/pkg/console/operator/health_route.go
+++ b/pkg/console/operator/health_route.go
@@ -70,6 +70,8 @@ func (co *consoleOperator) getCA() (*x509.CertPool, error) {
 	//}
 	caCertPool := x509.NewCertPool()
 
+	appendProxyCA(caCertPool, co.systemCABundle)
+
 	routerCA, rcaErr := co.configMapClient.ConfigMaps(api.OpenShiftConsoleNamespace).Get(api.RouterCAConfigMapName, metav1.GetOptions{})
 
 	if rcaErr != nil && apierrors.IsNotFound(rcaErr) {
@@ -84,6 +86,7 @@ func (co *consoleOperator) getCA() (*x509.CertPool, error) {
 	if rcaErr != nil {
 		klog.Infof("failed to GET configmap %s in %s (synced from %s)", api.RouterCAConfigMapName, api.OpenShiftConsoleNamespace, api.OpenShiftConfigManagedNamespace)
 	}
+
 	// fallback to our local ca in from our serviceaccount
 	// if we hit this path, are we likely to get self signed certs errors?
 	serviceAccountCAbytes, err := ioutil.ReadFile(api.OAuthEndpointCAFilePath)
@@ -105,4 +108,13 @@ func clientWithCA(caPool *x509.CertPool) *http.Client {
 			},
 		},
 	}
+}
+
+func appendProxyCA(caCertPool *x509.CertPool, systemCABundle []byte) *x509.CertPool {
+	// load the proxy CA from disk, if we have it
+	if len(systemCABundle) > 0 {
+		// caData = append(bytes.TrimSpace(caData), []byte("\n")...)
+		caCertPool.AppendCertsFromPEM(systemCABundle)
+	}
+	return caCertPool
 }
--- a/pkg/console/operator/health_route.go
+++ b/pkg/console/operator/health_route.go
@@ -63,6 +63,11 @@ func (co *consoleOperator) CheckRouteHealth(opConfig *operatorv1.Console, rt *ro
 }
 
 func (co *consoleOperator) getCA() (*x509.CertPool, error) {
+	// TODO: should I update to this? start with the SystemCertPool?
+	//rootCAs, _ := x509.SystemCertPool()
+	//if rootCAs == nil {
+	//	rootCAs = x509.NewCertPool()
+	//}
 	caCertPool := x509.NewCertPool()
 
 	routerCA, rcaErr := co.configMapClient.ConfigMaps(api.OpenShiftConsoleNamespace).Get(api.RouterCAConfigMapName, metav1.GetOptions{})
@@ -93,6 +98,7 @@ func (co *consoleOperator) getCA() (*x509.CertPool, error) {
 func clientWithCA(caPool *x509.CertPool) *http.Client {
 	return &http.Client{
 		Timeout: 5 * time.Second,
+		// TODO: do I need http.DefaultTransport.(*http.Transport)?
 		Transport: &http.Transport{
 			TLSClientConfig: &tls.Config{
 				RootCAs: caPool,
--- a/pkg/console/operator/health_route.go
+++ b/pkg/console/operator/health_route.go
@@ -1,0 +1,102 @@
+package operator
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"time"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	routev1 "github.com/openshift/api/route/v1"
+
+	"github.com/openshift/console-operator/pkg/api"
+	"github.com/openshift/console-operator/pkg/console/status"
+	routesub "github.com/openshift/console-operator/pkg/console/subresource/route"
+)
+
+const (
+	routerCABundleKey = "ca-bundle.crt"
+)
+
+func (co *consoleOperator) CheckRouteHealth(opConfig *operatorv1.Console, rt *routev1.Route) {
+	status.HandleDegraded(func() (conf *operatorv1.Console, prefix string, reason string, err error) {
+		prefix = "RouteHealth"
+
+		caPool, err := co.getCA()
+		if err != nil {
+			return opConfig, prefix, "FailedLoadCA", fmt.Errorf("failed to read CA to check route health: %v", err)
+		}
+		client := clientWithCA(caPool)
+
+		// TODO: deal with an environment with a MitM proxy?
+		url := "https://" + rt.Spec.Host + "/health"
+		req, err := http.NewRequest(http.MethodGet, url, nil)
+		if err != nil {
+			return opConfig, prefix, "FailedRequest", fmt.Errorf("failed to build request to route (%s): %v", url, err)
+		}
+		resp, err := client.Do(req)
+		if err != nil {
+			return opConfig, prefix, "FailedGet", fmt.Errorf("failed to GET route (%s): %v", url, err)
+		}
+		defer resp.Body.Close()
+
+		if resp.StatusCode != http.StatusOK {
+			return opConfig, prefix, "StatusError", fmt.Errorf("route not yet available, %s returns '%s'", url, resp.Status)
+
+		}
+		return opConfig, prefix, "", nil
+	}())
+
+	status.HandleAvailable(opConfig, "Route", "FailedAdmittedIngress", func() error {
+		if !routesub.IsAdmitted(rt) {
+			return errors.New("console route is not admitted")
+		}
+		return nil
+	}())
+}
+
+func (co *consoleOperator) getCA() (*x509.CertPool, error) {
+	caCertPool := x509.NewCertPool()
+
+	routerCA, rcaErr := co.configMapClient.ConfigMaps(api.OpenShiftConsoleNamespace).Get(api.RouterCAConfigMapName, metav1.GetOptions{})
+
+	if rcaErr != nil && apierrors.IsNotFound(rcaErr) {
+		//  using CA ca-bundle.crt from configmap router-ca from openshift-config-managed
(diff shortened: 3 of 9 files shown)


pkg/console/operator/health.go
```diff
@@ -0,0 +1,85 @@
...
+	deploymentv1 "k8s.io/api/apps/v1"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	routev1 "github.com/openshift/api/route/v1"
+
+	"github.com/openshift/console-operator/pkg/console/status"
+	deploymentsub "github.com/openshift/console-operator/pkg/console/subresource/deployment"
+	routesub "github.com/openshift/console-operator/pkg/console/subresource/route"
+)
+
```
benjaminapetersen: Pulling these out into separate funcs so we can call them at other points in the sync loop.  For example, we can check route health much earlier.

pkg/console/operator/sync_v400.go
```diff
@@ -64,6 +63,7 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console,
 	rt, rtChanged, rtErrReason, rtErr := co.SyncRoute(set.Operator)
 	toUpdate = toUpdate || rtChanged
 	status.HandleProgressingOrDegraded(updatedOperatorConfig, "RouteSync", rtErrReason, rtErr)
+	co.CheckRouteHealth(updatedOperatorConfig, rt)
```
benjaminapetersen: We can check route health right away, rather than waiting for the whole loop to complete. This should let us aggregate status.

pkg/console/operator/sync_v400.go
```diff
@@ -112,6 +112,7 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console,
 	actualDeployment, depChanged, depErrReason, depErr := co.SyncDeployment(set.Operator, cm, serviceCAConfigMap, trustedCAConfigMap, sec, rt, set.Proxy, customLogoCanMount)
 	toUpdate = toUpdate || depChanged
 	status.HandleProgressingOrDegraded(updatedOperatorConfig, "DeploymentSync", depErrReason, depErr)
+	co.CheckDeploymentHealth(updatedOperatorConfig, actualDeployment, toUpdate)
```
benjaminapetersen: Deployment health is still last.

benjaminapetersen:
Revised per @spadgett comment about ordering.  Def a good point, we want to be able to check health even if other resources `err` and abort the loop.

benjaminapetersen:
/retest

Variety of flakes

pkg/console/operator/health.go
jhadvig: > Pulling these out into separate funcs so we can call them at other points in the sync loop

wondering if that will be true even after we refact our big sync loop into controllers ?

pkg/console/operator/health.go
```diff
@@ -0,0 +1,85 @@
...
+		if !deploymentsub.IsReady(deployment) {
+			return opConfig, prefix, "InsufficientReplicas", errors.New(fmt.Sprintf("%v pods available for console deployment", deployment.Status.ReadyReplicas))
+		}
+		if !deploymentsub.IsReadyAndUpdated(deployment) {
+			return opConfig, prefix, "FailedUpdate", errors.New(fmt.Sprintf("%v replicas ready at version %s", deployment.Status.ReadyReplicas, os.Getenv("RELEASE_VERSION")))
+		}
+		return opConfig, prefix, "", nil
+	}())
+
+	status.HandleAvailable(func() (conf *operatorv1.Console, prefix string, reason string, err error) {
```
jhadvig: This whole block is a dup. of the `status.HandleAvailable()` above

pkg/console/operator/health.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+		if !deploymentsub.IsReadyAndUpdated(deployment) {
+			return opConfig, prefix, "FailedUpdate", errors.New(fmt.Sprintf("%v replicas ready at version %s", deployment.Status.ReadyReplicas, os.Getenv("RELEASE_VERSION")))
+		}
+		return opConfig, prefix, "", nil
+	}())
+
+	status.HandleProgressing(opConfig, "SyncLoopRefresh", "InProgress", func() error {
+		if toUpdate {
+			return errors.New("Changes made during sync updates, additional sync expected.")
```
jhadvig: Thinking if it wouldn't make sense to also mention the resource, just to have easier debugging.

```suggestion
			return errors.New("Deployment changes made during sync updates, additional sync expected.")
```

pkg/console/operator/health.go
benjaminapetersen: Prob will not matter at that point.  But I think we might want this now (and maybe backport).

benjaminapetersen:
Interestingly, test failure:

```
brand_builtins_test.go:50: operator has not reached settled state in 4m0s attempts due to [RouteHealthDegraded] - timed out waiting for the condition 
```

pkg/console/operator/health.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/openshift/console-operator/pkg/console/status"
+	deploymentsub "github.com/openshift/console-operator/pkg/console/subresource/deployment"
+	routesub "github.com/openshift/console-operator/pkg/console/subresource/route"
+)
+
+func (co *consoleOperator) CheckRouteHealth(opConfig *operatorv1.Console, rt *routev1.Route) {
+	status.HandleDegraded(func() (conf *operatorv1.Console, prefix string, reason string, err error) {
+		prefix = "RouteHealth"
+		client := http.Client{Timeout: time.Duration(5 * time.Second)}
+		req, err := http.NewRequest(http.MethodHead, "https://"+rt.Spec.Host+"/health", nil)
```
benjaminapetersen:
```
  - lastTransitionTime: "2019-11-04T20:21:52Z"
    message: 'RouteHealthDegraded: failed to GET route: Head https://console-openshift-console.apps.bpetersen.devcluster.openshift.com/health:
      x509: certificate signed by unknown authority'
    reason: RouteHealthDegradedFailedGet
    status: "True"
    type: Degraded
```
May need to have a way to ignore self signed certs here.

benjaminapetersen: Rebasing to see of the CA changes resolved this.

pkg/console/operator/health.go
benjaminapetersen: Removed.

pkg/console/operator/health.go
benjaminapetersen: The `toUpdate` bool though could be any of the resources, it may be something else (that may or may not eventually cause the deployment to roll)

benjaminapetersen: That first `if toUpdate` is a little bit unfortunate in that way, but hopefully all of our other statuses are clear enough that we indicate what is going on.

benjaminapetersen:
/retest 

```
     logging_test.go:25: setting console operator to 'Debug' LogLevel ...
    logging_test.go:83: checking if '--log-level=*=DEBUG' flag is set on the console deployment container command...
    logging_test.go:16: waiting for cleanup to reach settled state...
    console-operator.go:320: waiting for observed generation 14 to match generation 15... 
    console-operator.go:369: waited 10 seconds to reach settled state...
    console-operator.go:369: waited 30 seconds to reach settled state...
    console-operator.go:369: waited 60 seconds to reach settled state...
    console-operator.go:369: waited 90 seconds to reach settled state...
    console-operator.go:369: waited 120 seconds to reach settled state...
    logging_test.go:16: operator has not reached settled state in 4m0s attempts due to [RouteHealthDegraded] - timed out waiting for the condition 
```

I want that status :)
But retest to see if this is a flake or a real deal.

benjaminapetersen:
/retest

`level=fatal msg="Terraform destroy: failed to destroy using Terraform"`

pkg/console/operator/health.go
benjaminapetersen: Have to pipe through the certs.  Working on this.

pkg/console/operator/health.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/openshift/console-operator/pkg/console/status"
+	deploymentsub "github.com/openshift/console-operator/pkg/console/subresource/deployment"
+	routesub "github.com/openshift/console-operator/pkg/console/subresource/route"
+)
+
+func (co *consoleOperator) CheckRouteHealth(opConfig *operatorv1.Console, rt *routev1.Route) {
+	status.HandleDegraded(func() (conf *operatorv1.Console, prefix string, reason string, err error) {
+		prefix = "RouteHealth"
+		client := http.Client{Timeout: time.Duration(5 * time.Second)}
```
stlaz: You will need the router CA to trust the router certs, get it at the CM `openshift-config-managed/router-ca`, grab the contents of the `ca-bundle.crt` key. Use them in your client's `Transport.TLSClientTransport.RootCAs`

stlaz: That's going to be the first step, you will need to also consider the operator being run in an environment with an MitM proxy

pkg/api/api.go
```diff
@@ -31,3 +31,8 @@ const (
 	OCCLIDownloadsCustomResourceName    = "oc-cli-downloads"
 	ODOCLIDownloadsCustomResourceName   = "odo-cli-downloads"
 )
+
+const (
```
benjaminapetersen: promoted these up since I need the paths.

pkg/console/operator/health_deployment.go
```diff
@@ -0,0 +1,41 @@
...
+	"os"
+
+	deploymentv1 "k8s.io/api/apps/v1"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	"github.com/openshift/console-operator/pkg/console/status"
+	deploymentsub "github.com/openshift/console-operator/pkg/console/subresource/deployment"
+)
+
```
benjaminapetersen: separated this out so that helpers for each check can be co-located with the status they support.

pkg/console/operator/health_route.go
```diff
@@ -0,0 +1,102 @@
...
+	status.HandleDegraded(func() (conf *operatorv1.Console, prefix string, reason string, err error) {
+		prefix = "RouteHealth"
+
+		caPool, err := co.getCA()
+		if err != nil {
+			return opConfig, prefix, "FailedLoadCA", fmt.Errorf("failed to read CA to check route health: %v", err)
+		}
+		client := clientWithCA(caPool)
+
+		// TODO: deal with an environment with a MitM proxy?
```
benjaminapetersen: @stlaz got the CA cert from `router-ca`, and a fallback to use the local `serviceaccount/ca` from the operator, if the `router-ca` is unavailable.  You mentioned a possible `proxy` MitM issue I should also take into account.  Care to elaborate? Thanks!

pkg/console/operator/health_route.go
```diff
@@ -0,0 +1,102 @@
...
+		url := "https://" + rt.Spec.Host + "/health"
+		req, err := http.NewRequest(http.MethodGet, url, nil)
```
benjaminapetersen: Note that `console` does not support a `httpMethodHead` request.

pkg/console/operator/health_route.go
```diff
@@ -0,0 +1,102 @@
...
+	}())
+}
+
+func (co *consoleOperator) getCA() (*x509.CertPool, error) {
+	caCertPool := x509.NewCertPool()
+
+	routerCA, rcaErr := co.configMapClient.ConfigMaps(api.OpenShiftConsoleNamespace).Get(api.RouterCAConfigMapName, metav1.GetOptions{})
+
+	if rcaErr != nil && apierrors.IsNotFound(rcaErr) {
+		//  using CA ca-bundle.crt from configmap router-ca from openshift-config-managed
```
benjaminapetersen: On error, this does not log any cert data, simply logs path (at high enough `-v`) of attempt to read cert (also, if we fail, the cert was not read, anyway).    We want to know exactly what went wrong, if it goes wrong, I think, without leaking anything.

pkg/console/operator/health.go
benjaminapetersen: Got this, see update & left you a comment regarding `MitM`.

benjaminapetersen:
No longer getting this issue with the CA wired up:

```
  - lastTransitionTime: "2019-11-04T20:21:52Z"
    message: 'RouteHealthDegraded: failed to GET route: Head https://console-openshift-console.apps.bpetersen.devcluster.openshift.com/health:
      x509: certificate signed by unknown authority'
    reason: RouteHealthDegradedFailedGet
    status: "True"
    type: Degraded
```
Flow now may report:

```
  - lastTransitionTime: "2019-11-15T18:48:57Z"
    message: 'RouteHealthDegraded: route not yet available, https://console-openshift-console.apps.bpetersen.devcluster.openshift.com/health
      returns ''503 Service Unavailable'''
```
then:
```
- lastTransitionTime: "2019-11-15T18:48:57Z"
    message: 'RouteHealthDegraded: route not yet available, https://console-openshift-console.apps.bpetersen.devcluster.openshift.com/health
      returns ''503 Service Unavailable'''
    reason: RouteHealthDegradedStatusError
```
until finally resolved when the endpoint responds:

```
status:
  conditions:
  - lastTransitionTime: "2019-11-15T19:04:20Z"
    reason: AsExpected
    status: "True"
    type: Available
  - lastTransitionTime: "2019-11-15T19:04:18Z"
    reason: AsExpected
    status: "False"
    type: Progressing
  - lastTransitionTime: "2019-11-15T19:04:18Z"
    reason: AsExpected
    status: "False"
    type: Degraded
  - lastTransitionTime: "2019-11-15T16:15:43Z"
    reason: AsExpected
    status: "True"
    type: Upgradeable
```

pkg/console/operator/health_route.go
```diff
@@ -0,0 +1,102 @@
...
+	status.HandleAvailable(opConfig, "Route", "FailedAdmittedIngress", func() error {
+		if !routesub.IsAdmitted(rt) {
+			return errors.New("console route is not admitted")
+		}
+		return nil
+	}())
+}
+
+func (co *consoleOperator) getCA() (*x509.CertPool, error) {
+	caCertPool := x509.NewCertPool()
```
benjaminapetersen: Possibly i should update to:

```golang
// start with system, fallback to new?
rootCAs, _ := x509.SystemCertPool()
if rootCAs == nil {
	rootCAs = x509.NewCertPool()
}
```

pkg/console/operator/health_route.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	klog.V(4).Infof("using CA on disk from %s", api.OAuthEndpointCAFilePath)
+	caCertPool.AppendCertsFromPEM(serviceAccountCAbytes)
+	return caCertPool, nil
+}
+
+func clientWithCA(caPool *x509.CertPool) *http.Client {
+	return &http.Client{
+		Timeout: 5 * time.Second,
+		Transport: &http.Transport{
```
benjaminapetersen: Possibly should use:

```golang
transport := http.DefaultTransport()
```

benjaminapetersen:
Also needs #334 to eliminate the `Removed` test flake.

pkg/console/operator/health_route.go
```diff
@@ -0,0 +1,120 @@
...
+		// TODO: do I need http.DefaultTransport.(*http.Transport)?
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				RootCAs: caPool,
+			},
+		},
+	}
+}
+
+func appendProxyCA(caCertPool *x509.CertPool, systemCABundle []byte) *x509.CertPool {
```
stlaz: If proxy is configured, it should already appear in the system CA bundle - the location you're copying it to is where golang looks when you call `x509.SystemCertPool()`

stlaz: ah, nvm, it was just the name of the function that confused me, I see what you did there

pkg/console/operator/health_route.go
```diff
@@ -0,0 +1,120 @@
...
+	status.HandleAvailable(opConfig, "Route", "FailedAdmittedIngress", func() error {
+		if !routesub.IsAdmitted(rt) {
+			return errors.New("console route is not admitted")
+		}
+		return nil
+	}())
+}
+
+func (co *consoleOperator) getCA() (*x509.CertPool, error) {
+	// TODO: should I update to this? start with the SystemCertPool?
```
stlaz: yes, this looks better IMO, and `SystemCertPool()` should be fresh since you're doing `--terminate-on-files` when starting your operator

pkg/console/operator/health_route.go
```diff
@@ -0,0 +1,120 @@
...
+		return caCertPool, nil
+	}
+	// this error is unexpected
+	if rcaErr != nil {
+		klog.Infof("failed to GET configmap %s in %s (synced from %s)", api.RouterCAConfigMapName, api.OpenShiftConsoleNamespace, api.OpenShiftConfigManagedNamespace)
+	}
+
+	// fallback to our local ca in from our serviceaccount
+	// if we hit this path, are we likely to get self signed certs errors?
+	serviceAccountCAbytes, err := ioutil.ReadFile(api.OAuthEndpointCAFilePath)
```
stlaz: This CA bundle actually does contain the router-ca, so you might just stick with it instead of getting the CA from the config map above.

pkg/console/operator/health_route.go
stlaz: You're using your own roots whereas the above transport only uses the default system trust bundle, so you will want to use your own transport.

test/e2e/metrics_test.go
```diff
@@ -155,6 +155,10 @@ func getClientWithCertAuth(t *testing.T) *http.Client {
 	// that are missing from the system trust roots
 	rootCAs.AppendCertsFromPEM(config.CAData)
 
+	// TODO: could get router-ca ConfigMap we sync'd from openshift-config-managed
```
stlaz: Not really, you're using a passthrough route to access the /metrics endpoint from an external endpoint, but the cert that's used for serving directly at the pod which terminates the TLS connection is not valid for the hostname of the route (and is actually signed by service-ca AFAIK).

So even if you had a valid CA that signed the server cert here, the cert verification would fail because of the hostname. Yet you need passthrough route to be able to use client cert auth. Another solution would be to grab a token for a service account that's allowed to access this endpoint, in which case the passthrough aspect of the route would not be necessary.

jhadvig:
CLosing in favor of https://github.com/openshift/console-operator/pull/399

PR #336, opened 2019-11-04
Revert "Skip Unmanaged and Removed TCs"

benjaminapetersen:
Reverts openshift/console-operator#333

We want to turn our tests back on sooner than later.  Opening this up while we dig into the flake.

@jhadvig


jhadvig:
Closing in favour of https://github.com/openshift/console-operator/pull/338 which contains the fix for the failing tests and also removes the skipping of the e2e tests.

PR #329, opened 2019-10-31
Call StandardSetup func in unmanaged e2e tests

jhadvig:
/assign @benjaminapetersen

PR #338, opened 2019-11-05
Missing service informer

jhadvig:
The issue of the failing Unmanaged e2e tests was that the new ServiceController haven't picked up changes in the consoleoperator config, eg: change from `Unmanaged` to `Managed` state.
Due to this the `console` service has not been recreated and the console pods could not start since cause of that the `console-serving-cert` secret was missing.

Also adding the `StandardSetup` call from https://github.com/openshift/console-operator/pull/329, which this PR shoud replace.

Also removing the skipping of the failing tests, which should replace https://github.com/openshift/console-operator/pull/336 since there is no point to have two PR for that.

/assign @benjaminapetersen

### DIFF
--- a/pkg/console/starter/starter.go
+++ b/pkg/console/starter/starter.go
@@ -189,11 +189,12 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 	)
 
 	consoleServiceController := service.NewServiceSyncController(
-		// operator config so we can update status
-		operatorConfigClient.OperatorV1().Consoles(),
-		// only needs to interact with the service resource
-		kubeClient.CoreV1(),
-		kubeInformersNamespaced.Core().V1().Services(),
+		// clients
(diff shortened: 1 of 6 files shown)


pkg/console/starter/starter.go
```diff
@@ -189,11 +189,12 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 	)
 
 	consoleServiceController := service.NewServiceSyncController(
-		// operator config so we can update status
-		operatorConfigClient.OperatorV1().Consoles(),
-		// only needs to interact with the service resource
-		kubeClient.CoreV1(),
-		kubeInformersNamespaced.Core().V1().Services(),
+		// clients
+		operatorConfigClient.OperatorV1().Consoles(), // operator config so we can update status
```
benjaminapetersen: Nice, mind pairing them up?

```
OperatorClient
OperatorInformer
ServiceClient
ServiceInformer
```

PR #340, opened 2019-11-05
Unify testing resources in e2e tests

jhadvig:
We haven't been consistent with the resources that we are `GET`ting and `DELETE`ing

/assign @benjaminapetersen

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -35,6 +35,17 @@ type TestingResource struct {
 	namespace string
 }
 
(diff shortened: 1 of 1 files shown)


jhadvig:
@benjaminapetersen had to update the PR. Forgot that we dont test availability of the `console-public` CM in the `openshift-config-managed` NS, in the `TestRemoved` and `TestUnamanaged`since we are only syncing the CM's data.
Will probably create a separate TC for each managemenet state of the operator, but now this should reflect the current state of the e2e tests.
PTAL

PR #342, opened 2019-11-06
Remove proxyConfig NotFound check

jhadvig:
If there is a `err`, then we should error out...

/assign @beanh66 

@spadgett FYI

### DIFF
--- a/pkg/console/operator/operator.go
+++ b/pkg/console/operator/operator.go
@@ -185,10 +185,8 @@ func (c *consoleOperator) Sync(obj metav1.Object) error {
 
 	proxyConfig, err := c.proxyConfigClient.Get(api.ConfigResourceName, metav1.GetOptions{})
 	if err != nil {
-		if !errors.IsNotFound(err) {
-			return err
-		}
-		proxyConfig = nil
+		klog.Errorf("proxy config error: %v", err)
+		return err
 	}
 
 	configs := configSet{


pkg/console/operator/operator.go
```diff
@@ -185,10 +185,7 @@ func (c *consoleOperator) Sync(obj metav1.Object) error {
 
 	proxyConfig, err := c.proxyConfigClient.Get(api.ConfigResourceName, metav1.GetOptions{})
 	if err != nil {
-		if !errors.IsNotFound(err) {
-			return err
-		}
-		proxyConfig = nil
+		return err
```
benjaminapetersen: We should still `klog.Errorf("something")` correct?

jhadvig:
@benjaminapetersen comment addressed.... PTAL

jhadvig:
```
Installing from release registry.svc.ci.openshift.org/ci-op-qtjwi2mx/release@sha256:8297f4d0a8b843237d3624cdc71f3d1efdd956cadbcf8cd06eb6fdd589b19302
level=fatal msg="failed to fetch Terraform Variables: failed to fetch dependency of \"Terraform Variables\": failed to fetch dependency of \"Bootstrap Ignition Config\": failed to fetch dependency of \"Master Machines\": failed to generate asset \"Platform Credentials Check\": validate AWS credentials: checking install permissions: error simulating policy: Throttling: Rate exceeded\n\tstatus code: 400, request id: 1e1b33a7-3ebc-4bf1-84e9-6989c3128c46"
```

/retest

spadgett:
@jhadvig we also need to be always mounting the trusted ca config map

PR #343, opened 2019-11-08
Enable multi-arch support in downloads

yselkowitz:
In order for console-operator deployment to succeed, cli-artifacts needs to be available on all arches for downloads-openshift-console. However, in that case, /usr/bin/oc (inherited from cli) is a native binary, and we want to provide all primary Linux architectures to match those on mirror.openshift.com, regardless of cluster architecture.

This depends on https://github.com/openshift/oc/pull/153 to provide the new binaries in cli-artifacts.

Because this affects cluster deployment, we are looking to get this into 4.2 and later branches (we can provide separate PRs for stable branches as needed).

### DIFF
--- a/manifests/07-downloads-deployment.yaml
+++ b/manifests/07-downloads-deployment.yaml
@@ -89,11 +89,18 @@ spec:
               os.mkdir(arch)
               for operating_system in ['linux', 'mac', 'windows']:
                   os.mkdir(os.path.join(arch, operating_system))
+          for arch in ['arm64', 'ppc64le', 's390x']:
+              os.mkdir(arch)
+              for operating_system in ['linux']:
+                  os.mkdir(os.path.join(arch, operating_system))
 
           for arch, operating_system, path in [
-                  ('amd64', 'linux', '/usr/bin/oc'),
+                  ('amd64', 'linux', '/usr/share/openshift/linux_amd64/oc'),
                   ('amd64', 'mac', '/usr/share/openshift/mac/oc'),
                   ('amd64', 'windows', '/usr/share/openshift/windows/oc.exe'),
+                  ('arm64', 'linux', '/usr/share/openshift/linux_arm64/oc'),
+                  ('ppc64le', 'linux', '/usr/share/openshift/linux_ppc64le/oc'),
+                  ('s390x', 'linux', '/usr/share/openshift/linux_s390x/oc'),
                   ]:
               basename = os.path.basename(path)
               target_path = os.path.join(arch, operating_system, basename)
--- a/pkg/console/controllers/clidownloads/controller.go
+++ b/pkg/console/controllers/clidownloads/controller.go
@@ -162,7 +162,7 @@ func PlatformBasedOCConsoleCLIDownloads(host, cliDownloadsName string) *v1.Conso
 		{"Linux for IBM Power, little endian", "ppc64le/linux", "oc.tar"},
 		{"Linux for IBM Z", "s390x/linux", "oc.tar"},
 		{"Mac", "amd64/mac", "oc.zip"},
-		{"Windows", "amd64/windows", "oc.zip"},
+		{"Windows 64-bit", "amd64/windows", "oc.zip"},
 	}
 
 	links := []v1.CLIDownloadLink{}
--- a/pkg/console/controllers/clidownloads/controller.go
+++ b/pkg/console/controllers/clidownloads/controller.go
@@ -122,7 +122,7 @@ func (c *CLIDownloadsSyncController) sync() error {
 	}
 
 	host := routesub.GetCanonicalHost(consoleRoute)
-	ocConsoleCLIDownloads := PlatformBasedOCConsoleCLIDownloads(host, "amd64", api.OCCLIDownloadsCustomResourceName)
+	ocConsoleCLIDownloads := PlatformBasedOCConsoleCLIDownloads(host, api.OCCLIDownloadsCustomResourceName)
 	_, ocCLIDownloadsErrReason, ocCLIDownloadsErr := ApplyCLIDownloads(c.consoleCliDownloadsClient, ocConsoleCLIDownloads)
 	status.HandleDegraded(updatedOperatorConfig, "OCDownloadsSync", ocCLIDownloadsErrReason, ocCLIDownloadsErr)
 	if ocCLIDownloadsErr != nil {
@@ -150,16 +150,19 @@ func GetPlatformURL(baseURL, platform, archiveType string) string {
 	return fmt.Sprintf("%s/%s/%s", baseURL, platform, archiveType)
 }
 
-func PlatformBasedOCConsoleCLIDownloads(host, arch, cliDownloadsName string) *v1.ConsoleCLIDownload {
-	baseURL := fmt.Sprintf("%s/%s", util.HTTPS(host), arch)
+func PlatformBasedOCConsoleCLIDownloads(host, cliDownloadsName string) *v1.ConsoleCLIDownload {
+	baseURL := fmt.Sprintf("%s", util.HTTPS(host))
 	platforms := []struct {
 		label    string
 		key      string
 		archType string
 	}{
-		{"Linux", "linux", "oc.tar"},
-		{"Mac", "mac", "oc.zip"},
-		{"Windows", "windows", "oc.zip"},
+		{"Linux for x86_64", "amd64/linux", "oc.tar"},
+		{"Linux for ARM 64", "arm64/linux", "oc.tar"},
+		{"Linux for IBM Power, little endian", "ppc64le/linux", "oc.tar"},
+		{"Linux for IBM Z", "s390x/linux", "oc.tar"},
+		{"Mac", "amd64/mac", "oc.zip"},
+		{"Windows", "amd64/windows", "oc.zip"},
 	}
 
 	links := []v1.CLIDownloadLink{}
--- a/pkg/console/controllers/clidownloads/controller_test.go
+++ b/pkg/console/controllers/clidownloads/controller_test.go
@@ -20,13 +20,40 @@ func TestGetPlatformURL(t *testing.T) {
 		want string
 	}{
 		{
-			name: "Test assembling linux specific URL",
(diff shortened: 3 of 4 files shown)


yselkowitz:
/hold
until the cli-artifacts change is merged.

manifests/07-downloads-deployment.yaml
```diff
@@ -89,11 +89,18 @@ spec:
               os.mkdir(arch)
               for operating_system in ['linux', 'mac', 'windows']:
                   os.mkdir(os.path.join(arch, operating_system))
+          for arch in ['arm64', 'ppc64le', 's390x']:
```
benjaminapetersen: To make this usable you would also need to update the the generated downloads links:
https://github.com/openshift/console-operator/blob/master/pkg/console/controllers/clidownloads/controller.go#L155
This will update the links in the console itself.

benjaminapetersen: This is when it would be ideal to have 1 place for this data.  

I'm tempted to suggest we would add this annotation to the deployment:

```yaml
  annotations:
    release.openshift.io/create-only: "true"
```
And then cut the `spec` and add from within the operator itself.  This would let the operator generate both the `cliDownloadsLinks` CR & this script from the same source struct.

@spadgett @wking not sure if there are any objections to further work in this deployment.  Personally I'm not super jazzed about adding logic for string manipulation, since this script is just a bit embedded string, but I don't really love the manual sync of multiple resources either.

spadgett: If the script is static, I'd keep it as a manifest. I'd only generate the downloads deployment in the operator if we need some logic, for instance, if we decide to make the node selector customizable.

cc @bparees

benjaminapetersen: Its probably time to add an `e2e` test to this and validate that an `http` request to each of these endpoints is valid.

benjaminapetersen: For docs purposes, the old link to the mirror is gone:
https://github.com/openshift/console/pull/2951/files#diff-f2e8b14c70b3c8bf99b5eed3b019db60L46
We no longer hard-code this:
https://github.com/openshift/console/pull/2951/files#diff-20ab917f3e7dc4814ce688ea3e52044eL67
Since we now do what I referenced above with the `CLIDownloadsController`.

wking:
/ok-to-test

Pretty sure downloads will die without a full cli-artifacts to work from, but let's see...

benjaminapetersen:
errors in code:

```
# github.com/openshift/console-operator/pkg/console/controllers/clidownloads [github.com/openshift/console-operator/pkg/console/controllers/clidownloads.test]
pkg/console/controllers/clidownloads/controller_test.go:107:60: too many arguments in call to PlatformBasedOCConsoleCLIDownloads
have (string, string, string)
want (string, string)
```
Will need to fix before this will run.

benjaminapetersen:
Opened #344 to add tests for our `oc downloads`.  I'd like to see some test coverage merge first, then this PR can update.

benjaminapetersen:
/retest

`timed out waiting for the condition`
lots of timeouts, likely a flake.

yselkowitz:
Repushed with corresponding changes for the unit test.

benjaminapetersen:
E2E merged:
https://github.com/openshift/console-operator/pull/344
You can now update this and we should be good to move forward

pkg/console/controllers/clidownloads/controller.go
```diff
@@ -150,16 +150,19 @@ func GetPlatformURL(baseURL, platform, archiveType string) string {
 	return fmt.Sprintf("%s/%s/%s", baseURL, platform, archiveType)
 }
 
-func PlatformBasedOCConsoleCLIDownloads(host, arch, cliDownloadsName string) *v1.ConsoleCLIDownload {
-	baseURL := fmt.Sprintf("%s/%s", util.HTTPS(host), arch)
+func PlatformBasedOCConsoleCLIDownloads(host, cliDownloadsName string) *v1.ConsoleCLIDownload {
```
benjaminapetersen: This is good, but will break the e2e now:

https://github.com/openshift/console-operator/blob/master/test/e2e/downloads_test.go#L34

Just need to update that to not pass the `arch` arg.

benjaminapetersen:
/retest

expected e2e func to break.

benjaminapetersen:
/retest e2e-aws-operator

Expected e2e func to be broken.

yselkowitz:
/hold cancel
https://github.com/openshift/oc/pull/153 has been merged

manifests/07-downloads-deployment.yaml
wking: With openshift/oc#168's uniform paths, I think we want to lean on the consistency with something like:

```python
for arch in ['amd64', 'arm64', 'ppc64le', 's390x']:
[   os.mkdir(arch)
    for operating_system in ['linux', 'darwin', 'windows']:
        os_path = os.path.join(arch, operating_system)  # might want to translate darwin -> mac here, or we can just link to darwin paths from the console
        os.mkdir(os_path)
        basename = 'oc'
        if operating_system == 'windows':
            basename += '.exe'
        target_path = os.path.join(os_path, basename)
        source_path = os.path.join('/usr', 'share', 'openshift', '{}_{}'.format(operating_system, arch), executable)
        try:
            os.stat(source_path)
        except OSError as error:
            print(error)
            continue
        os.symlink(source_path, target_path)
        ...

File: pkg/console/controllers/clidownloads/controller.go [hunk 3]
```diff
@@ -150,16 +150,19 @@ func GetPlatformURL(baseURL, platform, archiveType string) string {
...
 	}{
-		{"Linux", "linux", "oc.tar"},
-		{"Mac", "mac", "oc.zip"},
-		{"Windows", "windows", "oc.zip"},
+		{"Linux for X86-64", "amd64/linux", "oc.tar"},
+		{"Linux for ARM 64", "arm64/linux", "oc.tar"},
+		{"Linux for Power LE", "ppc64le/linux", "oc.tar"},
+		{"Linux for Z", "s390x/linux", "oc.tar"},
+		{"Mac", "amd64/mac", "oc.zip"},
+		{"Windows", "amd64/windows", "oc.zip"},
```
wking: I'd prefer if these were "for amd64" or "for x86-64" or whatever Red Hat's preferred name for the architecture is instead of leaving it implicit.  Incidentally, [Wikipedia uses x86-64][1], not X86-64.

[1]: https://en.wikipedia.org/wiki/X86-64

manifests/07-downloads-deployment.yaml
yselkowitz: Mac and Windows still only cover one architecture, but that treats them as if they might cover all of them with an error if it (not surprisingly) does not.  That seems like a lot of overhead for a result for which we already know the answer.

pkg/console/controllers/clidownloads/controller.go
yselkowitz: By "implicit" are you referring to "Mac" and "Windows"?  My patch does not change the name of those, and doing so might be more appropriate for a separate PR.

yselkowitz: Repushed with (hopefully) more "offical" names for the Linux architectures (not that there is ever any consistency with these names).

wking: > By "implicit" are you referring to "Mac" and "Windows"?

Yup.

> My patch does not change the name of those, and doing so might be more appropriate for a separate PR.

I'm not a maintainer here, but "add qualifying arch names to download links for all existing downloads" seems in-scope for a PR that adds multi-arch support ;).

manifests/07-downloads-deployment.yaml
wking: > Mac and Windows still only cover one architecture, but that treats them as if they might cover all of them with an error if it (not surprisingly) does not.

A warning, not an error.

> That seems like a lot of overhead for a result for which we already know the answer.

We don't know the answer, it depends on which cli-artifacts gets used for this deployment.  Maybe later they do grow additional architectures, or maybe we drop an arch, or grow a new OS, or whatever.  I don't think this deployment should be opinionated.  It should just look in the places that *might* be there, serve what it has, and 404 requests for content it doesn't have.

yselkowitz: Is it really possible to have a cli-artifacts out of sync with console-operator?  if not, then we *do* know which architectures are available, do we not?

pkg/console/controllers/clidownloads/controller.go
yselkowitz: I changed "Windows" to "Windows 64-bit" in a new commit.  I see no use to adding anything to "Mac", as only 64-bit Intel is currently supported.

benjaminapetersen:
/lgtm

I think we are good to go with this now.

pkg/console/controllers/clidownloads/controller.go
wking: I don't think "64-bit" is sufficiently qualified (seems like it would also include arm64, etc.).  I also think we need to be explicit about Mac, because folks could be running macOS/Darwin on other arches even if we don't support running `oc` on those arches.  I've filed #353 to address both points, PTAL.

manifests/07-downloads-deployment.yaml
wking: > Is it really possible to have a cli-artifacts out of sync with console-operator?

Sure.  For example, we could decide we want to publish a `windows/arm64` `oc`, add it to cli-artifacts, and my standard-path approach (which I just pushed up as #354), would immediately serve it (although you'd still need a console bump to get an entry in `PlatformBasedOCConsoleCLIDownloads`).  And if we wanted to avoid a `PlatformBasedOCConsoleCLIDownloads` bump, we could serve the choices dynamically from the downloads pod.  Hmm... :)

benjaminapetersen: It would definitely be nice to have a single source of truth and not have to sync these two items.  That said, most who use the console will probably want one of the "mainstream" versions of `oc`, so clusttering up the UI with all the things may not be that valuable.

PR #344, opened 2019-11-12
Add e2e test for oc downloads links

benjaminapetersen:
```bash
go test -run TestDownloadsEndpoint -count 1
checking for console-operator availability
found operator deployment
link for Download oc for Linux 200 OK (https://console-openshift-console.apps.bpetersen.devcluster.openshift.com/amd64/linux/oc.tar)
link for Download oc for Mac 200 OK (https://console-openshift-console.apps.bpetersen.devcluster.openshift.com/amd64/mac/oc.zip)
link for Download oc for Windows 200 OK (https://console-openshift-console.apps.bpetersen.devcluster.openshift.com/amd64/windows/oc.zip)
PASS
ok  	github.com/openshift/console-operator/test/e2e	5.274s
```

/assign @spadgett

PR #346, opened 2019-11-13
Bug 1768684: Avoid console_url metric as empty string

benjaminapetersen:
Ensure route.host has a value before tracking console_url metric to avoid the value "".

https://bugzilla.redhat.com/show_bug.cgi?id=1768684


benjaminapetersen:
Hmm, didn't expect it to reference Bugzilla bug 1764313.  This is fine.

pkg/console/operator/sync_v400.go
```diff
@@ -214,6 +205,25 @@ func (co *consoleOperator) SyncConsolePublicConfig(consoleURL string) (*corev1.C
...
+		return
+	}
+	// if status.consoleURL is empty, and the consoleURL has a value, set it
+	if consoleConfig.Status.ConsoleURL == "" && len(consoleURL) != 0 {
+		co.consoleMetrics.ConsoleURL.WithLabelValues(consoleURL).Set(1)
+		return
+	}
+	// if status.consoleURL is different than consoleURL,
+	// clear the old, then set new
+	if consoleConfig.Status.ConsoleURL != consoleURL {
```
jhadvig: will this ever be `true`? since this check is already done in `SyncConsoleConfig()` and is setting the `consoleConfig.Status.ConsoleURL` to `consoleURL` ?

pkg/console/operator/sync_v400.go
```diff
@@ -192,20 +192,11 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console,
...
-		co.consoleMetrics.ConsoleURL.WithLabelValues(consoleConfig.Status.ConsoleURL).Set(0)
-	}
-	if len(consoleURL) != 0 {
-		// only update to new if we have a url
-		co.consoleMetrics.ConsoleURL.WithLabelValues(consoleURL).Set(1)
-	}
-
 	if updated.Status.ConsoleURL != consoleURL {
 		klog.V(4).Infof("updating console.config.openshift.io with url: %v", consoleURL)
 		updated.Status.ConsoleURL = consoleURL
```
jhadvig: lets put content of this check to `handleConsoleURLMetric()` as well?

pkg/console/metrics/metrics.go
```diff
@@ -30,3 +35,48 @@ func Register() *ConsoleMetrics {
 	})
 	return singleton
 }
+
```
benjaminapetersen: This seems a much better abstraction.  All the metrics logic in the metrics package.  I think this may handle all of the cases.

@jhadvig @spadgett for thoughts, anything missing?

benjaminapetersen: I could potentially add a `klog.V(4)` message to each.

benjaminapetersen: Updated with logs. 2 cases (unitialized, and unhandled) are error cases. Rest, logging if bump the `-v` flag to `4+`.

pkg/console/metrics/metrics.go
```diff
@@ -30,3 +35,53 @@ func Register() *ConsoleMetrics {
...
+func HandleConsoleURL(oldURL, newURL string) {
+	// don't call before register
+	if !singleton.initialized {
```
spadgett: Can this ever happen?

benjaminapetersen: Not at present, no.  
Thinking about future additions and establishing a simple pattern.

pkg/console/operator/sync_v400.go
benjaminapetersen: pulling this out into the `metrics` package entirely.  I think its a better pattern/abstraction.

pkg/console/metrics/metrics.go
spadgett: I'd remove this check if it's not possible now and add it when we have a specific need.

pkg/console/operator/sync_v400.go
jhadvig: +1 for moving

spadgett:
Looks like you have govet errors

pkg/console/metrics/console_url.go
```diff
@@ -0,0 +1,53 @@
+package metrics
+
+import "k8s.io/klog"
+
+func HandleConsoleURL(oldURL, newURL string) {
+	// don't call before register
+	if !singleton.initialized {
```
spadgett: I would remove this check if we don't need it. Keep it simple

benjaminapetersen:
Removed the `initialized` bool & double checked `gofmt`.  Should be good.

benjaminapetersen:
```
 --- FAIL: TestRemoved (46.23s) 
```
See #344 for a fix for this test, which adds 3x retries before reporting failure.

benjaminapetersen:
This should now pass with the fix that went in with the downloads e2e.

pkg/console/transport/transport.go
```diff
@@ -0,0 +1,62 @@
...
+	"k8s.io/apimachinery/pkg/util/net"
+	ktransport "k8s.io/client-go/transport"
+)
+
+// TODO: this is borrowed from:
+//   https://github.com/openshift/cluster-authentication-operator/blob/df3160d1ab9f24448c853bfe567c44a2763017cb/pkg/operator2/transport.go
+//   but is intended to be moved to library-go at some point (see comments within the file)
+//   When that happens, remove this code & replace with the library-go version
+
+// transportFor returns an http.Transport for the given ca and client cert data (which may be empty)
```
jhadvig:
```suggestion
// transportFor returns an http.RoundTripper for the given ca and client cert data (which may be empty)
```

pkg/console/transport/transport.go
```diff
@@ -0,0 +1,62 @@
...
+func transportFor(serverName string, caData, certData, keyData []byte) (http.RoundTripper, error) {
```
jhadvig: Is this func actually called somewhere ?

benjaminapetersen: Wup, didn't intend to include this.  I'll remove.  Its potentially for a different PR.

pkg/console/transport/transport.go
benjaminapetersen: Removed

spadgett:
> error: received unexpected HTTP status: 500 Internal Server Error

/retest

benjaminapetersen:
Manual :)

PR #347, opened 2019-11-13
Add liveness and readiness probes to console operator

benjaminapetersen:
We have several endpoints we can hit on the operator:

```
{
  "paths": [
    "/apis",
    "/healthz",
    "/healthz/log",
    "/healthz/ping",
    "/livez",
    "/livez/log",
    "/livez/ping",
    "/metrics",
    "/readyz",
    "/readyz/log",
    "/readyz/ping",
    "/readyz/shutdown"
  ]
}
```
So adding both `/healthz` and `/readyz` probes.  Initial prob config is basically the same as that of the `console` `deployment` otherwise.

/assign @spadgett  @jhadvig 

To manually test these endpoints, do something like `oc create -f` on:

```yaml
apiVersion: route.openshift.io/v1
kind: Route
metadata:
  name: console-operator-route-for-testing
spec:
  port:
    targetPort: https
  tls:
    insecureEdgeTerminationPolicy: Redirect
    termination: passthrough
  to:
    # this is fine, the service name is a misnomer, its generic to the operator but initial intent was only metrics
    kind: Service
    name: metrics
    weight: 100
  wildcardPolicy: None
```

then:

```bash
# health
curl --insecure -H "Authorization: Bearer $(oc whoami --show-token)" $(oc get route console-operator-route-for-testing --template 'https://{{.spec.host}}/healthz')
# ready
curl --insecure -H "Authorization: Bearer $(oc whoami --show-token)" $(oc get route metrics-route-for-testingconsole-operator-route-for-testing --template 'https://{{.spec.host}}/readyz')
```

### DIFF
--- a/manifests/07-operator.yaml
+++ b/manifests/07-operator.yaml
@@ -63,6 +63,16 @@ spec:
           requests:
             memory: "100Mi"
             cpu: "10m"
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 8443
+            scheme: HTTPS
+        readinessProbe:
+          httpGet:
+            path: /readyz
+            port: 8443
+            scheme: HTTPS
       volumes:
       - name: config
         configMap:


manifests/07-operator.yaml
```diff
@@ -63,6 +63,25 @@ spec:
           requests:
             memory: "100Mi"
             cpu: "10m"
+        livenessProbe:
+          failureThreshold: 3
+          httpGet:
+            path: /healthz
+            port: 8443
+            scheme: HTTPS
+          initialDelaySeconds: 150
```
spadgett: Why delay 150 seconds?

I think we can just take the defaults for the props other than `httpGet`

benjaminapetersen: Copy/paste from console.  Not sure why we had such a long initial delay there?

benjaminapetersen:
Agree, defaults seem fine.

```golang
type Probe struct {
	// The action taken to determine the health of a container
	Handler `json:",inline" protobuf:"bytes,1,opt,name=handler"`
	// Number of seconds after the container has started before liveness probes are initiated.
	// More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#container-probes
	// +optional
	InitialDelaySeconds int32 `json:"initialDelaySeconds,omitempty" protobuf:"varint,2,opt,name=initialDelaySeconds"`
	// Number of seconds after which the probe times out.
	// Defaults to 1 second. Minimum value is 1.
	// More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#container-probes
	// +optional
	TimeoutSeconds int32 `json:"timeoutSeconds,omitempty" protobuf:"varint,3,opt,name=timeoutSeconds"`
	// How often (in seconds) to perform the probe.
	// Default to 10 seconds. Minimum value is 1.
	// +optional
	PeriodSeconds int32 `json:"periodSeconds,omitempty" protobuf:"varint,4,opt,name=periodSeconds"`
	// Minimum consecutive successes for the probe to be considered successful after having failed.
	// Defaults to 1. Must be 1 for liveness and startup. Minimum value is 1.
	// +optional
	SuccessThreshold int32 `json:"successThreshold,omitempty" protobuf:"varint,5,opt,name=successThreshold"`
	// Minimum consecutive failures for the probe to be considered failed after having succeeded.
	// Defaults to 3. Minimum value is 1.
	// +optional
	FailureThreshold int32 `json:"failureThreshold,omitempty" protobuf:"varint,6,opt,name=failureThreshold"`
}
```

manifests/07-operator.yaml
benjaminapetersen: updated

benjaminapetersen:
@spadgett good to go?

PR #350, opened 2019-11-13
[WIP] Route Sync Controller

benjaminapetersen:
WIP, to keep us moving along in the effort to break out our mega-controller into several smaller controllers.  The `RouteSyncController` is a bit more involved than the previous in that the main Operator `sync` still cares about the `route`. However, responsiblity for `update` is extracted, the legacy `sync` should only be interested in `READ` in order to stamp out its other resources.

Opening this up to get some e2e feedback.

Note that #334 should merge first.

### DIFF
--- a/pkg/console/controllers/route/controller.go
+++ b/pkg/console/controllers/route/controller.go
@@ -1,0 +1,114 @@
+package route
+
+import (
+	"fmt"
+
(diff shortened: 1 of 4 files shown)


jhadvig:
@benjaminapetersen we need rebase here

PR #353, opened 2019-11-20
pkg/console/controllers/clidownloads/controller: Arch qualifiers for Mac/Windows

wking:
Folks could presumably be running these OSes on other architectures.  For example, [the GOOS/GOARCH docs][1] provide for darwin on 386, amd64, arm, and arm64 and windows on 386 and amd64 (although [Windows runs on ARM64][2] and possibly other architectures as well).

I'm sticking with `x86_64` for consistency with the Linux entry from 865d84d7bb (#343).  Previous discussion starting [here][3].  CC @yselkowitz.

[1]: https://golang.org/doc/install/source#environment
[2]: https://docs.microsoft.com/en-us/windows/arm/
[3]: https://github.com/openshift/console-operator/pull/343#discussion_r347735631

### DIFF
--- a/pkg/console/controllers/clidownloads/controller.go
+++ b/pkg/console/controllers/clidownloads/controller.go
@@ -161,8 +161,8 @@ func PlatformBasedOCConsoleCLIDownloads(host, cliDownloadsName string) *v1.Conso
 		{"Linux for ARM 64", "arm64/linux", "oc.tar"},
 		{"Linux for IBM Power, little endian", "ppc64le/linux", "oc.tar"},
 		{"Linux for IBM Z", "s390x/linux", "oc.tar"},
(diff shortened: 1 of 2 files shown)


benjaminapetersen:
/lgtm

Works for me. I think both @wking / @yselkowitz have stronger opinions about the link names than I do. These are sufficiently clear.

yselkowitz:
These names should be those expected by users of those platforms.  One does not generally say "Windows for x86_64", and Windows for ARM is rare enough that it needs to be called out as such and would not be confused (not to mention that golang has not support for it right now).

The only versions of Mac currently supported are 64-bit Intel.  All versions of Mac for 32-bit Intel are no longer supported by Apple, and this would not be confused with iOS (which is also included in "darwin") which runs on ARM.  Therefore there is no need to qualify Mac.

PR #354, opened 2019-11-20
manifests/07-downloads-deployment: Serve from standardized paths

wking:
Lean on the path consistency from openshift/oc@e949088d (openshift/oc#153) to make it easier to scale as we add/remove arch/OS support.  With this commit, the downloads deployment is less opinionated about what is in the cli-artifacts bundle.  It looks for all possible arch/OS combinations, serves what it finds, and 404s requests for binaries it could not find.

I've also renamed mac -> darwin in the canonical URIs, so there's no mental overhead for translating from [the conventional GOOS/GOARCH][1].  I've added symlinks in the pod to support external folks who had hard-coded the old URIs, but we should be able to drop those symlinks after a reasonable deprecation period.  I have not overhauled the server to 301 requests to the deprecated URI, and we probably want that before dropping the symlinks.

CC @yselkowitz, previous discussion [here][2] and in later comments in that thread.

[1]: https://golang.org/doc/install/source#environment
[2]: https://github.com/openshift/console-operator/pull/343#discussion_r347734803

### DIFF
--- a/manifests/07-downloads-deployment.yaml
+++ b/manifests/07-downloads-deployment.yaml
@@ -59,7 +59,7 @@ spec:
         - '-c'
         - |
           cat <<EOF >>/tmp/serve.py
-          import BaseHTTPServer, os, re, signal, SimpleHTTPServer, socket, sys, tarfile, tempfile, threading, time, zipfile
+          import BaseHTTPServer, errno, os, re, signal, SimpleHTTPServer, socket, sys, tarfile, tempfile, threading, time, zipfile
 
           signal.signal(signal.SIGTERM, lambda signum, frame: sys.exit(0))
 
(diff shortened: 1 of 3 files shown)


manifests/07-downloads-deployment.yaml
```diff
@@ -85,32 +85,36 @@ spec:
           temp_dir = tempfile.mkdtemp()
           print('serving from {}'.format(temp_dir))
           os.chdir(temp_dir)
-          for arch in ['amd64']:
+          for arch in ['amd64', 'arm64', 'ppc64le', 's390x']:
```
benjaminapetersen: Have reference for where these come from handy?

manifests/07-downloads-deployment.yaml
```diff
@@ -85,32 +85,36 @@ spec:
...
-              for operating_system in ['linux']:
-                  os.mkdir(os.path.join(arch, operating_system))
+              for operating_system in ['linux', 'darwin', 'windows']:
+                  os_path = os.path.join(arch, operating_system)
+                  try:
+                      os.mkdir(os_path)
+                  except OSError as error:
+                      if error.errno != errno.EEXIST:
+                          raise
+                  basename = 'oc'
```
benjaminapetersen: Not gonna lie, this is a bit uglier than the prev version. 😃

manifests/07-downloads-deployment.yaml
benjaminapetersen: Perhaps I just need a referesher about the process that spits out the `cli-artifacts`.

wking: > Perhaps I just need a referesher about the process that spits out the `cli-artifacts`.

[Dockerfile][1] hits [this Makefile line][2].  But with this approach, there's no reason we couldn't fill in [all the GOOS and GOARCH known by Go][3].

[1]: https://github.com/openshift/oc/blob/386b42b6a3ce0c3060f5c95b9fea5f464745f90a/images/cli-artifacts/Dockerfile.rhel
[2]: https://github.com/openshift/oc/blob/386b42b6a3ce0c3060f5c95b9fea5f464745f90a/Makefile#L99
[3]: https://golang.org/doc/install/source#environment

manifests/07-downloads-deployment.yaml
wking: > Not gonna lie, this is a bit uglier than the prev version.

Anything in particular that I can try to clean up?

benjaminapetersen: No, I think its fine.  Just had to refresh myself on some python-isms.

For fun, its not immediately obvious why we end with a list comprehension:
```
[Thread(i, socket=sock) for i in range(100)]
```
but not storing or using the list. 
Also followed by `time.sleep(9e9)`.  
`9e9` > `2**32` == max array length? 
But both of those have been part of this file since the beginning.  
Fun trivia 😉

benjaminapetersen:
Needs a rebase.

jhadvig:
@wking could you please rebase this PR

wking:
> could you please rebase this PR

Sorry, missed these.  And since openshift/oc#171 partially walked back the standardized paths, this approach will no longer work without some touchups.  I'll close until we get openshift/oc#172 or similar to restore the standardized paths which would let us revive this approach.

PR #355, opened 2019-11-26
bug 1772759: fix console operator panic when RouterCAValidationDegraded 

benjaminapetersen:
Reverts router-ca changes in #328.
Not sure yet if this is the direction we want to go, considering https://github.com/openshift/enhancements/pull/126.

I suspect that some of this code can return in a future PR.  We will likely not need the `resource-syncer`, and instead will be able to annotate a new `configmap` with the appropriate `router-ca` annotation to get it injected like the `proxy`. Then we can assess if it exists & mount in the deployment, or not.

/assign @spadgett @jhadvig


pkg/console/starter/starter.go
```diff
@@ -259,6 +242,9 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 			"Progressing",
 			"Available",
 			// "Faililng"
+			// added in 4.3, can remove in 4.4, to ensure operators running 4.3.x are not wedged.
```
benjaminapetersen: I'm not sure if this change is strictly necessary.  These were new conditions with the router-ca sync commit within the 4.3 time frame, so if the code is not running in production we can ignore.

jhadvig: yeah not sure we need this change, if not would rather not add it...

spadgett: I would take this out, and we can backport the revert to 4.3.

spadgett:
/hold

I'd remove the second commit

pkg/console/starter/starter.go
```diff
@@ -242,6 +242,9 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 			"Progressing",
 			"Available",
 			// "Faililng"
```
spadgett: Not for this PR, but it looks like we can remove these now?

benjaminapetersen: Yeah we should eliminate.

benjaminapetersen:
Removed the second commit.

PR #356, opened 2019-12-02
Remove stale conditions from RemoveStaleConditionsController()

benjaminapetersen:
Per this comment: https://github.com/openshift/console-operator/pull/355#discussion_r352743247

I'm removing the stale conditions, but leaving the controller with a comment for future use.  The potential to wedge the operator permanently inclines me strongly to keep the controller present, even if inactive for now.

/assign @spadgett @jhadvig

### DIFF
--- a/pkg/console/starter/starter.go
+++ b/pkg/console/starter/starter.go
@@ -23,7 +23,6 @@ import (
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
 	"github.com/openshift/library-go/pkg/operator/management"
 	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
-	"github.com/openshift/library-go/pkg/operator/staleconditions"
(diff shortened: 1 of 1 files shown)


pkg/console/starter/starter.go
```diff
@@ -252,13 +252,13 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 
 	staleConditionsController := staleconditions.NewRemoveStaleConditions(
```
spadgett: I'd delete or maybe comment out this whole block and `staleConditionsController.Run` if we're not using it.

jhadvig: agree, dont think this is necessary anymore :)

benjaminapetersen: I can comment it out, good with that.
With an explicit reminder comment about re-enabling the `Run()`, that tends to be a gotcha.

PR #357, opened 2019-12-02
[WIP] Add ConsolePodsMustSettle() helper to test/e2e

benjaminapetersen:
We should be testing that the console pods also settle down into a happy state in our e2e tests.

### DIFF
--- a/test/e2e/framework/console.go
+++ b/test/e2e/framework/console.go
@@ -6,7 +6,8 @@ import (
 	"testing"
 	"time"
 
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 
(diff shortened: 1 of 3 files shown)


benjaminapetersen:
The possibility that sibling PRs such as: 
- #250 and [console klog update](https://github.com/openshift/console/pull/1862) 

could break our merge queue or make it to production without an automated verification that our pods will continue to function concerns me a bit.

test/e2e/framework/console.go
```diff
@@ -0,0 +1,87 @@
...
+	if deploymentsub.AllPodsHealthy(deployment) {
+		return nil, true
+	}
+
+	fmt.Printf("something wrong, %d/%d replicas ready of expected %d\n", deployment.Status.ReadyReplicas, deployment.Status.Replicas, deployment.Status.Replicas)
+
+	if deployment.Status.Replicas > 2 {
+		// todo: may or may not be a bad thing, we need to investigate further
+		//    this may simply indicate churn, new pods are rolling out, old pods being terminated
+		//    do we need to consider this?
```
jhadvig: since `deployment.status.replicas` is targeting "non-terminated pods" we should be ok

PR #366, opened 2020-01-07
RBAC for console backend to request openshift-config-managed configmaps

jhadvig:
Adding:
 - `Role` for GETting ConfigMaps from the `openshift-config-managed` namespace. Haven't found any existing Role on the cluster that would be suitable.
 - `ServiceAccount` for the console, since there are only default ServiceAccounts in the `openshift-console` namespace
 - `RoleBinding` - for binding the added `Role` and `ServiceAccount`

Story: https://issues.redhat.com/browse/CONSOLE-1993

/assign @benjaminapetersen

### DIFF
--- a/manifests/03-rbac-role-ns-openshift-config-managed.yaml
+++ b/manifests/03-rbac-role-ns-openshift-config-managed.yaml
@@ -35,3 +35,16 @@ rules:
   - console-public
   verbs:
   - get
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: console-configmap-reader
+  namespace: openshift-config-managed
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
--- a/manifests/06-sa.yaml
+++ b/manifests/06-sa.yaml
@@ -3,3 +3,9 @@ kind: ServiceAccount
 metadata:
   name: console-operator
   namespace: openshift-console-operator
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: console
+  namespace: openshift-console
--- a/manifests/04-rbac-rolebinding.yaml
+++ b/manifests/04-rbac-rolebinding.yaml
@@ -80,4 +80,17 @@ subjects:
   - kind: ServiceAccount
     name: console-operator
     namespace: openshift-console-operator
-
+---
+kind: RoleBinding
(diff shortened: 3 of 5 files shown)


manifests/03-rbac-role-ns-openshift-config-managed.yaml
```diff
@@ -35,3 +35,16 @@ rules:
   - console-public
   verbs:
   - get
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: managed-configmap-reader
```
benjaminapetersen: So currently we have:

```bash
oc get role -n openshift-config-managed
NAME               AGE
console-operator   28h
console-public     28h
machine-approver   28h
```
For this reason I'm inclined to name it `console-configmap-reader`.  Since its in `openshift-config-managed` namespace.  That would give us:

```bash
oc get role -n openshift-config-managed
NAME                              AGE
console-configmap-reader          28h
console-operator                  28h
console-public                    28h
machine-approver                  28h
```

benjaminapetersen: We did speak about `console-public` being somewhat "different" as its more for other components being able to get the `console-public` configmap so those components can `oc get configmap console-public -n openshift-config-managed --template '{{.data.consoleURL}}'` and use the console url value.

manifests/06-sa.yaml
```diff
@@ -3,3 +3,9 @@ kind: ServiceAccount
 metadata:
   name: console-operator
   namespace: openshift-console-operator
+---
+apiVersion: v1
```
benjaminapetersen: Oh this is interesting.  Does the console not yet have a service account?

benjaminapetersen: Guess so:

```bash
oc get serviceaccount -n openshift-console
NAME       SECRETS   AGE
builder    2         28h
default    2         28h
deployer   2         28h

oc get serviceaccount -n openshift-console-operator
NAME               SECRETS   AGE
builder            2         28h
console-operator   2         28h
default            2         28h
deployer           2         28h
```

jhadvig: yeah, only the default ones. For that reason I've created a new one

jhadvig:
@benjaminapetersen PR updated, set the console podSpec to use the `console` ServiceAccount. Tokens are mounted automatically.

PR #370, opened 2020-01-09
Add example for console extensions

benjaminapetersen:
I've been asked several times to supply examples for console extensions.  It seems appropriate to provide a few in our `/examples` directory.

API: https://github.com/openshift/api/tree/master/console/v1
(Documentation on API objects is quite good, so it is not terribly difficult to discern how to use these.  The most difficult parts are the embedded yaml in `ConsoleYAMLSample` and the `matchExpressions` for `ConsoleLink`)

TODO:
- clean up example text, some of it may be a bit silly
- remove those titled `badux` as these point out gaps in validation. examples should lead only to a good ux.

/assign @spadgett @jhadvig @rhamilto 
For feedback, in case there is anything else we want to include or exclude.

### DIFF
--- a/examples/cr-console-extensions.yaml
+++ b/examples/cr-console-extensions.yaml
@@ -1,0 +1,352 @@
+# a few examples of each of the supported extensions for console
+apiVersion: console.openshift.io/v1
+kind: ConsoleLink
+metadata:
+  name: example
+spec:
+  href: 'https://www.example.com'
+  location: HelpMenu
+  text: Link 1
+---
+apiVersion: console.openshift.io/v1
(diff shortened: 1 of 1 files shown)


rhamilto:
This is great.  Thinking out loud, any reason we don't point to the examples in https://github.com/openshift/console/blob/master/frontend/public/models/yaml-templates.ts so we don't maintain them two places?

benjaminapetersen:
No good reason.  Other than this is my home-grown list during dev :)

jhadvig:
Closing since this is addressed in https://github.com/openshift/console-operator/pull/436

PR #374, opened 2020-01-21
[WIP] Console-1933 IPv6 support

benjaminapetersen:
Ensures that the console can support either IPv4 or IPv6 (but not yet both).

RE [doc changes needed](https://docs.google.com/document/d/1zznvkK87y7UvOJZqs2DM4oWzx94P8Lg6YUb43whil3g/edit#heading=h.wp47a79o7bw0)

/assign @russellb 
/cc @spadgett @jhadvig 

- [x] Operator config `bindNetwork: "tcp"`
   - Per https://github.com/openshift-kni/console-operator/commit/31b4d3910f2d476d3351d14f0999a75c46195914
- [x] Console config handles IPv4 or IPv6 based on `network.config.openshift.io`
   - Per https://github.com/openshift-kni/console-operator/commit/01865a640a24ee0bd719306c7a5b08802667e84d
- [ ] Downloads Deployment handles IPv4 or IPv6
   - Per https://github.com/openshift-kni/console-operator/commit/f395cb1a571fc867046750bc57a490e1d5264f24
- [ ] tests

### DIFF
--- a/pkg/console/subresource/deployment/downloads.go
+++ b/pkg/console/subresource/deployment/downloads.go
@@ -27,7 +27,7 @@ func DownloadsDeployment(ipVMode config.IPMode) *appsv1.Deployment {
 	deployment := &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      api.OpenShiftDownloadsName,
-			Namespace: api.OpenShiftConsoleName,
+			Namespace: api.OpenShiftConsoleNamespace,
 			Labels:    labels,
 		},
 		Spec: appsv1.DeploymentSpec{
--- a/pkg/console/subresource/deployment/downloads.go
+++ b/pkg/console/subresource/deployment/downloads.go
@@ -78,7 +78,7 @@ func downloadsContainer(ipVMode config.IPMode) corev1.Container {
 				corev1.ResourceMemory: resource.MustParse("50Mi"),
 			},
 		},
-		Args: downloadsArgs(ipVMode config.IPMode),
+		Args: downloadsArgs(ipVMode),
 	}
 }
 
--- a/pkg/console/subresource/deployment/downloads.go
+++ b/pkg/console/subresource/deployment/downloads.go
@@ -1,0 +1,204 @@
+package deployment
+
+import (
+	"fmt"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+
+	"github.com/openshift/console-operator/pkg/api"
+	"github.com/openshift/console-operator/pkg/console/config"
+)
+
+func DownloadsDeployment(ipVMode config.IPMode) *appsv1.Deployment {
+
+	replicas := int32(ConsoleReplicas)
+
+	labels := map[string]string{
+		"app":       "console",
+		"component": "downloads",
+	}
+
+	gracePeriod := int64(1)
+
+	deployment := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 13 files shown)


russellb:
For reference, ere is the diff with the IPv6 fork where we hard coded Ipv6 supoprt to get things working: https://github.com/openshift/console-operator/compare/release-4.3...openshift-kni:4.3-ipv6

spadgett:
@russellb @markmc Is it possible for console to always listen on `https://[::]:8443` and skip reading the network config?

russellb:
> @russellb @markmc Is it possible for console to always listen on `https://[::]:8443` and skip reading the network config?

`::` means all IPv6 addresses and `0.0.0.0` means all IPv4 addresses.

golang is smart about this, though.  When you open a `tcp` listening socket, it will do the right thing and use both Ipv4 and Ipv6 if both are available.  That's why the change to the manifest to add `bindNetwork: tcp` works.

When I made the change to the https URL before, `bindNetwork` wasn't a valid option there.  I didn't look very far though, since I just wanted to make it work quickly.  It's something you could dig into though.

cc @danwinship

spadgett:
@russellb Console currently ignores `bindNetwork`. It was there for API compatibility reasons with the 3.x console config. It doesn't seem like it would be a difficult change to add, though.

Here's what console does today:

https://github.com/openshift/console/blob/master/cmd/bridge/main.go#L529-L543

If there's a way we can make this work for both cases without needing to read the network config, that would simplify things a lot.

russellb:
> @russellb Console currently ignores `bindNetwork`. It was there for API compatibility reasons with the 3.x console config. It doesn't seem like it would be a difficult change to add, though.
> 
> Here's what console does today:
> 
> https://github.com/openshift/console/blob/master/cmd/bridge/main.go#L529-L543
> 
> If there's a way we can make this work for both cases without needing to read the network config, that would simplify things a lot.

I wonder if this change wasn't needed, and I thought it was when we did it originally, trying to get the console to work.

Looking at the code for http.Server() in https://golang.org/src/net/http/server.go, it looks like it should always be using a bindNetwork of "tcp", which should mean that "0.0.0.0:6443" should work for both IPv4 and IPv6.

Why don't we try without that change first.

@danwinship sound right to you?

```
func (srv *Server) ListenAndServe() error {
	if srv.shuttingDown() {
		return ErrServerClosed
	}
	addr := srv.Addr
	if addr == "" {
		addr = ":http"
	}
	ln, err := net.Listen("tcp", addr)
	if err != nil {
		return err
	}
	return srv.Serve(ln)
}
```

russellb:
Note that you’ll still have to check the networking config to fix the download code, I presume.

pkg/console/subresource/deployment/downloads.go
```diff
@@ -0,0 +1,204 @@
...
+
+	// if we get a IPv4v6 value, we will have to panic as we don't know how to support both yet
+	if ipVMode == config.IPv4v6Mode {
+		panic(fmt.Sprintf("mode not supported: %v", ipVMode))
+	}
+
+	if ipVMode == config.IPv6Mode {
+		addressFamily = "AF_INET6"
+	}
+
```
benjaminapetersen: This isn't working correctly, the `|` is dropped so the pod doesn't start.  Will fix.

danwinship:
Is this / are these on the pod network or the host network?

For pod-network services, you can just bind dual-stack (`0.0.0.0` and `bindNetwork: tcp`) and you're done. Likewise for hostNetwork services in 4.4.

But hostNetwork services in 4.3 that are single-stack IPv4 in 4.3.0 need to stay single-stack IPv4 in 4.3.z, to avoid surprising behavior changes (aka security problems) in IPv4 clusters on nodes that happen to also have IPv6 addresses.

> If there's a way we can make this work for both cases without needing to read the cluster network config, that would simplify things a lot.

Since we don't support dual-stack clusters in 4.3, a pod can figure out if it's in a single-stack IPv4 or single-stack IPv6 cluster based on what address family its PodIP (or HostIP) is.

benjaminapetersen:
/close

In favor of #375 if it meets our needs.

PR #375, opened 2020-01-23
Bug 1794546: Console-1933: IPv6 support (simpler alternative)

benjaminapetersen:
Simpler/preferred change over #374 

RE [doc changes needed](https://docs.google.com/document/d/1zznvkK87y7UvOJZqs2DM4oWzx94P8Lg6YUb43whil3g/edit#heading=h.wp47a79o7bw0)

/assign @russellb 
/cc @spadgett @jhadvig 

- [x] Operator config `bindNetwork: "tcp"`
   - Per https://github.com/openshift-kni/console-operator/commit/31b4d3910f2d476d3351d14f0999a75c46195914
- [x] Console config handles IPv4 or IPv6 based on `network.config.openshift.io`
   - Per https://github.com/openshift-kni/console-operator/commit/01865a640a24ee0bd719306c7a5b08802667e84d
- [x] Downloads Deployment handles IPv4 or IPv6
   - Per https://github.com/openshift-kni/console-operator/commit/f395cb1a571fc867046750bc57a490e1d5264f24
- [x] tests updated
   - all unit tests updated to verify the expected output
- [ ] IPv6 test?
   - All current tests should continue to pass on IPv4
   - Unclear how to test an IPv6 cluster at this time

### DIFF
--- a/manifests/07-downloads-deployment.yaml
+++ b/manifests/07-downloads-deployment.yaml
@@ -113,8 +113,11 @@ spec:
                   zip.write(path, basename)
 
           # Create socket
-          addr = ('', 8080)
-          sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+          # IPv6 should handle IPv4 passively so long as it is not bound to a
+          # specific address or set to IPv6_ONLY
+          # https://stackoverflow.com/questions/25817848/python-3-does-http-server-support-ipv6
+          addr = ('::', 8080)
+          sock = socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
           sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
           sock.bind(addr)
           sock.listen(5)
--- a/manifests/05-config.yaml
+++ b/manifests/05-config.yaml
@@ -9,3 +9,5 @@ data:
     kind: GenericOperatorConfig
     leaderElection:
       namespace: openshift-console-operator
+    servingInfo:
+      bindNetwork: "tcp"
--- a/pkg/console/subresource/configmap/configmap_test.go
+++ b/pkg/console/subresource/configmap/configmap_test.go
@@ -78,7 +78,7 @@ customization:
   branding: ` + DEFAULT_BRAND + `
   documentationBaseURL: ` + DEFAULT_DOC_URL + `
 servingInfo:
-  bindAddress: https://0.0.0.0:8443
+  bindAddress: https://[::]:8443
   certFile: /var/serving-cert/tls.crt
   keyFile: /var/serving-cert/tls.key
 providers: {}
@@ -124,7 +124,7 @@ customization:
   branding: ` + DEFAULT_BRAND + `
   documentationBaseURL: ` + DEFAULT_DOC_URL + `
 servingInfo:
-  bindAddress: https://0.0.0.0:8443
+  bindAddress: https://[::]:8443
   certFile: /var/serving-cert/tls.crt
   keyFile: /var/serving-cert/tls.key
 providers: {}
@@ -178,7 +178,7 @@ customization:
   branding: online 
   documentationBaseURL: https://docs.okd.io/4.4/
 servingInfo:
-  bindAddress: https://0.0.0.0:8443
+  bindAddress: https://[::]:8443
   certFile: /var/serving-cert/tls.crt
   keyFile: /var/serving-cert/tls.key
 providers: {}
@@ -241,7 +241,7 @@ customization:
   branding: ` + string(operatorv1.BrandDedicated) + `
   documentationBaseURL: ` + mockOperatorDocURL + `
 servingInfo:
-  bindAddress: https://0.0.0.0:8443
+  bindAddress: https://[::]:8443
   certFile: /var/serving-cert/tls.crt
   keyFile: /var/serving-cert/tls.key
 providers: {}
@@ -311,7 +311,7 @@ customization:
   customLogoFile: /var/logo/logo.svg
   customProductName: custom-product-name
 servingInfo:
-  bindAddress: https://0.0.0.0:8443
+  bindAddress: https://[::]:8443
   certFile: /var/serving-cert/tls.crt
   keyFile: /var/serving-cert/tls.key
 providers: {}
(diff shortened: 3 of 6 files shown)


manifests/07-downloads-deployment.yaml
```diff
@@ -113,8 +113,11 @@ spec:
                   zip.write(path, basename)
 
           # Create socket
-          addr = ('', 8080)
-          sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+          # IPv6 should handle IPv4 passively so long as it is not bound to a
```
benjaminapetersen: @wking for a glance. This is a best guess at this one, but I don't have a clear test path (and have never written python 😄 😄).

Anything beyond this and I think I'd rather see the script built into the `cli-artifacts` image (or an image based on it) so we can work with the code more easily.

manifests/07-downloads-deployment.yaml
```diff
@@ -113,8 +113,11 @@ spec:
...
+          # specific address or set to IPv6_ONLY
+          # https://stackoverflow.com/questions/25817848/python-3-does-http-server-support-ipv6
+          addr = ('::', 8080)
```
spadgett: We should confirm that `::` doesn't break ipv4. It's not clear to me from the stack overflow thread

manifests/07-downloads-deployment.yaml
benjaminapetersen: Verified in my own non-IPv6 cluster that this still runs.

spadgett:
@benjaminapetersen Do we have any CLI download test coverage in CI?

manifests/07-downloads-deployment.yaml
bparees: if it breaks ipv4 will it not fail our CI jobs?

benjaminapetersen: Yup, we have e2e tests for this so it would fail CI if this is a problem.

I ran it in my cluster, was able to verify downloads still download via the UI.  pod is running, etc.  I think we are good.

benjaminapetersen:
@spadgett yes, we have:
https://github.com/openshift/console-operator/blob/master/test/e2e/downloads_test.go#L24
which will loop our CLIDownloads CRD & verify each link responds with a payload.

We could potentially add one in console to also verify the downloads links do respond with files:
- https://github.com/openshift/console/blob/master/frontend/integration-tests/tests/crd-extensions.scenario.ts#L15
But thats prob not super necessary.

benjaminapetersen:
/retest 

`could not wait for build: the build console-operator failed after 2m8s with reason DockerBuildFailed: Docker build strategy has failed.`

benjaminapetersen:
/cherry-pick release-4.3

This should backport cleanly.

manifests/07-downloads-deployment.yaml
eparis: Can we configure nodes with the `ipv6.disable=1` kernel command line and see what happens if we blacklist the ipv6 kernel module (via an MC?)

benjaminapetersen:
/retest

```
level=fatal msg="failed to fetch Cluster: failed to generate asset \"Cluster\": failed to create cluster: failed to apply using Terraform"
```

manifests/07-downloads-deployment.yaml
danwinship: Yeah, this will most likely fail if `ipv6_disable`. Go-based code doesn't have to worry about that because go has code to deal with that and fall back to `0.0.0.0` if you try to bind to `::` but IPv6 is unavailable.

benjaminapetersen: So do we need to hold on this, or make a bug for follow?

benjaminapetersen:
/retest

unexpected flake:

```
     util.go:26: patching Data on the console ConfigMap
    util.go:33: polling for patched Data on the console ConfigMap
    unmanaged_test.go:18: waiting for cleanup to reach settled state...
    console-operator.go:369: waited 10 seconds to reach settled state...
=== RUN   TestEditUnmanagedService
--- FAIL: TestEditUnmanagedService (69.10s)
    unmanaged_test.go:12: waiting for setup to reach settled state...
    unmanaged_test.go:13: changing console operator state to 'Unmanaged'...
    util.go:49: patching Annotation on the console Service
    util.go:56: polling for patched Annotation on the console Service
    unmanaged_test.go:18: waiting for cleanup to reach settled state...
    util.go:95: patching DisplayName on the odo-cli-downloads ConsoleCLIDownloads custom resource
    util.go:102: polling for patched DisplayName on the odo-cli-downloads ConsoleCLIDownloads custom resource
    unmanaged_test.go:18: waiting for cleanup to reach settled state...
    console-operator.go:369: waited 10 seconds to reach settled state...
    console-operator.go:369: waited 30 seconds to reach settled state...
FAIL
FAIL	github.com/openshift/console-operator/test/e2e	918.176s
[ERROR] PID 231: hack/test-e2e.sh:5: `KUBERNETES_CONFIG=${KUBECONFIG} go test -timeout 30m -v ./test/e2e/` exited with status 1.
[INFO] 		Stack Trace: 
[INFO] 		  1: hack/test-e2e.sh:5: `KUBERNETES_CONFIG=${KUBECONFIG} go test -timeout 30m -v ./test/e2e/`
[INFO]   Exiting with code 1. 
```
will keep an eye on that.

manifests/07-downloads-deployment.yaml
russellb: You can resolve this by going back to dynamically changing the downloader based on whether the cluster was configured for IPv6 or not.

benjaminapetersen:
/retest

```
level=fatal msg="failed to fetch Cluster: failed to generate asset \"Cluster\": failed to create cluster: failed to apply using Terraform"
```

russellb:
Have you filed a bug somewhere to follow-up on the issue raised by @eparis, where this would break on a host with IPv6 disabled?

manifests/07-downloads-deployment.yaml
benjaminapetersen: Ah, this  inline python script is not great for any kind of additional complexity.  Sounds like an overhaul of the downloads bit.

benjaminapetersen:
Created this bug:
- https://bugzilla.redhat.com/show_bug.cgi?id=1795325

I am presently uncertain about the severity & will be fishing around for further details.

PR #379, opened 2020-02-12
Bug 1777486: Add sharing-config configmap from openshift-monitoring to the console-config

jhadvig:
Monitoring and logging URLs will be available at `console-config` in `openshift-console` namespace so that they available.

Screenshot 2020-02-12 at 17 39 40

/assign @benjaminapetersen

### DIFF
--- a/alt-07-operator.yaml
+++ b/alt-07-operator.yaml
@@ -1,0 +1,83 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: console-operator
+  namespace: openshift-console-operator
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      name: console-operator
+  template:
+    metadata:
+      labels:
+        name: console-operator
+    spec:
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
+      tolerations:
+      - key: node-role.kubernetes.io/master
+        operator: Exists
+        effect: "NoSchedule"
+      - key: "node.kubernetes.io/unreachable"
+        operator: "Exists"
+        effect: "NoExecute"
+        tolerationSeconds: 120
+      - key: "node.kubernetes.io/not-ready"
+        operator: "Exists"
+        effect: "NoExecute"
+        tolerationSeconds: 120
+      priorityClassName: system-cluster-critical
+      serviceAccountName: console-operator
+      containers:
+      - name: console-operator
+        terminationMessagePolicy: FallbackToLogsOnError
+        image: docker.io/jhadvig/console-operator:latest
+        ports:
+        - containerPort: 60000
+          name: metrics
+        command:
+        - console
+        - operator
+        args:
+        - "-v=2"
+        - "--config=/var/run/configmaps/config/controller-config.yaml"
+        imagePullPolicy: Always
+        volumeMounts:
+        - mountPath: /var/run/configmaps/config
+          name: config
+        - mountPath: /var/run/secrets/serving-cert
+          name: serving-cert
+        env:
+        - name: IMAGE
+          value: registry.svc.ci.openshift.org/ocp/4.4-2020-02-11-035631@sha256:668cd22560b41876d5e633418f3b378e5c14eb26837f6aa62d4edf17a1db22e5
+        - name: RELEASE_VERSION
+          value: "0.0.1-snapshot"
+        - name: OPERATOR_NAME
+          value: "console-operator"
+        - name: POD_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
+        resources:
+          requests:
+            memory: "100Mi"
+            cpu: "10m"
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 8443
+            scheme: HTTPS
+        readinessProbe:
+          httpGet:
+            path: /readyz
+            port: 8443
+            scheme: HTTPS
+      volumes:
+      - name: config
+        configMap:
+          name: console-operator-config
+      - name: serving-cert
+        secret:
+          secretName: serving-cert
+          optional: true
--- a/manifests/04-rbac-rolebinding.yaml
+++ b/manifests/04-rbac-rolebinding.yaml
@@ -94,3 +94,17 @@ subjects:
   - kind: ServiceAccount
     name: console
     namespace: openshift-console
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: console-operator
+  namespace: openshift-monitoring
+roleRef:
+  kind: Role
+  name: monitoring-configmap-reader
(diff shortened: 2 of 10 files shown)


manifests/03-rbac-monitoring.yaml
```diff
@@ -0,0 +1,14 @@
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: monitoring-configmap-reader
```
benjaminapetersen: @s-urbaniak just to verify, this role doesn't already exist in some format that we should bind to instead.

manifests/03-rbac-logging.yaml
```diff
@@ -0,0 +1,14 @@
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: logging-configmap-reader
```
benjaminapetersen: @jcantrill just to verify, this role doesn't already exist in some format that we should bind to instead.

manifests/04-rbac-rolebinding.yaml
```diff
@@ -94,3 +94,31 @@ subjects:
   - kind: ServiceAccount
     name: console
     namespace: openshift-console
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: console-operator-monitoring
```
benjaminapetersen: I think we can just name it `console-operator`. its the binding for the `console-operator` and lists the roles it binds (which should all be named something specific, I believe).  The one we have in `openshift-config` is just `console-operator`.

manifests/04-rbac-rolebinding.yaml
```diff
@@ -94,3 +94,31 @@ subjects:
...
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+  - kind: ServiceAccount
+    name: console-operator
+    namespace: openshift-console-operator
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: console-operator-logging
```
benjaminapetersen: Same, can prob just name `console-operator`?

pkg/console/operator/sync_v400.go
```diff
@@ -303,7 +303,17 @@ func (co *consoleOperator) SyncConfigMap(
 		useDefaultCAFile = false
 	}
 
-	defaultConfigmap, _, err := configmapsub.DefaultConfigMap(operatorConfig, consoleConfig, managedConfig, infrastructureConfig, consoleRoute, useDefaultCAFile)
+	monitoringSharingConfig, mscErr := co.configMapClient.ConfigMaps(api.OpenShiftMonitoringNamespace).Get(api.SharingConfigMapName, metav1.GetOptions{})
+	if mscErr != nil && !apierrors.IsNotFound(mscErr) {
+		return nil, false, "FailedMonitoringSharingConfig", mscErr
```
benjaminapetersen: Shall we do `FailedGeMonitoringSharingConfig`? Its long, but the `GET` matches with other `Get` and `Apply` we have.  Though I see above we made an oops and just did `FailedManagedConfig` instead of `FailedGetManagedConfig`.

benjaminapetersen:
/approve
and generally `/lgtm`, just a couple quick Qs to verify and we can move this ahead!

manifests/03-rbac-logging.yaml
jcantrill: @benjaminapetersen This is so the console can read our configmap that has route info in it?  Maybe this? https://github.com/openshift/cluster-logging-operator/blob/master/pkg/k8shandler/visualization.go#L405-L443 Something  we should consider removing?

manifests/04-rbac-rolebinding.yaml
```diff
@@ -94,3 +94,31 @@ subjects:
   - kind: ServiceAccount
     name: console
     namespace: openshift-console
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: console-operator
```
jhadvig: @benjaminapetersen updated both names as suggested 👍

manifests/03-rbac-logging.yaml
jhadvig: We will need to add the Role and RoleBinding to the cluster-logging-operator manifest, due to CI error that occurs when creating the `logging-configmap-reader` Role, since the `openshift-logging` namespace is not present on the cluster by default

alt-07-operator.yaml
```diff
@@ -0,0 +1,83 @@
+apiVersion: apps/v1
```
benjaminapetersen: Oops include?

pkg/console/subresource/consoleserver/types.go
```diff
@@ -39,9 +39,11 @@ type ServingInfo struct {
 
 // ClusterInfo holds information the about the cluster such as master public URL and console public URL.
 type ClusterInfo struct {
-	ConsoleBaseAddress string `yaml:"consoleBaseAddress,omitempty"`
-	ConsoleBasePath    string `yaml:"consoleBasePath,omitempty"`
-	MasterPublicURL    string `yaml:"masterPublicURL,omitempty"`
+	ConsoleBaseAddress string            `yaml:"consoleBaseAddress,omitempty"`
```
benjaminapetersen: Do we need the console PR first to accept these?

pkg/console/subresource/consoleserver/config_builder_test.go
```diff
@@ -257,6 +308,12 @@ servingInfo:
   keyFile: /var/serving-cert/tls.key
 clusterInfo:
   masterPublicURL: https://foobar.com/api
+  monitoringURLs:
```
benjaminapetersen: I may be inclined to do:

```yaml
  monitoring:
    alertmanagerURL: https://alertmanager.url.com
    grafanaURL: https://grafana.url.com
    prometheusURL: https://prometheus.url.com
  logging:
    kibanaAppURL: https://kibanaApp.url.com
```
Just eliminates `URLs` from the wrapper object, since everything underneath has `URL` as a suffix, anyway.  I don't expect we will ever have other config, but if we had `non-url` config, we can still stick it in the same spot then.

manifests/03-rbac-logging.yaml
benjaminapetersen: Ah, yes so:

```yaml
Role: sharing-config-reader
RoleBinding: openshift-logging-sharing-config-reader-binding (system:authenticated)
```
That was basically my question, should we use this for `console-operator` instead of creating new Role & RoleBinding for ourselves.

jhadvig:
Closing this one since we decided to move the `sharing-config`s from both `openshift-logging` and `openshift-monitoring` namespaces into `openshift-config-managed` and properly name them.
This shall be handled by the owned operators

PR #380, opened 2020-02-18
Bug 1798315: Bump deps

jhadvig:
Bumping the deps together with `structured-merge-diff`from which the `version` package is coming from is fixing the issue of invalid gitCommit and gitVersion values.

/assign @benjaminapetersen


benjaminapetersen:
/lgtm 

so long as the tests pass.  +1 to pin, thx!

PR #382, opened 2020-02-25
Add Helm and HelmChartRepo types

jhadvig:
Adding missing [types](https://github.com/openshift/console/blob/master/pkg/serverconfig/types.go#L63-L70) to have type parity with the console repo

/assign @benjaminapetersen

### DIFF
--- a/pkg/console/subresource/consoleserver/types.go
+++ b/pkg/console/subresource/consoleserver/types.go
@@ -79,3 +79,12 @@ type Customization struct {
 type Providers struct {
 	StatuspageID string `yaml:"statuspageID,omitempty"`
 }
+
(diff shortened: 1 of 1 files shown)


pkg/console/subresource/consoleserver/types.go
```diff
@@ -79,3 +79,12 @@ type Customization struct {
 type Providers struct {
 	StatuspageID string `yaml:"statuspageID,omitempty"`
 }
+
```
benjaminapetersen: @pedjak fyi, we should keep this file in sync if you make future changes to the console config.

benjaminapetersen:
/approve
/lgtm

no need to backport.

PR #384, opened 2020-02-27
Bug 1806787: Expose openshift_console_operator_build_info metric

jhadvig:
FIxed operator for sending `openshift_console_operator_build_info` metrics. Also refactored the metrics itself.

/assign @benjaminapetersen


pkg/console/metrics/metrics.go
```diff
@@ -0,0 +1,88 @@
+package metrics
```
benjaminapetersen: I think I prefer leaving the file as `console_url.go`, as it contains a bunch of helpers only for `console_url`.

pkg/console/metrics/metrics.go
```diff
@@ -0,0 +1,88 @@
...
+func hostDied(old, new string) bool {
+	return len(old) != 0 && len(new) == 0
+}
+
+func RegisterVersion(major, minor, gitCommit, gitVersion string) {
+	defer recoverMetricPanic()
+	consoleBuildInfo.WithLabelValues(major, minor, gitCommit, gitVersion).Set(1)
+}
+
+// We will never want to panic our operator because of metric saving.
```
benjaminapetersen: This is a good add.

pkg/console/metrics/register.go
```diff
@@ -1,32 +0,0 @@
...
-type ConsoleMetrics struct {
-	ConsoleURL *k8smetrics.GaugeVec
-}
-
-var singleton *ConsoleMetrics
-var once sync.Once
-
-func Register() *ConsoleMetrics {
-	// thread safe
-	once.Do(func() {
```
benjaminapetersen: Reason to eliminate?

pkg/console/metrics/metrics.go
```diff
@@ -0,0 +1,88 @@
...
+	consoleBuildInfo = k8smetrics.NewGaugeVec(
+		&k8smetrics.GaugeOpts{
+			Name: "openshift_console_operator_build_info",
+			Help: "A metric with a constant '1' value labeled by major, minor, git commit & git version from which OpenShift Console Operator was built.",
+		},
+		[]string{"major", "minor", "gitCommit", "gitVersion"},
+	)
+)
+
+func init() {
```
benjaminapetersen: I believe I orinally used `init()` but then moved away from it as "best practices" indicated its better to do so (thought its been a while since I worked on this).  Timing is a potential problem that earlier examples didn't account for.

pkg/console/metrics/register.go
benjaminapetersen: I believe timing is a problem I was aiming to resolve,a lot of examples don't account for this.

pkg/console/metrics/metrics.go
jhadvig: @benjaminapetersen right but our `version.go` is still using  `init()`, also other operators (for which the metrics as build_info are working) are using this approach
Timing is not the best here I agree. So we have two options:
- try to fix it
- put on 4.5

benjaminapetersen: Fair enough, its late, this fixes the bug.  We can roll ahead, but prob should circle around and review approach again in 4.5.

benjaminapetersen:
/retest

```
 Mar  4 16:33:43.849: INFO:  > ERROR: (gcloud.compute.instance-groups.list-instances) could not parse resource []
Mar  4 16:33:43.849: INFO:  > 
Mar  4 16:33:43.849: INFO: Cluster image sources lookup failed: exit status 1
Mar  4 16:33:43.851: INFO: Waiting up to 30m0s for all (but 100) nodes to be schedulable
Mar  4 16:33:43.874: INFO: Waiting up to 10m0s for all pods (need at least 0) in namespace 'kube-system' to be running and ready
STEP: Destroying namespace "e2e-test-router-scoped-sbqcr" for this suite.
Mar  4 16:39:08.081: INFO: Running AfterSuite actions on all nodes
Mar  4 16:39:08.081: INFO: Running AfterSuite actions on node 1
fail [github.com/openshift/origin/test/extended/util/framework.go:1565]: Unexpected error:
    <*errors.errorString | 0xc0002d63f0>: {
        s: "timed out waiting for the condition",
    }
    timed out waiting for the condition
occurred 
```

PR #386, opened 2020-03-04
Bug 1813775: Publish license file of oc on downloads-openshift-console

eramoto:
Publish the license file with oc binary on downloads-openshift-console
since oc binary is applied Apache License 2.0.

This depends on openshift/oc#337 to provide the license in cli-artifacts.

### DIFF
--- a/manifests/07-downloads-deployment.yaml
+++ b/manifests/07-downloads-deployment.yaml
@@ -89,10 +89,14 @@ spec:
               os.mkdir(arch)
               for operating_system in ['linux', 'mac', 'windows']:
                   os.mkdir(os.path.join(arch, operating_system))
+                  target_path = os.path.join(arch, operating_system, 'LICENSE')
+                  os.symlink('/usr/share/openshift/LICENSE', target_path)
           for arch in ['arm64', 'ppc64le', 's390x']:
               os.mkdir(arch)
               for operating_system in ['linux']:
                   os.mkdir(os.path.join(arch, operating_system))
+                  target_path = os.path.join(arch, operating_system, 'LICENSE')
+                  os.symlink('/usr/share/openshift/LICENSE', target_path)
 
           for arch, operating_system, path in [
                   ('amd64', 'linux', '/usr/share/openshift/linux_amd64/oc'),
--- a/pkg/console/controllers/clidownloads/controller.go
+++ b/pkg/console/controllers/clidownloads/controller.go
@@ -188,6 +188,41 @@ The oc binary offers the same capabilities as the kubectl binary, but it is furt
 	}
 }
 
+func LicenseDownloads(host, cliDownloadsName string) *v1.ConsoleCLIDownload {
+	baseURL := fmt.Sprintf("%s", util.HTTPS(host))
+	platforms := []struct {
+		label    string
+		key      string
+		archType string
+	}{
+		{"Linux for x86_64", "amd64/linux", "LICENSE"},
+		{"Linux for ARM 64", "arm64/linux", "LICENSE"},
+		{"Linux for IBM Power, little endian", "ppc64le/linux", "LICENSE"},
+		{"Linux for IBM Z", "s390x/linux", "LICENSE"},
+		{"Mac for x86_64", "amd64/mac", "LICENSE"},
+		{"Windows for x86_64", "amd64/windows", "LICENSE"},
+	}
+
+	links := []v1.CLIDownloadLink{}
+	for _, platform := range platforms {
+		links = append(links, v1.CLIDownloadLink{
+			Href: GetPlatformURL(baseURL, platform.key, platform.archType),
+			Text: fmt.Sprintf("Download LICENSE at %s", platform.label),
+		})
+	}
+
+	return &v1.ConsoleCLIDownload{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: cliDownloadsName,
+		},
+		Spec: v1.ConsoleCLIDownloadSpec{
+			Description: `Apache License v2.0 for the OpenShift command line interface.`,
+			DisplayName: "LICENSE - license of OpenShift Command Line Interface",
+			Links:       links,
+		},
+	}
+}
+
 func ODOConsoleCLIDownloads() *v1.ConsoleCLIDownload {
 	return &v1.ConsoleCLIDownload{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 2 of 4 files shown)


manifests/07-downloads-deployment.yaml
```diff
@@ -89,10 +89,14 @@ spec:
               os.mkdir(arch)
               for operating_system in ['linux', 'mac', 'windows']:
                   os.mkdir(os.path.join(arch, operating_system))
+                  target_path = os.path.join(arch, operating_system, 'LICENSE')
```
benjaminapetersen: You may want to add a new `TestLicense()` to 
https://github.com/openshift/console-operator/blob/master/test/e2e/downloads_test.go

manifests/07-downloads-deployment.yaml
```diff
@@ -89,10 +89,14 @@ spec:
...
+                  os.symlink('/usr/share/openshift/LICENSE', target_path)
```
jhadvig: Is this path also valid for windows ? not really sure about that...

manifests/07-downloads-deployment.yaml
jhadvig: +1 on adding testes

manifests/07-downloads-deployment.yaml
eramoto: `target_path` is path after FQDN, so URL is `<scheme>://<FQDN>/<taret_path>`.
For windows, it is `https://<FQDN>/adm64/windows/LICENSE`.

manifests/07-downloads-deployment.yaml
eramoto: I added a new test `TestDonwloadsLicense()` and some tests for pkg.

The new commit have passed `make verify` and `make test-unit` on my local machine.

pkg/console/controllers/clidownloads/controller_test.go
```diff
@@ -148,3 +148,141 @@ The oc binary offers the same capabilities as the kubectl binary, but it is furt
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "amd64/oc-cli-downloads",
+				},
+				Spec: v1.ConsoleCLIDownloadSpec{
+					Description: `Apache License v2.0 for the OpenShift command line interface.`,
+					DisplayName: "LICENSE - license of OpenShift Command Line Interface",
+					Links: []v1.CLIDownloadLink{
+						{
+							Href: "https://www.example.com/amd64/linux/LICENSE",
+							Text: "Download LICENSE at Linux for x86_64",
```
benjaminapetersen, line 254: The link seems a little clunky, perhaps a text swap:

```
// before
Download LICENSE at Linux for x86_64
Download LICENSE at Linux for IBM Z
// after
Download LICENSE for Linux x86_64
Download LICENSE for Linux IBM Z
Download LICENSE for Windows x86_64
```

pkg/console/controllers/clidownloads/controller.go
```diff
@@ -188,6 +188,41 @@ The oc binary offers the same capabilities as the kubectl binary, but it is furt
...
+		{"Linux for IBM Power, little endian", "ppc64le/linux", "LICENSE"},
+		{"Linux for IBM Z", "s390x/linux", "LICENSE"},
+		{"Mac for x86_64", "amd64/mac", "LICENSE"},
+		{"Windows for x86_64", "amd64/windows", "LICENSE"},
+	}
+
+	links := []v1.CLIDownloadLink{}
+	for _, platform := range platforms {
+		links = append(links, v1.CLIDownloadLink{
+			Href: GetPlatformURL(baseURL, platform.key, platform.archType),
```
benjaminapetersen, line 209: Perhaps:
Download LICENSE for Linux x86_64
Download LICENSE for Linux IBM Z
Download LICENSE for Windows x86_64

benjaminapetersen:
/approve

but i think we need to tweak the link text

spadgett:
Long term, I think we should add a `licenseURL` property to the `ConsoleCLIDownload` resource so that the frontend can do a better job displaying it. But let's get the basic fix in.

pkg/console/controllers/clidownloads/controller.go
```diff
@@ -188,6 +188,41 @@ The oc binary offers the same capabilities as the kubectl binary, but it is furt
 	}
 }
 
+func LicenseDownloads(host, cliDownloadsName string) *v1.ConsoleCLIDownload {
```
benjaminapetersen, line 191: Also, this function is never called by the controller, so the license download is not being created.

manifests/07-downloads-deployment.yaml
benjaminapetersen: Ran it on a cluster?

pkg/console/controllers/clidownloads/controller.go
benjaminapetersen: You will need to add something like this to call it:

```golang

// up by the oc & odo calls
ocLicenseCLIDownloads := LicenseDownloads(host, api.OCCLIDownloadsCustomResourceName+"license")
_, ocLicenseCLIDownloadsReason, ocLicenseCLIDownloadsErr := ApplyCLIDownloads(c.consoleCliDownloadsClient, ocLicenseCLIDownloads)
status.HandleDegraded(updatedOperatorConfig, "OCDownloadsSync", ocLicenseCLIDownloadsReason, ocLicenseCLIDownloadsErr)
if ocLicenseCLIDownloadsErr != nil {
	return ocLicenseCLIDownloadsErr
}
```

benjaminapetersen:
With the addition of my latest comment to call the `License` func, this is what the cli downloads page will look like:

Screen Shot 2020-04-09 at 1 58 18 PM

Per @spadgett comments, this is probably acceptable, with the goal of instead of combining these into a single resource in the future:

Screen Shot 2020-04-09 at 2 01 31 PM

benjaminapetersen:
I'd suggest a tweak to have the link text just match `oc`:
```
Download oc for Linux for x86_64
Download LICENSE for Linux for x86_64 // rather than "at Linux"
```
Screen Shot 2020-04-09 at 2 06 52 PM

benjaminapetersen:
/hold

The license file doesn't actually exist. 
We should have an e2e test on this PR to verify that the payload is there before we merge.

Created #412 based on this branch to enable the CR (also on hold).

spadgett:
I wouldn't make this a separate `ConsoleCLIDownload` resource. I'd just add the additional license links to the existing `oc` download resource.

benjaminapetersen:
I tend to agree. Its a little strange having them separate resources

- [ ] Add e2e to verify that the license file exists (this will help with backports, we should be confident the files exist if we are going to backport the PR).
- [ ] Combine License into the existing oc downloads resource.

jhadvig:
Closing in favour of https://github.com/openshift/console-operator/pull/439

PR #402, opened 2020-03-26
Custom console route

jhadvig:
Adding option do specify the custom hostname for console.
When when custom hostname is specified a new `console-custom` route is created. In case the new hostname doesnt contain the cluster domain as a suffix, secret reference needs to be specified, that points to the secret in the `openshift-config` namespace, where it needs to be created manually and contain **tls** cert and its key, which will serve as a **serving-cert** for the console pod. Also we will maintain the default route so it stays reserved for us.

/assign @benjaminapetersen 

API changes - https://github.com/openshift/api/pull/594

### DIFF
--- a/manifests/03-rbac-role-ns-console.yaml
+++ b/manifests/03-rbac-role-ns-console.yaml
@@ -63,6 +63,7 @@ rules:
   - route.openshift.io
   resources:
   - routes
+  - routes/custom-host
   verbs:
   - get
   - list
--- a/manifests/03-rbac-role-ns-openshift-config.yaml
+++ b/manifests/03-rbac-role-ns-openshift-config.yaml
@@ -8,6 +8,7 @@ rules:
       - ""
     resources:
       - configmaps
+      - secrets
     verbs:
       - get
       - list
--- a/pkg/console/controllers/route/controller.go
+++ b/pkg/console/controllers/route/controller.go
@@ -4,12 +4,16 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
 	// k8s
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -21,6 +25,7 @@ import (
 	// openshift
 	operatorsv1 "github.com/openshift/api/operator/v1"
 	routev1 "github.com/openshift/api/route/v1"
+	configclientv1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	operatorclientv1 "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1"
 	v1 "github.com/openshift/client-go/operator/informers/externalversions/operator/v1"
 	routeclientv1 "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
@@ -31,7 +36,6 @@ import (
 	"github.com/openshift/console-operator/pkg/api"
 	customerrors "github.com/openshift/console-operator/pkg/console/errors"
 	"github.com/openshift/console-operator/pkg/console/status"
-	"github.com/openshift/console-operator/pkg/console/subresource/route"
 	routesub "github.com/openshift/console-operator/pkg/console/subresource/route"
 )
 
@@ -43,8 +47,10 @@ const (
 type RouteSyncController struct {
 	// clients
 	operatorConfigClient operatorclientv1.ConsoleInterface
+	ingressClient        configclientv1.IngressInterface
 	routeClient          routeclientv1.RoutesGetter
 	configMapClient      coreclientv1.ConfigMapsGetter
+	secretClient         coreclientv1.SecretsGetter
 	// names
 	targetNamespace string
 	routeName       string
@@ -57,10 +63,13 @@ type RouteSyncController struct {
 }
 
 func NewRouteSyncController(
+	// top level config
+	configClient configclientv1.ConfigV1Interface,
 	// clients
 	operatorConfigClient operatorclientv1.ConsoleInterface,
 	routev1Client routeclientv1.RoutesGetter,
 	configMapClient coreclientv1.ConfigMapsGetter,
+	secretClient coreclientv1.SecretsGetter,
 	// informers
 	operatorConfigInformer v1.ConsoleInformer,
 	routeInformer routesinformersv1.RouteInformer,
@@ -74,8 +83,10 @@ func NewRouteSyncController(
 ) *RouteSyncController {
 	ctrl := &RouteSyncController{
 		operatorConfigClient: operatorConfigClient,
+		ingressClient:        configClient.Ingresses(),
 		routeClient:          routev1Client,
 		configMapClient:      configMapClient,
+		secretClient:         secretClient,
 		targetNamespace:      targetNamespace,
 		routeName:            routeName,
 		// events
@@ -110,59 +121,184 @@ func (c *RouteSyncController) sync() error {
 		return nil
 	case operatorsv1.Removed:
 		klog.V(4).Infoln("console is in a removed state: deleting route")
-		return c.removeRoute()
+		if err = c.removeRoute(api.OpenshiftConsoleCustomRouteName); err != nil {
+			return err
+		}
+		return c.removeRoute(api.OpenShiftConsoleName)
 	default:
 		return fmt.Errorf("unknown state: %v", operatorConfig.Spec.ManagementState)
 	}
 
 	updatedOperatorConfig := operatorConfig.DeepCopy()
-	_, _, errReason, err := c.SyncRoute(updatedOperatorConfig)
 
-	status.HandleProgressingOrDegraded(updatedOperatorConfig, "RouteSync", errReason, err)
-	status.SyncStatus(c.ctx, c.operatorConfigClient, updatedOperatorConfig)
+	defaultRouteErrReason, defaultRouteErr := c.SyncDefaultRoute(updatedOperatorConfig)
+	status.HandleProgressingOrDegraded(updatedOperatorConfig, "DefaultRouteSync", defaultRouteErrReason, defaultRouteErr)
+	if defaultRouteErr != nil {
+		return defaultRouteErr
+	}
+
+	customRouteErrReason, customRouteErr := c.SyncCustomRoute(updatedOperatorConfig)
+	status.HandleProgressingOrDegraded(updatedOperatorConfig, "CustomRouteSync", customRouteErrReason, customRouteErr)
 
 	return err
 }
 
-func (c *RouteSyncController) removeRoute() error {
-	klog.V(2).Info("deleting console route")
-	defer klog.V(2).Info("finished deleting console route")
-	return c.routeClient.Routes(c.targetNamespace).Delete(c.ctx, route.Stub().Name, metav1.DeleteOptions{})
+func (c *RouteSyncController) removeRoute(routeName string) error {
+	err := c.routeClient.Routes(c.targetNamespace).Delete(c.ctx, routeName, metav1.DeleteOptions{})
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+	return err
 }
 
-// apply route
-// - be sure to test that we don't trigger an infinite loop by stomping on the
-//   default host name set by the server, or any other values. The ApplyRoute()
-//   logic will have to be sound.
-// - update to ApplyRoute() once the logic is settled
-func (c *RouteSyncController) SyncRoute(operatorConfig *operatorsv1.Console) (consoleRoute *routev1.Route, isNew bool, reason string, err error) {
-	// ensure we have a route. any error returned is a non-404 error
+func (c *RouteSyncController) SyncDefaultRoute(operatorConfig *operatorsv1.Console) (string, error) {
+	requiredDefaultRoute := routesub.DefaultRoute(operatorConfig)
 
-	rt, rtIsNew, rtErr := routesub.GetOrCreate(c.ctx, c.routeClient, routesub.DefaultRoute(operatorConfig))
-	if rtErr != nil {
-		return nil, false, "FailedCreate", rtErr
-	}
-	// Check if the console is reachable
-	c.CheckRouteHealth(operatorConfig, rt)
-
-	// we will not proceed until the route is valid. this eliminates complexity with the
-	// configmap, secret & oauth client as they can be certain they have a host if we pass this point.
-	host := routesub.GetCanonicalHost(rt)
-	if len(host) == 0 {
-		return nil, false, "FailedHost", customerrors.NewSyncError(fmt.Sprintf("route is not available at canonical host %s", rt.Status.Ingress))
+	defaultRoute, _, defaultRouteError := routesub.ApplyRoute(c.routeClient, c.recorder, requiredDefaultRoute)
+	if defaultRouteError != nil {
+		return "FailedDefaultRouteApply", defaultRouteError
 	}
 
-	if validatedRoute, changed := routesub.Validate(rt); changed {
-		// if validation changed the route, issue an update
-		if _, err := c.routeClient.Routes(api.TargetNamespace).Update(c.ctx, validatedRoute, metav1.UpdateOptions{}); err != nil {
-			// error is unexpected, this is a real error
-			return nil, false, "InvalidRouteCorrection", err
+	if len(routesub.GetCanonicalHost(defaultRoute)) == 0 {
+		return "FailedDefaultRouteHost", customerrors.NewSyncError(fmt.Sprintf("default route is not available at canonical host %s", defaultRoute.Status.Ingress))
+	}
+
+	if !routesub.IsCustomRouteSet(operatorConfig) {
+		c.CheckRouteHealth(operatorConfig, defaultRoute)
+	}
+
+	return "", defaultRouteError
+}
+
+// Custom route sync needs to:
+// 1. validate if the reference for secret with TLS certificate and key is defined in operator config(in case a non-openshift cluster domain is used)
+// 2. if secret is defined, verify the TLS certificate and key
+// 4. create the custom console route, if custom TLS certificate and key are defined use them
+// 5. apply the custom route
+func (c *RouteSyncController) SyncCustomRoute(operatorConfig *operatorsv1.Console) (string, error) {
+	if !routesub.IsCustomRouteSet(operatorConfig) {
+		return "", nil
+	}
+
+	customSecret, configErr := c.ValidateCustomRouteConfig(operatorConfig)
+	if configErr != nil {
+		return "InvalidCustomRouteConfig", configErr
+	}
+
+	customTLSCert, secretValidationErr := ValidateCustomCertSecret(customSecret)
+	if secretValidationErr != nil {
+		return "InvalidCustomTLSSecret", secretValidationErr
+	}
+
+	requiredCustomRoute := routesub.CustomRoute(operatorConfig, customTLSCert)
+	customRoute, _, customRouteError := routesub.ApplyRoute(c.routeClient, c.recorder, requiredCustomRoute)
+	if customRouteError != nil {
+		return "FailedCustomRouteApply", customRouteError
+	}
+
+	if len(routesub.GetCanonicalHost(customRoute)) == 0 {
+		return "FailedCustomRouteHost", customerrors.NewSyncError(fmt.Sprintf("custom route is not available at canonical host %s", customRoute.Status.Ingress))
+	}
+
+	if routesub.IsCustomRouteSet(operatorConfig) {
+		c.CheckRouteHealth(operatorConfig, customRoute)
+	}
+
+	return "", customRouteError
+}
+
+func (c *RouteSyncController) ValidateCustomRouteConfig(operatorConfig *operatorsv1.Console) (*corev1.Secret, error) {
+	// get ingress
+	ingress, err := c.ingressClient.Get(c.ctx, api.ConfigResourceName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	// Check if the custom hostname has cluster domain suffix, which indicates
+	// if a secret that contains TLS certificate and key needs to exist in the
+	// `openshift-config` namespace and referenced in  the operator config.
+	// If the suffix matches the cluster domain, then the secret is optional.
+	// If the suffix doesn't matches the cluster domain, then the secret is mandatory.
+	if !routesub.IsCustomRouteSecretSet(operatorConfig) {
+		if !strings.HasSuffix(operatorConfig.Spec.Route.Hostname, ingress.Spec.Domain) {
+			return nil, fmt.Errorf("secret reference for custom route TLS secret is not defined")
 		}
-		// abort sync, route changed, let it settle & retry
-		return nil, true, "InvalidRoute", customerrors.NewSyncError("route is invalid, correcting route state")
+		return nil, nil
 	}
-	// only return the route if it is valid with a host
-	return rt, rtIsNew, "", rtErr
+
+	return c.secretClient.Secrets(api.OpenShiftConfigNamespace).Get(c.ctx, operatorConfig.Spec.Route.Secret.Name, metav1.GetOptions{})
+}
+
+// Validate secret that holds custom TLS certificate and key.
+// Secret has to contain `tls.crt` and `tls.key` data keys
+// where the certificate and key are stored and both need
+// to be in valid format.
+// Return the custom TLS certificate and key
+func ValidateCustomCertSecret(customCertSecret *corev1.Secret) (*routesub.CustomTLSCert, error) {
+	if customCertSecret == nil {
+		return nil, nil
+	}
+	if customCertSecret.Type != corev1.SecretTypeTLS {
+		return nil, fmt.Errorf("custom cert secret is not in %q type, instead uses %q type", corev1.SecretTypeTLS, customCertSecret.Type)
+	}
+
+	customTLS := &routesub.CustomTLSCert{}
+	cert, certExist := customCertSecret.Data["tls.crt"]
+	if !certExist {
+		return nil, fmt.Errorf("custom cert secret data doesn't contain 'tls.crt' entry")
+	}
+
+	certificateVerifyErr := certificateVerifier(cert)
+	if certificateVerifyErr != nil {
+		return nil, fmt.Errorf("failed to verify custom certificate PEM: " + certificateVerifyErr.Error())
+	}
+	customTLS.Certificate = string(cert)
+
+	key, keyExist := customCertSecret.Data["tls.key"]
+	if !keyExist {
+		return nil, fmt.Errorf("custom cert secret data doesn't contain 'tls.key' entry")
+	}
+
+	privateKeyVerifyErr := privateKeyVerifier(key)
+	if privateKeyVerifyErr != nil {
+		return nil, fmt.Errorf("failed to verify custom key PEM: " + privateKeyVerifyErr.Error())
+	}
+	customTLS.Key = string(key)
+
+	return customTLS, nil
+}
+
+func certificateVerifier(customCert []byte) error {
+	block, _ := pem.Decode([]byte(customCert))
+	if block == nil {
+		return fmt.Errorf("failed to decode certificate PEM")
+	}
+	certificate, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return err
+	}
+	now := time.Now()
+	if now.After(certificate.NotAfter) {
+		return fmt.Errorf("custom TLS certificate is expired")
+	}
+	if now.Before(certificate.NotBefore) {
+		return fmt.Errorf("custom TLS certificate is not valid yet")
+	}
+	return nil
+}
+
+func privateKeyVerifier(customKey []byte) error {
+	block, _ := pem.Decode([]byte(customKey))
+	if block == nil {
+		return fmt.Errorf("failed to decode key PEM")
+	}
+	if _, err := x509.ParsePKCS8PrivateKey(block.Bytes); err != nil {
+		if _, err = x509.ParsePKCS1PrivateKey(block.Bytes); err != nil {
+			if _, err = x509.ParseECPrivateKey(block.Bytes); err != nil {
+				return fmt.Errorf("block %s is not valid key PEM", block.Type)
+			}
+		}
+	}
+	return nil
 }
 
 func (c *RouteSyncController) Run(workers int, stopCh <-chan struct{}) {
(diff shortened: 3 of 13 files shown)


pkg/console/controllers/route/controller.go
```diff
@@ -104,12 +109,23 @@ func (c *RouteSyncController) sync() error {
...
+		if err = c.removeCustomRoute(); err != nil {
+			return err
+		}
+		return c.removeRoute(api.OpenShiftConsoleName)
 	default:
 		return fmt.Errorf("unknown state: %v", operatorConfig.Spec.ManagementState)
 	}
 
 	updatedOperatorConfig := operatorConfig.DeepCopy()
+
```
benjaminapetersen: This high up in the controller, I'm a bit inclined to always call both functions, but hide away the conditional details within them:

```golang
// just always call SyncCustom, but within the func return early if the custom route IS NOT being used:
//   if ! customRoute { return }
_, _, customErrReason, err := c.SyncCustomRoute(updatedOperatorConfig)
status.HandleProgressingOrDegraded(updatedOperatorConfig, "CustomRouteSync", customErrReason, err)
// and always call default, but return early if the custom route IS being used:
// the return early on this one is just the health check, like you have, since we are still creating the route.
// if customRoute { return }
_, _, defaultErrReason, err := c.SyncRoute(updatedOperatorConfig)
status.HandleProgressingOrDegraded(updatedOperatorConfig, "RouteSync", defaultErrReason, err)
status.SyncStatus(c.operatorConfigClient, updatedOperatorConfig)

return err;
```

pkg/console/controllers/route/controller.go
```diff
@@ -159,6 +186,48 @@ func (c *RouteSyncController) SyncRoute(operatorConfig *operatorsv1.Console) (co
...
+// 2. sync the secret with the serving-cert, from the openshift-config,into the openshift-console namespace
+// 3. create the custom console route
+// 4. check custom route health
+// 5. validate the route
+func (c *RouteSyncController) SyncCustomRoute(operatorConfig *operatorsv1.Console) (consoleRoute *routev1.Route, isNew bool, reason string, err error) {
+	if configErr := routesub.ValidateCustomRoute(operatorConfig, c.routeClient); configErr != nil {
+		return nil, true, "InvalidCustomRouteConfig", configErr
+	}
+
+	if scscErr := c.SyncCustomServigCertSecret(operatorConfig); scscErr != nil {
```
benjaminapetersen: nit: `SyncCustomServigCertSecret` has `Servig`, but should be `Serving`.

pkg/console/controllers/route/controller.go
```diff
@@ -159,6 +186,48 @@ func (c *RouteSyncController) SyncRoute(operatorConfig *operatorsv1.Console) (co
...
+	}
+
+	cr, crIsNew, crErr := routesub.GetOrCreate(c.routeClient, routesub.CustomRoute(operatorConfig))
+	if crErr != nil {
+		return nil, false, "FailedCreateCustomRoute", crErr
+	}
+	c.CheckRouteHealth(operatorConfig, cr)
+
+	if _, changed := routesub.Validate(cr); changed {
+		return nil, true, "InvalidCustomRoute", customerrors.NewSyncError("custom route is invalid, correcting route state")
```
benjaminapetersen: So this deviates a little bit from the `Validate()` in `SyncRoute()`.  Looking into this a bit to make sure its ok.

pkg/console/controllers/route/controller.go
```diff
@@ -159,6 +186,48 @@ func (c *RouteSyncController) SyncRoute(operatorConfig *operatorsv1.Console) (co
...
+	}
+
+	return cr, crIsNew, "", crErr
+}
+
```
benjaminapetersen: This should be good to update the resource sync source.

pkg/console/controllers/route/controller.go
```diff
@@ -159,6 +186,48 @@ func (c *RouteSyncController) SyncRoute(operatorConfig *operatorsv1.Console) (co
...
+func (c *RouteSyncController) SyncCustomServigCertSecret(operatorConfig *operatorsv1.Console) error {
+	source := resourcesynccontroller.ResourceLocation{}
+
+	if routesub.IsCustomRouteSecretSet(operatorConfig) {
+		source.Name = operatorConfig.Spec.Route.Secret.Name
+		source.Namespace = api.OpenShiftConfigNamespace
+	}
+	// if no secret with serving-cert provided, sync an empty source to delete
+	return c.resourceSyncer.SyncSecret(
+		resourcesynccontroller.ResourceLocation{Namespace: api.OpenShiftConsoleNamespace, Name: api.OpenShiftCustomLogoConfigMapName},
```
benjaminapetersen: Prob should update `api.OpenShiftCustomLogoConfigMapName` to `api.CustomRouteServingCertName` (but can be the same value reference).

pkg/console/controllers/route/controller.go
benjaminapetersen: I'm also leaning towards

```golang
c.SyncDefaultRoute // adding default for future clarity
c.SyncCustomRoute
```

pkg/console/subresource/route/route.go
```diff
@@ -88,6 +101,41 @@ func Validate(route *routev1.Route) (*routev1.Route, bool) {
...
+
+// Check if reference for secret holding custom serving-cert is set
+func IsCustomRouteSecretSet(operatorConfig *operatorv1.Console) bool {
+	if len(operatorConfig.Spec.Route.Secret.Name) != 0 {
+		return true
+	}
+	return false
+}
+
+func ValidateCustomRoute(operatorConfig *operatorv1.Console, client routeclient.RoutesGetter) error {
```
benjaminapetersen: I'm not sure I follow the logic entirely here. 
Its not really validating the custom route its.... checking if the custom is still on the default ingress domain?

pkg/console/subresource/route/route.go
```diff
@@ -88,6 +101,41 @@ func Validate(route *routev1.Route) (*routev1.Route, bool) {
...
+	defaultRoute, err := client.Routes(api.OpenShiftConsoleNamespace).Get(api.OpenShiftConsoleRouteName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	if is := isClusterDomain(operatorConfig.Spec.Route.Hostname, defaultRoute.Spec.Host); !is {
```
benjaminapetersen: The question here is, do we trust the `default route`, or should we pull in the `Ingress Config`?

pkg/console/subresource/route/route.go
```diff
@@ -88,6 +101,41 @@ func Validate(route *routev1.Route) (*routev1.Route, bool) {
...
+		if IsCustomRouteSecretSet(operatorConfig) {
+			return fmt.Errorf("secret reference for custom serving-cert is not defined")
+		}
+	}
+
+	return err
+}
+
+func isClusterDomain(customRouteHost, defaultRouteHost string) bool {
```
benjaminapetersen: So this is answering "Is the custom host still on the default ingress domain"?

pkg/console/subresource/route/route.go
```diff
@@ -88,6 +101,41 @@ func Validate(route *routev1.Route) (*routev1.Route, bool) {
 	return route, changed
 }
 
+func IsCustomRouteSet(operatorConfig *operatorv1.Console) bool {
```
benjaminapetersen: I think this will work:
```golang
func IsCustomRouteSet(operatorConfig *operatorv1.Console) bool {
	return operatorConfig.Spec.Route.Hostname) != 0
}
```

pkg/console/subresource/route/route.go
```diff
@@ -88,6 +101,41 @@ func Validate(route *routev1.Route) (*routev1.Route, bool) {
...
+	if len(operatorConfig.Spec.Route.Hostname) != 0 {
+		return true
+	}
+	return false
+}
+
+// Check if reference for secret holding custom serving-cert is set
+func IsCustomRouteSecretSet(operatorConfig *operatorv1.Console) bool {
```
benjaminapetersen:
```golang
func IsCustomRouteSecretSet(operatorConfig *operatorv1.Console) bool {
	return len(operatorConfig.Spec.Route.Secret.Name) != 0
}
```

pkg/console/subresource/deployment/deployment.go
```diff
@@ -461,3 +464,12 @@ func customLogoVolume() volumeConfig {
 		path:        "/var/logo/",
 		isConfigMap: true}
 }
+
```
benjaminapetersen: This is to overwrite the existing serving cert with the custom, so we don't have to change the impl in the console to look at a different dir, correct?

pkg/console/controllers/route/controller.go
```diff
@@ -159,6 +186,48 @@ func (c *RouteSyncController) SyncRoute(operatorConfig *operatorsv1.Console) (co
...
 }
 
+// Custom route sync needs to:
+// 1. validate if the reference for secret with serving-cert is defined in operator config(in case a non-openshift cluster domain is used)
+// 2. sync the secret with the serving-cert, from the openshift-config,into the openshift-console namespace
+// 3. create the custom console route
+// 4. check custom route health
+// 5. validate the route
+func (c *RouteSyncController) SyncCustomRoute(operatorConfig *operatorsv1.Console) (consoleRoute *routev1.Route, isNew bool, reason string, err error) {
+	if configErr := routesub.ValidateCustomRoute(operatorConfig, c.routeClient); configErr != nil {
```
stlaz: As Ben suggested, move the `IsCustomRouteSet()` check here and return early if false

pkg/console/subresource/route/route.go
jhadvig: So I think we can trust the default route, since we will be maintaining it anyways. Also when creating we are using the `default` as the router name. Otherwise we are raising a sync error.
Should be ok then ?

pkg/console/subresource/route/route.go
jhadvig: yes

pkg/console/subresource/route/route.go
jhadvig: definitely 👍

pkg/console/subresource/deployment/deployment.go
jhadvig: yes
Thought that we should not need both serving-certs

pkg/console/controllers/route/controller.go
```diff
@@ -159,6 +186,48 @@ func (c *RouteSyncController) SyncRoute(operatorConfig *operatorsv1.Console) (co
...
+	c.CheckRouteHealth(operatorConfig, cr)
+
+	if _, changed := routesub.Validate(cr); changed {
+		return nil, true, "InvalidCustomRoute", customerrors.NewSyncError("custom route is invalid, correcting route state")
+	}
+
+	return cr, crIsNew, "", crErr
+}
+
+func (c *RouteSyncController) SyncCustomServigCertSecret(operatorConfig *operatorsv1.Console) error {
```
stlaz: why are you syncing this? Also, I'd advise to use the name "RouteCert" instead of serving-cert because the latter is more commonly used for the certificate the pod uses for serving, rather than the certificate the route uses.

pkg/console/subresource/route/route.go
jhadvig: yeah, probably bad naming hare. Maybe `isDefaultIngressDomain` ?

pkg/console/controllers/route/controller.go
jhadvig: yep ! will update !

pkg/console/subresource/deployment/deployment.go
```diff
@@ -84,6 +84,9 @@ func DefaultDeployment(operatorConfig *operatorv1.Console, cm *corev1.ConfigMap,
 	gracePeriod := int64(40)
 	tolerationSeconds := int64(120)
 	volumeConfig := defaultVolumeConfig()
+	if routesub.IsCustomRouteSecretSet(operatorConfig) {
+		volumeConfig = useCustomServingCertSecret(operatorConfig.Spec.Route.Secret.Name, volumeConfig)
```
stlaz: why are you mounting this? The contents are supposed to be used in the route's `spec.tls.termination.{cert,key,caCertificate}` fields

pkg/console/subresource/route/route.go
```diff
@@ -39,24 +43,33 @@ func GetOrCreate(client routeclient.RoutesGetter, required *routev1.Route) (*rou
...
 
+func CustomRoute(cr *operatorv1.Console) *routev1.Route {
+	route := &routev1.Route{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        api.OpenshiftConsoleCustomRouteName,
+			Namespace:   api.OpenShiftConsoleNamespace,
+			Labels:      util.SharedLabels(),
+			Annotations: map[string]string{},
+		},
+		Spec: spec(),
```
stlaz: where's all the fancy certificate stuff you're doing this for?

pkg/console/controllers/route/controller.go
jhadvig: thought that since the `Validate()` is just checking the `To`, `Port`, `TLS`, `WildcardPolicy` then it cant hurt to check those on the custom route, since these should be the same

pkg/console/controllers/route/controller.go
jhadvig: you mean that here we should call the resource sync for syncing the custom Secret with serving-cert?

pkg/console/subresource/deployment/deployment.go
stlaz: Also, you should have a hard requirement on the secret you're syncing to contain all of `tls.crt`, `tls.key` and `ca.crt` fields

pkg/console/controllers/route/controller.go
```diff
@@ -118,10 +134,18 @@ func (c *RouteSyncController) sync() error {
...
 
-func (c *RouteSyncController) removeRoute() error {
-	klog.V(2).Info("deleting console route")
-	defer klog.V(2).Info("finished deleting console route")
-	return c.routeClient.Routes(c.targetNamespace).Delete(route.Stub().Name, &metav1.DeleteOptions{})
+func (c *RouteSyncController) removeRoute(routeName string) error {
+	klog.V(2).Infof("deleting %s route", routeName)
+	defer klog.V(2).Infof("finished deleting %s route", routeName)
+	return c.routeClient.Routes(c.targetNamespace).Delete(routeName, &metav1.DeleteOptions{})
+}
```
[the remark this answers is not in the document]
stlaz: actually, you would want to remove the other log line as well as that would appear on every sync even if the route was already removed...

jhadvig:
@stlaz so to point of having two routes is to have the default one reserved, in case that somebody could create one which would block admitting the default route if recreated. The default one will be active only in case that the custom route suffix will match the cluster domain

pkg/console/controllers/route/controller.go
jhadvig: > why are you syncing this?

we need to since the secret that contains the custom serving-cert from `openshift-config`, where the user/admin needs to create it manually, into the `openshift-console`, so we can mount it into the `console` pod

pkg/console/subresource/deployment/deployment.go
jhadvig: > why are you mounting this? The contents are supposed to be used in the route's spec.tls.termination.{cert,key,caCertificate} fields

I though that the secret needs to be mounted into the `console` pod.

> why are you mounting this? The contents are supposed to be used in the route's spec.tls.termination.{cert,key,caCertificate} fields

also `ca.crt`?
all of them needs to be used in the route's `spec.tls.termination.{cert,key,caCertificate}` fields?

stlaz: I think you would want to have the CA certificate there as well to prevent issues that the routing team has - https://github.com/openshift/cluster-ingress-operator/pull/329. Generally, it would make your life easier.

Also, you should validate the secret and don't progress further if it's invalid.

pkg/console/controllers/route/controller.go
stlaz: But you don't want to mount it in the pod, you want to have that inserted into the route

jhadvig:
@stlaz @benjaminapetersen I've update the PR based on your comments.
Removed mounting the secret as a `serving-cert` into the `console` pod and instead setting the cert and key on the custom route.

PTAL

jhadvig:
The unit tests are failing cause will update them after after we align on the logic, since that will be pretty big change itself

pkg/console/operator/sync_v400.go
```diff
@@ -57,12 +57,23 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console,
...
+
+	// Need to make difference of the default and active route due to the custom console hostname.
+	activeRoute := defaultRoute
+	if routesub.IsCustomRouteSet(updatedOperatorConfig) {
+		customRoute, customRouteErr := co.routeClient.Routes(api.OpenShiftConsoleNamespace).Get(api.OpenshiftConsoleCustomRouteName, metav1.GetOptions{})
+		status.HandleProgressingOrDegraded(updatedOperatorConfig, "CustomRoutePropagation", "FailedGetCustomRoute", customRouteErr)
+		if customRouteErr != nil {
+			return customRouteErr
+		}
+		activeRoute = customRoute
```
stlaz, line 73:
```go
routeName := api.OpenShiftConsoleName
if routesub.IsCustomRouteSet(updatedOperatorConfig) {
    routeName := api.OpenShiftConsoleCustomRouteName
}

doMagicHere()
```
also, please don't use `rt` as a variable name for `route`, `rt` usually stands for `roundTripper`, this could be confusing a random passerby.

jhadvig:
@benjaminapetersen @stlaz comments addressed. PTAL

pkg/console/controllers/route/controller.go
```diff
@@ -110,59 +122,151 @@ func (c *RouteSyncController) sync() error {
 		return nil
 	case operatorsv1.Removed:
 		klog.V(4).Infoln("console is in a removed state: deleting route")
-		return c.removeRoute()
+		if err = c.removeRoute(api.OpenshiftConsoleCustomRouteName); err != nil {
```
benjaminapetersen, line 125: This is still fine if it 404?  IE, set to `Removed` when the custom route is not in use?

pkg/console/controllers/route/controller.go
```diff
@@ -110,59 +122,151 @@ func (c *RouteSyncController) sync() error {
...
+}
+
+func (c *RouteSyncController) ValidateCustomRouteConfig(operatorConfig *operatorsv1.Console) (*corev1.Secret, error) {
+	// get ingress
+	ingress, err := c.ingressClient.Get(c.ctx, api.ConfigResourceName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	// Check if the custom hostname has cluster domain suffix. If it does,
+	// then the secret that contains TLS certificate and key needs to be
```
benjaminapetersen, line 213: I'm a little confused by the comment, it implies something should be done, but it immediately returns `nil, nil`.

pkg/console/controllers/route/controller.go
```diff
@@ -110,59 +122,151 @@ func (c *RouteSyncController) sync() error {
...
+	if err != nil {
+		return nil, err
+	}
+	return customCertSecret, nil
+}
+
+// Validate secret that holds custom TLS certificate and key.
+// Secret has to contain `tls.crt` and `tls.key` data keys
+// where the certificate and key are stored.
+// Return the custom TLS certificate and key
```
benjaminapetersen, line 231: Nice!  Tested this with a dummy bad file?

pkg/console/starter/starter.go
```diff
@@ -232,10 +232,13 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 	)
 
 	consoleRouteController := route.NewRouteSyncController(
+		// top level config
+		configClient.ConfigV1(),
 		// operator config
 		operatorConfigClient.OperatorV1().Consoles(),
 		routesClient.RouteV1(),
 		kubeClient.CoreV1(),
+		kubeClient.CoreV1(),
```
benjaminapetersen, line 241: I think we only need to pass `kubeClient.CoreV1()` once.

pkg/console/controllers/route/controller.go
```diff
@@ -57,10 +64,13 @@ type RouteSyncController struct {
 }
 
 func NewRouteSyncController(
+	// top level config
+	configClient configclientv1.ConfigV1Interface,
 	// clients
 	operatorConfigClient operatorclientv1.ConsoleInterface,
 	routev1Client routeclientv1.RoutesGetter,
 	configMapClient coreclientv1.ConfigMapsGetter,
+	secretClient coreclientv1.SecretsGetter,
```
benjaminapetersen, line 73: Nit: We probably could fix this so that `NewRouteSyncController()` accepts the `KubeV1Client` once, but then separates them out when we create the controller:

```golang
ctrl := &RouteSyncController{
   configMapClient:   KubeV1Client.ConfigMaps()
   secretsClient:  KubeV1Client.Secrets()
```
But I see our other controllers have the same issue.  It would be a tad cleaner if:

```golang
// We are a little incorrect as the secrets client is 1 step into the KubeV1Client as .Secrets
c.secretClient.Secrets(api.OpenShiftConfigNamespace).Get(c.ctx, operatorConfig.Spec.Route.Secret.Name, metav1.GetOptions{})
// This would be ideal
c.secretClient(api.OpenShiftConfigNamespace).Get(c.ctx, operatorConfig.Spec.Route.Secret.Name, metav1.GetOptions{})
```

benjaminapetersen: So we can chip away at it, or just fix them all as a separate PR.

pkg/console/starter/starter.go
benjaminapetersen: Ah, I see what we did, we have our clients slightly misnamed.  Hmm. And we do it in all our controllers. :)

pkg/console/subresource/route/route.go
```diff
@@ -24,6 +30,49 @@ const (
...
 )
 
+// holds information about custom TLS certificate and its key
+type CustomTLSCert struct {
+	Certificate string
+	Key         string
+}
+
+func ApplyRoute(client routeclient.RoutesGetter, recorder events.Recorder, required *routev1.Route) (*routev1.Route, bool, error) {
+	existing, err := client.Routes(required.Namespace).Get(context.TODO(), required.Name, metav1.GetOptions{})
```
benjaminapetersen, line 40: This is interesting:

```golang
client.Routes(required.Namespace).Get(context.TODO()
```
The `context.TODO()` throws me.

pkg/console/subresource/route/route.go
```diff
@@ -24,6 +30,49 @@ const (
...
+		// reportCreateEvent(recorder, requiredCopy, err)
+		return actual, true, err
+	}
+	if err != nil {
+		return nil, false, err
+	}
+	existingCopy := existing.DeepCopy()
+
+	if len(GetCanonicalHost(existingCopy)) == 0 {
+		return nil, false, customerrors.NewSyncError(fmt.Sprintf("route is not available at canonical host %s", existingCopy.Status.Ingress))
```
benjaminapetersen, line 54: Does this apply, even when using a custom route? `route is not available at canonical host %s`

pkg/console/subresource/route/route.go
```diff
@@ -41,53 +90,41 @@ func GetOrCreate(ctx context.Context, client routeclient.RoutesGetter, required
...
-	if tlsSame := equality.Semantic.DeepEqual(route.Spec.TLS, tls()); !tlsSame {
-		changed = true
-		route.Spec.TLS = tls()
-	}
-
-	if wildcardSame := equality.Semantic.DeepEqual(route.Spec.WildcardPolicy, wildcard()); !wildcardSame {
-		changed = true
-		route.Spec.WildcardPolicy = wildcard()
+func CustomRoute(cr *operatorv1.Console, tlsConfig *CustomTLSCert) *routev1.Route {
+	route := &routev1.Route{
```
benjaminapetersen, line 112: Nit:  could probably use the `DefaultStub()` and then just override the `.Name`.

pkg/console/subresource/route/route.go
benjaminapetersen: Must be the new norm?

stlaz: It came with kube 1.18 which adopted this from Go:

https://golang.org/pkg/context:
```
Incoming requests to a server should create a Context, and outgoing calls
to servers should accept a Context. The chain of function calls between
them must propagate the Context, optionally replacing it with a derived
Context created using WithCancel, WithDeadline, WithTimeout, or WithValue.
When a Context is canceled, all Contexts derived from it are also canceled. 
```

pkg/console/subresource/route/route.go
```diff
@@ -24,6 +30,49 @@ const (
...
+	Key         string
+}
+
+func ApplyRoute(client routeclient.RoutesGetter, recorder events.Recorder, required *routev1.Route) (*routev1.Route, bool, error) {
+	existing, err := client.Routes(required.Namespace).Get(context.TODO(), required.Name, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		requiredCopy := required.DeepCopy()
+		actual, err := client.Routes(requiredCopy.Namespace).
+			Create(context.TODO(), resourcemerge.WithCleanLabelsAndAnnotations(requiredCopy).(*routev1.Route), metav1.CreateOptions{})
+		// reportCreateEvent(recorder, requiredCopy, err)
```
stlaz, line 45: remove?

pkg/console/subresource/route/route.go
```diff
@@ -24,6 +30,49 @@ const (
...
+
+	if specSame && !*modified {
+		klog.V(4).Infof("%s route exists and is in the correct state", existingCopy.ObjectMeta.Name)
+		return existingCopy, false, nil
+	}
+
+	existingCopy.Spec = required.Spec
+
+	if klog.V(4) {
+		klog.Infof("Route %q changes: %v", required.Namespace+"/"+required.Name, resourceapply.JSONPatchNoError(existing, required))
```
stlaz, line 69: that's a CVE, logging private key

pkg/console/controllers/route/controller.go
jhadvig: If the custom route is not used then the it would just return 404 which `removeRoute()` takes care of. We could add the check for `IsCustomRouteSet()`  in this case

pkg/console/subresource/route/route.go
jhadvig: yeah this came from k8s. we can either solve it like this or pass the context as argument to the  `ApplyRoute(ctx context.Context, ...)`

pkg/console/subresource/route/route.go
jhadvig: It should apply to custom route as well. The `routerName` should be set by the `ingressController` to default as well and the route needs to be admitted also

pkg/console/controllers/route/controller.go
jhadvig: about to add !

pkg/console/controllers/route/controller.go
stlaz: but the user may still want their certificate to be used instead of the ones that the router adds by default, correct?

pkg/console/controllers/route/controller.go
```diff
@@ -110,59 +122,151 @@ func (c *RouteSyncController) sync() error {
...
+func (c *RouteSyncController) ValidateCustomRouteConfig(operatorConfig *operatorsv1.Console) (*corev1.Secret, error) {
+	// get ingress
+	ingress, err := c.ingressClient.Get(c.ctx, api.ConfigResourceName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	// Check if the custom hostname has cluster domain suffix. If it does,
+	// then the secret that contains TLS certificate and key needs to be
+	// specified on the operator config.
+	if strings.HasSuffix(operatorConfig.Spec.Route.Hostname, ingress.Spec.Domain) {
```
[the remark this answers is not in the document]
jhadvig: Then it wont be admitted.

pkg/console/controllers/route/controller.go
```diff
@@ -110,59 +122,151 @@ func (c *RouteSyncController) sync() error {
...
+	for _, key := range dataKeys {
+		tls, certExist := customCertSecret.Data[key]
+		if !certExist {
+			return nil, fmt.Errorf("custom cert secret data doesn't contain %q entry", key)
+		}
+
+		block, _ := pem.Decode([]byte(tls))
+		if block == nil {
+			return nil, fmt.Errorf("failed to parse custom certificate")
+		}
```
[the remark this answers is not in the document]
jhadvig: you mean the `nil` check ?

pkg/console/controllers/route/controller.go
jhadvig: would rather deal with it in one PR, if thats ok :)

jhadvig:
@stlaz @benjaminapetersen comments addressed and added.. PTAL :)

manifests/03-rbac-role-ns-console.yaml
```diff
@@ -63,6 +63,7 @@ rules:
   - route.openshift.io
   resources:
   - routes
+  - routes/custom-host
```
jhadvig, line 66: This is needed to set the `route.spec.tls` fields

pkg/console/controllers/route/controller.go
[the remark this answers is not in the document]
stlaz: I thought you're going to use the client-go keyutil which doesn't require PEM decoding, but this is fine,, then :)

pkg/console/subresource/route/route.go
stlaz: this is no place for this check though, you should be able to apply the route no matter what its status is.

Also, are you 100% sure that this will be valid even for routes outside the cluster domain?

pkg/console/controllers/route/controller_test.go
```diff
@@ -0,0 +1,245 @@
...
+
+	// console-operator
+
+	"github.com/go-test/deep"
+	routesub "github.com/openshift/console-operator/pkg/console/subresource/route"
+)
+
+const (
+	validCertificate = `
+-----BEGIN CERTIFICATE-----
```
stlaz, line 22: thanks, I completely forgot you should also be checking at least the expiry date. For example, this certificate appears to have expired 4 years ago.
```
            Not Before: Apr  5 15:15:55 2013 GMT
            Not After : Apr  4 15:15:55 2015 GM
```

pkg/console/controllers/route/controller.go
```diff
@@ -110,59 +124,205 @@ func (c *RouteSyncController) sync() error {
...
+	}
+	customTLS.Key = string(key)
+
+	return customTLS, nil
+}
+
+func certificateVerifier(customCert []byte) (*pem.Block, error) {
+	block, _ := pem.Decode([]byte(customCert))
+	if block == nil {
+		return nil, fmt.Errorf("failed to parse custom certificate")
```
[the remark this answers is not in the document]
stlaz: ah, nvm, `pem.Decode` does not return error, my bad

manifests/03-rbac-role-ns-openshift-config.yaml
```diff
@@ -8,6 +8,7 @@ rules:
       - ""
     resources:
       - configmaps
+      - secrets
```
benjaminapetersen, line 11: Too bad we have to add this, I liked that we didn't read secrets from here.

benjaminapetersen: Not that we ARE reading them all, just that now we can. but we are only doing a GET on the one that matters to us.

pkg/console/controllers/route/controller.go
benjaminapetersen: oh just meant looks good to me.
positive comment, no issues :)

manifests/03-rbac-role-ns-openshift-config.yaml
jhadvig: Other solution would be to have the secret name as API, then we could specify only that single one... but not sure if thats worth it.

pkg/console/controllers/route/controller.go
```diff
@@ -21,17 +25,17 @@ import (
 	// openshift
 	operatorsv1 "github.com/openshift/api/operator/v1"
 	routev1 "github.com/openshift/api/route/v1"
+	configclientv1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	operatorclientv1 "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1"
 	v1 "github.com/openshift/client-go/operator/informers/externalversions/operator/v1"
 	routeclientv1 "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
 	routesinformersv1 "github.com/openshift/client-go/route/informers/externalversions/route/v1"
+	customerrors "github.com/openshift/console-operator/pkg/console/errors"
```
stlaz, line 33: nit: should have stayed below

pkg/console/controllers/route/controller.go
```diff
@@ -110,59 +121,190 @@ func (c *RouteSyncController) sync() error {
...
+func ValidateCustomCertSecret(customCertSecret *corev1.Secret) (*routesub.CustomTLSCert, error) {
+	if customCertSecret == nil {
+		return nil, nil
+	}
+	if customCertSecret.Type != corev1.SecretTypeTLS {
+		return nil, fmt.Errorf("custom cert secret is not in %q type, instead uses %q type", corev1.SecretTypeTLS, customCertSecret.Type)
+	}
+	if len(customCertSecret.Data) == 0 {
+		return nil, fmt.Errorf("custom cert secret doesn't contain 'Data' field")
+	}
```
stlaz, line 245: you probably don't need to check this since you're checking the presence of the interesting keys later on

pkg/console/controllers/route/controller.go
```diff
@@ -110,59 +121,190 @@ func (c *RouteSyncController) sync() error {
...
+		return fmt.Errorf("failed to decode key PEM")
+	}
+	_, err := x509.ParsePKCS8PrivateKey(block.Bytes)
+	if err != nil {
+		_, err = x509.ParsePKCS1PrivateKey(block.Bytes)
+		if err != nil {
+			_, err = x509.ParseECPrivateKey(block.Bytes)
+			if err != nil {
+				return fmt.Errorf("block %s is not valid", block.Type)
+			}
```
stlaz, line 304: nit:
```golang
if _, err := x509.x509.ParsePKCS8PrivateKey(block.Bytes); err != nil {
    if _, err = x509.ParsePKCS1PrivateKey(block.Bytes); err != nil {
    ...
    }
}
```

pkg/console/controllers/route/controller.go
```diff
@@ -110,59 +121,190 @@ func (c *RouteSyncController) sync() error {
...
+	if block == nil {
+		return fmt.Errorf("failed to decode key PEM")
+	}
+	_, err := x509.ParsePKCS8PrivateKey(block.Bytes)
+	if err != nil {
+		_, err = x509.ParsePKCS1PrivateKey(block.Bytes)
+		if err != nil {
+			_, err = x509.ParseECPrivateKey(block.Bytes)
+			if err != nil {
+				return fmt.Errorf("block %s is not valid", block.Type)
```
stlaz, line 303: `is not a valid private key PEM`

jhadvig:
@stlaz comments addressed.. lets merge this baby 😄

stlaz:
/lgtm
:partying_face:

PR #412, opened 2020-04-09
Publish license

benjaminapetersen:
based on #386 

/hold 
as the license file doesn't currently exist.

### DIFF
--- a/manifests/07-downloads-deployment.yaml
+++ b/manifests/07-downloads-deployment.yaml
@@ -89,10 +89,14 @@ spec:
               os.mkdir(arch)
               for operating_system in ['linux', 'mac', 'windows']:
                   os.mkdir(os.path.join(arch, operating_system))
+                  target_path = os.path.join(arch, operating_system, 'LICENSE')
+                  os.symlink('/usr/share/openshift/LICENSE', target_path)
           for arch in ['arm64', 'ppc64le', 's390x']:
               os.mkdir(arch)
               for operating_system in ['linux']:
                   os.mkdir(os.path.join(arch, operating_system))
+                  target_path = os.path.join(arch, operating_system, 'LICENSE')
+                  os.symlink('/usr/share/openshift/LICENSE', target_path)
 
           for arch, operating_system, path in [
                   ('amd64', 'linux', '/usr/share/openshift/linux_amd64/oc'),
--- a/pkg/console/controllers/clidownloads/controller.go
+++ b/pkg/console/controllers/clidownloads/controller.go
@@ -188,6 +188,41 @@ The oc binary offers the same capabilities as the kubectl binary, but it is furt
 	}
 }
 
+func LicenseDownloads(host, cliDownloadsName string) *v1.ConsoleCLIDownload {
+	baseURL := fmt.Sprintf("%s", util.HTTPS(host))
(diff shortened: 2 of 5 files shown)


manifests/07-downloads-deployment.yaml
```diff
@@ -89,10 +89,14 @@ spec:
               os.mkdir(arch)
               for operating_system in ['linux', 'mac', 'windows']:
                   os.mkdir(os.path.join(arch, operating_system))
+                  target_path = os.path.join(arch, operating_system, 'LICENSE')
+                  os.symlink('/usr/share/openshift/LICENSE', target_path)
```
jhadvig, line 93: Do we need to have a separate `LICENSE` file for each os? Can't hurt but II guess that they will be all the same.

pkg/console/controllers/clidownloads/controller.go
```diff
@@ -188,6 +195,41 @@ The oc binary offers the same capabilities as the kubectl binary, but it is furt
...
 
+func LicenseDownloads(host, cliDownloadsName string) *v1.ConsoleCLIDownload {
+	baseURL := fmt.Sprintf("%s", util.HTTPS(host))
+	platforms := []struct {
+		label    string
+		key      string
+		archType string
+	}{
+		{"Linux for x86_64", "amd64/linux", "LICENSE"},
+		{"Linux for ARM 64", "arm64/linux", "LICENSE"},
```
jhadvig, line 206:
```suggestion
		{"Linux for ARM 64 (unsupported)", "arm64/linux", "LICENSE"},
```

to be consistent with the `oc` binaries ?

pkg/console/controllers/clidownloads/controller.go
```diff
@@ -188,6 +195,41 @@ The oc binary offers the same capabilities as the kubectl binary, but it is furt
...
+		{"Linux for IBM Power, little endian", "ppc64le/linux", "LICENSE"},
```
jhadvig, line 207:
```suggestion
		{"Linux for IBM Power, little endian (unsupported)", "ppc64le/linux", "LICENSE"},
```

test/e2e/downloads_test.go
```diff
@@ -48,3 +48,31 @@ func TestDownloadsEndpoint(t *testing.T) {
 		t.Logf("%s %s\n", link.Text, resp.Status)
 	}
 }
+
+func TestDonwloadsLicense(t *testing.T) {
```
jhadvig, line 52:
```suggestion
func TestDownloadsLicense(t *testing.T) {
```

jhadvig:
Also do we wanna close https://github.com/openshift/console-operator/pull/386/files now ?

benjaminapetersen:
Not sure, we have several issues that need to be addressed.

manifests/07-downloads-deployment.yaml
sosiouxme: It's the same source for all, so the same license. In fact you could just use a local copy of the Apache 2.0 license and call it a day :) I don't see the license changing ever.

jhadvig:
Closing in favour of https://github.com/openshift/console-operator/pull/439

spadgett:
Replaced by #439

PR #416, opened 2020-04-21
Bug 1829236: CustomURL e2e

jhadvig:
/assign @benjaminapetersen


jhadvig:
Will extend the test suite with additional test that will be also setting custom TLS as a followUp.

Test will be failing until https://github.com/openshift/console-operator/pull/418 is merged.

@benjaminapetersen PTAL

PR #418, opened 2020-04-22
Bug 1826628: Delete custom route after it is unset from console operator config

jhadvig:
/assign @benjaminapetersen


jhadvig:
@benjaminapetersen added an additional commit that is handling setting the CA for custom route healthCheck

PR #419, opened 2020-04-23
Bug 1828272: Update 'Learning Portal' console helpMenu url to track where request it is coming from

dtaylor113:
Jan Kleinert
I'd like to append:
`?ref=webconsole`
That way we can filter for this traffic without stomping on any referral info


benjaminapetersen:
Should this be added to the blog link as well?

```yaml
kind: ConsoleLink
metadata:
  name: openshift-blog
spec:
  href: 'https://blog.openshift.com'
```

PR #420, opened 2020-04-29
Bug 1826873: Default console route should redirect to correct public URL when route hostname has been customized

jhadvig:
/assign @benjaminapetersen

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -14,12 +14,13 @@ const (
 	OpenShiftConsoleOperator            = "console-operator"
 	OpenShiftConsoleConfigMapName       = "console-config"
 	OpenShiftConsolePublicConfigMapName = "console-public"
-	OpenshiftConsoleCustomRouteName     = "console-custom"
 	ServiceCAConfigMapName              = "service-ca"
 	DefaultIngressCertConfigMapName     = "default-ingress-cert"
 	OpenShiftConsoleDeploymentName      = OpenShiftConsoleName
 	OpenShiftConsoleServiceName         = OpenShiftConsoleName
+	OpenshiftConsoleRedirectServiceName = "console-redirect"
 	OpenShiftConsoleRouteName           = OpenShiftConsoleName
+	OpenshiftConsoleCustomRouteName     = "console-custom"
 	OpenShiftConsoleDownloadsRouteName  = "downloads"
 	OAuthClientName                     = OpenShiftConsoleName
 	OpenShiftConfigManagedNamespace     = "openshift-config-managed"
@@ -32,4 +33,12 @@ const (
 	TrustedCABundleMountFile            = "tls-ca-bundle.pem"
 	OCCLIDownloadsCustomResourceName    = "oc-cli-downloads"
 	ODOCLIDownloadsCustomResourceName   = "odo-cli-downloads"
+
+	ConsoleContainerPortName    = "https"
+	ConsoleContainerPort        = 443
+	ConsoleContainerTargetPort  = 8443
+	RedirectContainerPortName   = "custom-route-redirect"
+	RedirectContainerPort       = 8444
+	RedirectContainerTargetPort = RedirectContainerPort
+	ConsoleServingCertName      = "console-serving-cert"
 )
--- a/pkg/console/controllers/service/controller.go
+++ b/pkg/console/controllers/service/controller.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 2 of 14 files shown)


pkg/console/subresource/deployment/deployment.go
```diff
@@ -321,6 +338,44 @@ func consoleContainer(cr *operatorv1.Console, volConfigList []volumeConfig, prox
...
 }
 
+func redirectContainer(redirectURL string) corev1.Container {
+	command := []string{
+		"console",
+		"redirect",
+		fmt.Sprintf("--redirect-url %s", redirectURL),
+	}
+
+	return corev1.Container{
```
spadgett, line 348: I don't think we need a separate container or new command. Couldn't the operator or console backend just listen on a different port for redirects?

jhadvig: We have been talking with @benjaminapetersen regarding the approach and we agreed that it would be best to have all this custom logic been handled by operator, instead of putting it into the console server. Not really sure if the redirect logic/code belongs there. On the other hand operator should be the right place since it should handle all customisations.

spadgett: But why does it take a separate container and command? Why not just expose a port from the existing process?

spadgett: It feels a little odd to put the redirect in the operator imo. The operator should be responsible for deploying and maintaining the app. A redirect to the correct URL feels like a function of the app itself, which would be the console backend's responsibility.

jhadvig: But having a server that should get a hostname from which it should redirect to the custom hostname and then watching two different hostnames seems weird.

jhadvig:
In order for the Service to pick up the changes https://github.com/openshift/library-go/pull/788 needs to go in.

pkg/console/subresource/deployment/deployment.go
spadgett: Console backend only needs to know the its current public URL, which it already has. It doesn't care where it's redirecting from. That is managed by the operator when it creates the service and updates the default route.

pkg/console/controllers/service/controller.go
```diff
@@ -108,23 +110,55 @@ func (c *ServiceSyncController) sync() error {
...
-	_, _, svcErr := resourceapply.ApplyService(c.serviceClient, c.recorder, service.DefaultService(nil))
+
+	requiredSvc := service.DefaultService(updatedOperatorConfig)
+	_, _, svcErr := resourceapply.ApplyService(c.serviceClient, c.recorder, requiredSvc)
 	status.HandleProgressingOrDegraded(updatedOperatorConfig, "ServiceSync", "FailedApply", svcErr)
+	if svcErr != nil {
+		return svcErr
+	}
+
+	redirectSvcErrReason, redirectSvcErr := c.SyncRedirectService(updatedOperatorConfig)
```
jhadvig, line 130: Have to create a new service for redirecting since the `Protocol` field on the Service is feature gated for 1.18. For that reason we cant really on updating it and we need to create a new service for redirecting

pkg/api/api.go
```diff
@@ -32,4 +33,12 @@ const (
 	TrustedCABundleMountFile            = "tls-ca-bundle.pem"
 	OCCLIDownloadsCustomResourceName    = "oc-cli-downloads"
 	ODOCLIDownloadsCustomResourceName   = "odo-cli-downloads"
+
+	ConsoleContainerPortName    = "https"
+	ConsoleContainerPort        = 443
+	ConsoleContainerTargetPort  = 8443
+	RedirectContainerPortName   = "redirect"
```
benjaminapetersen, line 40: Shall we call it `custom-route-redirect`, for clarity?

pkg/console/controllers/service/controller.go
```diff
@@ -108,23 +110,55 @@ func (c *ServiceSyncController) sync() error {
...
 
 	return svcErr
 }
 
-func (c *ServiceSyncController) removeService() error {
-	klog.V(2).Info("deleting console service")
-	defer klog.V(2).Info("finished deleting console service")
-	return c.serviceClient.Services(c.targetNamespace).Delete(c.ctx, service.Stub().Name, metav1.DeleteOptions{})
+func (c *ServiceSyncController) SyncRedirectService(operatorConfcig *operatorsv1.Console) (string, error) {
+	if !routesub.IsCustomRouteSet(operatorConfcig) {
```
benjaminapetersen, line 142: Since these will always appear as:

```
RedirectServiceSync: FailedDeleteRedirectService
RedirectServiceSync: FailedRedirectServiceApply
```
You can probably simplify them to:

```
RedirectServiceSync: FailedDelete
RedirectServiceSync: FailedApply
```
I believe that is the pattern we have elsewhere, such as `OAuthClientSecretSync", "FailedApply"` and `"ServiceSync", "FailedApply"`

pkg/console/subresource/route/route.go
```diff
@@ -76,12 +76,24 @@ func GetOrCreate(ctx context.Context, client routeclient.RoutesGetter, required
 	return route, isNew, nil
 }
 
+// Default `console` route poinst by default to the `console` service.
```
benjaminapetersen, line 79: Nit: `poinst` -> `points`.

pkg/api/api.go
jhadvig: Wanted to keep the name simple here

PR #421, opened 2020-05-03
Bug 1827050: Update to use revised ApplyDeployment

marun:
The new method hashes the spec of the deployment which removes the need for forcing rollout.


marun:
/hold

There is a bit more cleanup possible, updating shortly.

marun:
/hold cancel

Updated deployment composition stop adding annotations to the pod template (never necessary) and removed now incorrect comment indicating that ApplyDeployment did not consider the pod template in determining rollout.

spadgett:
> Out of curiosity, why are the deployment's annotations mirrored on the pod template?

To force a rollout. Changing the deployment annotations doesn't force a rollout, but changing the pod template annotations does.

marun:
> > Out of curiosity, why are the deployment's annotations mirrored on the pod template?
> 
> To force a rollout. Changing the deployment annotations doesn't force a rollout, but changing the pod template annotations does.

Whoops, I should have asked the reverse. Why mirror the pod's annotations on the deployment?

spadgett:
> Whoops, I should have asked the reverse. Why mirror the pod's annotations on the deployment?

I suspect at the time the comment was written it was accurate, but that's since been fixed in library-go.

marun:
/hold 

The library-go bump is resulting in test failure due to a metrics registration problem.

PR #422, opened 2020-05-03
WIP Bug 1834852: Add make target to run operator with telepresence

marun:
Once telepresence has been installed (see the run-telepresence.sh script for instructions), `make telepresence` will replace the operator's pod with a local process that has access to the same environment as the pod. This can speed up development since iteration doesn't rely on image build/push/pull. See https://telepresence.io for more details.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ include $(addprefix ./vendor/github.com/openshift/build-machinery-go/make/, \
     targets/openshift/deps-gomod.mk \
     targets/openshift/images.mk \
     targets/openshift/bindata.mk \
+    targets/openshift/operator/telepresence.mk \
 )
 
 # Run core verification and all self contained tests.
@@ -47,3 +48,8 @@ test-e2e:
 clean:
 	rm -rf $(OUT_DIR)
 .PHONY: clean
+
+# Configure the 'telepresence' target
+# See vendor/github.com/openshift/build-machinery-go/scripts/run-telepresence.sh for usage and configuration details
+export TP_DEPLOYMENT_YAML ?=./manifests/07-operator.yaml
+export TP_CMD_PATH ?=./cmd/console
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .DS_Store 
 .idea
 ./deploy
+telepresence.log
 
 # tools output reports into _output
 _output
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ The console-operator installs and maintains the web console on a cluster.
 
 ## Run on a 4.0.0 Cluster
 
-The console operator is installed by default and will automatically maintain a console.  
+The console operator is installed by default and will automatically maintain a console.
 
 ## Development Setup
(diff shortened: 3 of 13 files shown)


Makefile
```diff
@@ -7,6 +7,7 @@ include $(addprefix ./vendor/github.com/openshift/build-machinery-go/make/, \
     targets/openshift/deps-gomod.mk \
     targets/openshift/images.mk \
     targets/openshift/bindata.mk \
+	targets/openshift/operator/telepresence.mk \
```
benjaminapetersen, line 10: nit: spacing?

benjaminapetersen:
Care to add a README entry?

Makefile
marun: Done

marun:
> Care to add a README entry?

Done.

marun:
/retest

I don't think the test failures could be related to the minor dep bump or addition of a new make target.

marun:
@benjaminapetersen The flake was persistent, so I've proposed a fix.

test/e2e/framework/framework.go
```diff
@@ -224,59 +222,42 @@ func IsResourceAvailable(errChan chan error, client *ClientSet, resource Testing
...
 		}
 		if !errors.IsNotFound(err) {
-			return true, err
-		}
-		counter++
-		if counter == maxCount {
+			// Unable to retrieve resource due to error. Reset the counter and keep waiting.
+			counter = 0
+			t.Errorf("error checking that console %s %s still exists: %v", resource.kind, resource.name, err)
 			return true, nil
```
p0lyn0mial, line 254: I think that this will immediately yield - `PollImmediate tries a condition func until it returns true, an error, or the timeout is reached`

test/e2e/framework/framework.go
```diff
@@ -224,59 +222,42 @@ func IsResourceAvailable(errChan chan error, client *ClientSet, resource Testing
...
-	return failed
-}
-
-func LoopResources(client *ClientSet, inner func(errChan chan error, client *ClientSet, resource TestingResource)) error {
+func ConsoleResourcesUnavailable(t *testing.T, client *ClientSet) error {
 	resources := getTestingResources()
 
 	errChan := make(chan error)
 	for _, resource := range resources {
-		go inner(errChan, client, resource)
```
p0lyn0mial, line 248: where did `inner` function go?
is it still being used somewhere?

test/e2e/framework/framework.go
marun: Done

test/e2e/framework/framework.go
marun: Inner function was previously passed in as an argument, but since had only a single caller passing `IsResourceUnavailable` I decided to call it directly.

marun:
Marking WIP for now, as per the FIXME in the test fix commit there is further work to be done in the error handling around verification of removal.

marun:
@spadgett Ok! Will revisit this week.

PR #424, opened 2020-05-06
Bug 1830809: Use helperv1.UpdateStatus() for updating operators status

jhadvig:
/assign @benjaminapetersen


pkg/console/controllers/route/controller.go
```diff
@@ -130,17 +134,37 @@ func (c *RouteSyncController) sync() error {
 	}
 
 	updatedOperatorConfig := operatorConfig.DeepCopy()
+	updateConditionFuncs := []v1helpers.UpdateStatusFunc{}
 
 	defaultRouteErrReason, defaultRouteErr := c.SyncDefaultRoute(updatedOperatorConfig)
-	status.HandleProgressingOrDegraded(updatedOperatorConfig, "DefaultRouteSync", defaultRouteErrReason, defaultRouteErr)
+	defaultRouteSyncConditions := status.HandleProgressingOrDegraded("DefaultRouteSync", defaultRouteErrReason, defaultRouteErr)
+	updateConditionFuncs = append(updateConditionFuncs, defaultRouteSyncConditions...)
```
benjaminapetersen, line 141: Definitely want to eliminate the need for the added conditional logic everywhere.  I think we can tuck most of this away in our own status helpers, and perhaps just wrap the `v1helpers.UpdateStatus` call with our own (perhaps existing) helpers and not change our controllers much.

pkg/console/status/status.go
```diff
@@ -102,15 +104,15 @@ func setConditionValue(conditionType string, err error) operatorsv1.ConditionSta
...
-func IsDegraded(operatorConfig *operatorsv1.Authentication) bool {
-	for _, condition := range operatorConfig.Status.Conditions {
-		if strings.HasSuffix(condition.Type, operatorsv1.OperatorStatusTypeDegraded) &&
-			condition.Status == operatorsv1.ConditionTrue {
-			return true
-		}
-	}
-	return false
-}
+// func IsDegraded(operatorConfig *operatorsv1.Authentication) bool {
```
benjaminapetersen, line 107: Its interesting this was so easily removed.

pkg/console/controllers/route/controller.go
```diff
@@ -126,21 +132,51 @@ func (c *RouteSyncController) sync() error {
 		}
 		return c.removeRoute(api.OpenShiftConsoleName)
 	default:
-		return fmt.Errorf("unknown state: %v", operatorConfig.Spec.ManagementState)
+		return fmt.Errorf("unknown state: %v", updatedOperatorConfig.Spec.ManagementState)
 	}
 
-	updatedOperatorConfig := operatorConfig.DeepCopy()
+	conditionFuncs := []v1helpers.UpdateStatusFunc{}
```
benjaminapetersen, line 138: So can we do something like this in `status.go`:

```golang
type conditionHandler struct {
	client   v1helpers.OperatorClient
	updaters []v1helpers.UpdateStatusFunc
}

func (c conditionHandler) Add(updater v1helpers.UpdateStatusFunc) {
	c.updaters = append(c.updaters, updater)
}

func (c conditionHandler) Flush() error {
        // though, we really need to call all the funcs, catch the errors, then clear 
       // the updaters to an empty slice before next use: 
       // c.updaters = []v1helpers.UpdateStatusFunc
	if _, _, updateErr := v1helpers.UpdateStatus(c.client, c.updaters...); updateErr != nil {
		return updateErr
	}
	return nil 
}

func NewConditionHandler(client v1helpers.OperatorClient) conditionHandler {
	return conditionHandler{
		client: client,
	}
}
```
and then here do something like:

```golang
        // we could call this in `.sync()`, which would let us just throw it away at the end of 
        // each sync loop and not worry about clearing out the slice of []v1helpers.UpdateStatusFunc.
       // alternatively we create these and attach them to our controllers when we make RouteSyncController{} 
       // in NewRouteSyncController() func (and with all of our other controllers)...  but that means we defintely
       // have to have the clearing logic work to ensure a func doesn't linger across the .sync() loop calls. 
	conditionsHandler := status.NewConditionHandler(c.operatorClient)

	defaultRoute, defaultRouteErrReason, defaultRouteErr := c.SyncDefaultRoute(updatedOperatorConfig)
	defaultRouteSyncConditionsFunc := status.HandleProgressingOrDegraded("DefaultRouteSync", defaultRouteErrReason, defaultRouteErr)
	conditionsHandler.Add(defaultRouteSyncConditionsFunc)
	if defaultRouteErr != nil {
                // i guess this can return an error, also, and we need to deal.  ideally 
                // we clobber the nested if statements
		conditionsHandler.Flush() 
		return defaultRouteErr
	}
```

benjaminapetersen:
Curious this is still the `ci/prow/e2e-aws-console` `aws` instead of the `gcp` job.  Maybe that was intentional, @spadgett do you recall? I do not.

benjaminapetersen:
```
Running `gofmt -s -l` on 71 file(s).
go vet  ./...
# github.com/openshift/console-operator/pkg/console/status
vet: pkg/console/status/status_test.go:217:32: cannot use tt.input.operatorConfig (variable of type *v1.Console) as string value in argument to HandleProgressingOrDegraded
make: *** [verify-govet] Error 2
```

benjaminapetersen:
```
 ?   	github.com/openshift/console-operator/pkg/console/starter	[no test files]
FAIL	github.com/openshift/console-operator/pkg/console/status [build failed]
ok  	github.com/openshift/console-operator/pkg/console/subresource/configmap	1.815s 
```
Nothing too crazy or unexpected.

pkg/console/operator/sync_v400.go
```diff
@@ -159,17 +163,17 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console,
 	if consoleURL == "" {
 		err := customerrors.NewSyncError("waiting on route host")
 		klog.Errorf("%q: %v", "route", err)
-		return err
+		conditionHandler.FlushAndReturn(err)
```
marun, line 166: Is it your intent to ignore the error returned by FlushAndReturn?

benjaminapetersen:
rebase needed.

pkg/console/operator/sync_v400.go
jhadvig: err...

pkg/console/status/status.go
```diff
@@ -102,26 +102,66 @@ func setConditionValue(conditionType string, err error) operatorsv1.ConditionSta
...
+	generationFunc := func(oldStatus *operatorsv1.OperatorStatus) error {
+		oldStatus.ReadyReplicas = newReadyReplicas
+		return nil
+	}
+	c.statusFuncs = append(c.statusFuncs, generationFunc)
+}
+
+func (c *StatusHandler) UpdateDeploymentsLastGeneration(newLastGeneration int64) {
+	generationFunc := func(oldStatus *operatorsv1.OperatorStatus) error {
+		generationFound := false
```
marun, line 145: Please use library-go's `SetDeploymentGeneration` instead of reinventing it:

https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourcemerge/apps.go#L48

Here's an example of use in the openshift-apiserver operator:

https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master/pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go#L175

This also relates to something I mentioned on slack to you - the the provision of the generation argument to `ApplyDeployment` should be updated to use library-go's `ExpectedDeploymentGeneration`:

https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourcemerge/apps.go#L40

Example of use in openshift-apiserver operator: 

https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master/pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go#L367

vs what the console operator currently does (retrieves the generation from the target deployment rather than the operator status):

https://github.com/openshift/console-operator/blob/master/pkg/console/operator/sync_v400.go#L225

pkg/console/operator/sync_v400.go
```diff
@@ -221,8 +226,8 @@ func (co *consoleOperator) SyncDeployment(
 	proxyConfig *configv1.Proxy,
 	canMountCustomLogo bool) (consoleDeployment *appsv1.Deployment, changed bool, reason string, err error) {
 
+	updatedOperatorConfig := operatorConfig.DeepCopy()
 	requiredDeployment := deploymentsub.DefaultDeployment(operatorConfig, cm, serviceCAConfigMap, defaultIngressCertConfigMap, trustedCAConfigMap, sec, proxyConfig, canMountCustomLogo)
-	expectedGeneration := getDeploymentGeneration(co)
```
marun, line 225: `getDeploymentGeneration` is now unused, but I'd prefer to get this PR merged in current form and leave that cleanup to future efforts.

benjaminapetersen: Yup, we have sevearl cleanup items we will defer to get you unblocked.

PR #427, opened 2020-05-13
Bug 1835221: Sync loop should error out on resources that depend on route that is not admitted

jhadvig:
This basically refacts a bit the `GetCanonicalHost` function and adds unit test for it.

/assign @benjaminapetersen


jhadvig:
```
   2) Login test
A Jasmine spec timed out. Resetting the WebDriver Control Flow.
A Jasmine spec timed out. Resetting the WebDriver Control Flow.
      ✖ logs in via htpasswd identity provider (2 failures)
      • does not show admin nav items in Administration to htpasswd user
      • does not show admin nav items in Operators to htpasswd user
      • does not show admin nav items in Storage to htpasswd user
```

spadgett:
@jhadvig I think the test failure is legitimate.

PR #429, opened 2020-05-14
Bug 1835551: Inject proxy envars to console-operator deployment

jhadvig:
not sure if we should not backport to 4.4 ?

/assign @benjaminapetersen


benjaminapetersen:
Pondering if we should look up the proxy test suite & add it?

PR #430, opened 2020-05-18
[release-4.3] Bug 1836939: Bump godeps to get fix from library-go#799

jhadvig:
/assign @benjaminapetersen


benjaminapetersen:
/approve
/lgtm

So long as tests pass.  looks like we should be in good shape.

PR #432, opened 2020-05-21
Bug 1839621: Grant Console SA permissions to read webhooks configurations

sleshchenko:
This PR grants Console SA permissions to read webhooks configurations to check if the terminal operator is correctly installed.

It's needed for https://github.com/openshift/console/pull/5332

Now we try to figure out what is the right way to detect operator installed, like check deployment in openshift-operators, or ClusterServiceVersion, ... So, probably permissions list here will be extended

### DIFF
--- a/manifests/03-rbac-role-cluster.yaml
+++ b/manifests/03-rbac-role-cluster.yaml
@@ -89,4 +89,3 @@ rules:
   - validatingwebhookconfigurations
   verbs:
   - get
-  - watch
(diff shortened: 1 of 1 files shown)


manifests/03-rbac-role-cluster.yaml
```diff
@@ -82,3 +82,11 @@ rules:
...
   - list
   - watch
+- apiGroups:
+  - admissionregistration.k8s.io
+  resources:
+  - mutatingwebhookconfigurations
+  - validatingwebhookconfigurations
+  verbs:
+  - get
+  - watch
```
spadgett, line 92:
```suggestion
  - get
```

PR #438, opened 2020-06-15
Bug 1846922: "downloads" pod does not work on the node which is disabled IPv6

bysnupy:
* Version: OCP4.4+
* Description:
  When IPv6 is disabled, the socket binding will be failed, so we catch the exception and then IPv4 would use at that time.
* Reference: https://bugzilla.redhat.com/show_bug.cgi?id=1846922

### DIFF
--- a/manifests/07-downloads-deployment.yaml
+++ b/manifests/07-downloads-deployment.yaml
@@ -59,7 +59,7 @@ spec:
         - '-c'
         - |
           cat <<EOF >>/tmp/serve.py
-          import BaseHTTPServer, os, re, signal, SimpleHTTPServer, socket, sys, tarfile, tempfile, threading, time, zipfile
+          import BaseHTTPServer, errno, os, re, signal, SimpleHTTPServer, socket, sys, tarfile, tempfile, threading, time, zipfile
 
           signal.signal(signal.SIGTERM, lambda signum, frame: sys.exit(0))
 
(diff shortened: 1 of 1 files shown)


spadgett:
@wking Would you mind reviewing since you know the code the best?

/assign @wking

manifests/07-downloads-deployment.yaml
```diff
@@ -116,8 +116,15 @@ spec:
...
-          sock = socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
+          try:
+              addr = ('::', 8080)
+              sock = socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
+          except socket.error as err:
+              # errno.EAFNOSUPPORT is "socket.error: [Errno 97] Address family not supported by protocol"
+              # When IPv6 is disabled, socket will bind using IPv4.
+              if err.errno == errno.EAFNOSUPPORT:
+                  addr = ('', 8080)
+                  sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
wking, line 127: I think you need an:

```python
else:
    raise
```

here to handle the "not `EAFNOSUPPORT`" case.

manifests/07-downloads-deployment.yaml
```diff
@@ -59,7 +59,7 @@ spec:
         - '-c'
         - |
           cat <<EOF >>/tmp/serve.py
-          import BaseHTTPServer, os, re, signal, SimpleHTTPServer, socket, sys, tarfile, tempfile, threading, time, zipfile
+          import BaseHTTPServer, os, re, signal, SimpleHTTPServer, socket, sys, tarfile, tempfile, threading, time, zipfile, errno
```
wking, line 62: Nit: these imports are currently alphabetized.  Can you slot `errno` in before `os`?

manifests/07-downloads-deployment.yaml
bysnupy: @wking Thank you for your review, I modified as you mentioned.

manifests/07-downloads-deployment.yaml
bysnupy: I also modified it.

wking:
/lgtm

Not sure if we have CI coverage or not...

bysnupy:
@wking Could you mark LGTM again ? the label is removed.

spadgett:
> Not sure if we have CI coverage or not...

We do have some test coverage:

https://github.com/openshift/console-operator/blob/master/test/e2e/downloads_test.go

PR #445, opened 2020-07-02
Bug 1853270: Update console config on change

jhadvig:
We should be updating the console config status only when the consoleURL change

/assign @spadgett

### DIFF
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -199,13 +199,14 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console, 
 
 func (co *consoleOperator) SyncConsoleConfig(consoleConfig *configv1.Console, consoleURL string) (*configv1.Console, error) {
 	oldURL := consoleConfig.Status.ConsoleURL
-	updated := consoleConfig.DeepCopy()
-	if updated.Status.ConsoleURL != consoleURL {
-		klog.V(4).Infof("updating console.config.openshift.io with url: %v", consoleURL)
-		updated.Status.ConsoleURL = consoleURL
-	}
 	metrics.HandleConsoleURL(oldURL, consoleURL)
-	return co.consoleConfigClient.UpdateStatus(co.ctx, updated, metav1.UpdateOptions{})
+	if oldURL != consoleURL {
+		klog.V(4).Infof("updating console.config.openshift.io with url: %v", consoleURL)
+		updated := consoleConfig.DeepCopy()
+		updated.Status.ConsoleURL = consoleURL
+		return co.consoleConfigClient.UpdateStatus(co.ctx, updated, metav1.UpdateOptions{})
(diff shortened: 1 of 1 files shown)


pkg/console/operator/sync_v400.go
```diff
@@ -200,12 +200,13 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console,
 func (co *consoleOperator) SyncConsoleConfig(consoleConfig *configv1.Console, consoleURL string) (*configv1.Console, error) {
 	oldURL := consoleConfig.Status.ConsoleURL
 	updated := consoleConfig.DeepCopy()
```
spadgett, line 202: We could move this inside the `if` since we only need to copy if we're doing an update

pkg/console/operator/sync_v400.go
```diff
@@ -200,12 +200,13 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console,
...
+	metrics.HandleConsoleURL(oldURL, consoleURL)
 	if updated.Status.ConsoleURL != consoleURL {
 		klog.V(4).Infof("updating console.config.openshift.io with url: %v", consoleURL)
 		updated.Status.ConsoleURL = consoleURL
+		return co.consoleConfigClient.UpdateStatus(co.ctx, updated, metav1.UpdateOptions{})
```
spadgett, line 207:
```suggestion
	metrics.HandleConsoleURL(oldURL, consoleURL)
	if oldURL != consoleURL {
		klog.V(4).Infof("updating console.config.openshift.io with url: %v", consoleURL)
	        updated := consoleConfig.DeepCopy()
		updated.Status.ConsoleURL = consoleURL
		return co.consoleConfigClient.UpdateStatus(co.ctx, updated, metav1.UpdateOptions{})
```

pkg/console/operator/sync_v400.go
```diff
@@ -200,12 +200,13 @@ func (co *consoleOperator) sync_v400(updatedOperatorConfig *operatorv1.Console,
 func (co *consoleOperator) SyncConsoleConfig(consoleConfig *configv1.Console, consoleURL string) (*configv1.Console, error) {
 	oldURL := consoleConfig.Status.ConsoleURL
 	updated := consoleConfig.DeepCopy()
+	metrics.HandleConsoleURL(oldURL, consoleURL)
```
spadgett, line 203: Do we need to cal this when the URL hasn't changed?

spadgett:
/hold
since i asked some questions after tagging :)

pkg/console/operator/sync_v400.go
jhadvig: > Do we need to cal this when the URL hasn't changed?

`HandleConsoleURL` is handling all the cases, even if the URL has not changed, then it will just return.

jhadvig:

PR #454, opened 2020-08-10
Bug 1856667: Bump DEFAULT_DOC_URL for okd & ocp

jhadvig:
/assign @spadgett

### DIFF
--- a/pkg/console/subresource/configmap/brand_okd.go
+++ b/pkg/console/subresource/configmap/brand_okd.go
@@ -4,5 +4,5 @@ package configmap
 
 const (
 	DEFAULT_BRAND   = "okd"
-	DEFAULT_DOC_URL = "https://docs.okd.io/4.5/"
+	DEFAULT_DOC_URL = "https://docs.okd.io/latest/"
 )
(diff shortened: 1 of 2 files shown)


pkg/console/subresource/configmap/brand_okd.go
```diff
@@ -4,5 +4,5 @@ package configmap
 
 const (
 	DEFAULT_BRAND   = "okd"
-	DEFAULT_DOC_URL = "https://docs.okd.io/4.5/"
+	DEFAULT_DOC_URL = "https://docs.okd.io/4.6/"
```
spadgett, line 7: Not new, but this should be https://docs.okd.io/latest/

jhadvig: @spadgett but then the 4.6 release will always point to the latest, instead to the particular version. So in case we will be working on 4.7 then the 4.6 will be pointing to latest.

spadgett: OKD *only* has a `latest` doc URL for 4.x. We can check with the doc team if that is planned to change.

cc @ahardin-rh

ahardin-rh: @spadgett Correct. I don't think there is any plan to change that. 
cc @vikram-redhat

jhadvig:
@ahardin-rh thanks for clearing this out.

PR #455, opened 2020-08-11
Allow console service account to read HelmChartRepository CR

pedjak:
This change allows non-admin users logged through UI to browse helm charts
available in custom configured Helm chart repositories.

A separate PR will be submited to openshift/console making /api/helm/charts/index.yaml
to perform k8s api requests on behalf of the service account.

### DIFF
--- a/manifests/03-rbac-role-ns-openshift-config.yaml
+++ b/manifests/03-rbac-role-ns-openshift-config.yaml
@@ -13,3 +13,17 @@ rules:
       - get
       - list
       - watch
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: console
+  namespace: openshift-config
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - configmaps
+      - secrets
+    verbs:
+      - get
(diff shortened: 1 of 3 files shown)


manifests/03-rbac-role-ns-openshift-config.yaml
```diff
@@ -13,3 +13,17 @@ rules:
       - get
       - list
       - watch
+---
+kind: Role
```
jhadvig, line 17: Why is this Role and the RoleBinding needed ?

pedjak: HelmChartRepository instances could refer certificates stored in a configmap and/or secret in `openshift-config` namespace. See [the api](https://github.com/openshift/api/blob/master/helm/v1beta1/types_helm.go#L67-L76). Therefore, we need to allow service account to read them.

spadgett:
We should not merge the RBAC changes until the other PR is ready so we can review them together. It's difficult to review this in isolation.

This feels like the wrong approach to me. The console operator should watch these resources and pass it as config to the console backend. It's how we handle other similar config like proxy.

spadgett:
@pedjak Can you link the bug that is tracking this change?

pedjak:
Bug is Jira: https://issues.redhat.com/browse/APPSVC-655

Counterpart console PR: https://github.com/openshift/console/pull/6318

PR #456, opened 2020-08-12
Revert "Allow console service account to read HelmChartRepository CR"

spadgett:
Reverts openshift/console-operator#455

Reverting this PR based on https://github.com/openshift/console-operator/pull/455#issuecomment-672886657. The approach does not look right to me. We can reapply if it's really needed, but we need to see the corresponding console PR to properly review. It would be better not to grant the console service account this authority if we can avoid it, and the approach is inconsistent with how the console handles other cluster config.

/assign @jhadvig 
cc @pedjak


pedjak:
Bug is Jira: https://issues.redhat.com/browse/APPSVC-655

Counterpart console PR: openshift/console#6318

PR #458, opened 2020-08-17
Bug 1872375: Add RedHat Helm chart repository as default repo in OCP payload

pedjak:
Although console endpoint `/api/helm/charts/index.yaml` handles the situation
when there is no HelmChartRepository CR present in the cluster, we should align
us to other default cluster settings and provide the default HelmChartRepository CR in the payload

Prior introducing openshift/console#5933 all authenticated users could browse the charts from the chart repo.
This PR restores that functionality by introducing additional `helm-chartrepos-viewer` ClusterRole,
binding it to all authenticated users.

### DIFF
--- a/manifests/03-rbac-role-cluster.yaml
+++ b/manifests/03-rbac-role-cluster.yaml
@@ -97,3 +97,16 @@ rules:
   - validatingwebhookconfigurations
   verbs:
   - get
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: helm-chartrepos-viewer
+rules:
+  - apiGroups:
+      - helm.openshift.io
+    resources:
+      - helmchartrepositories
+    verbs:
+      - get
+      - list
--- a/manifests/04-rbac-rolebinding-cluster.yaml
+++ b/manifests/04-rbac-rolebinding-cluster.yaml
@@ -54,3 +54,18 @@ subjects:
   - kind: ServiceAccount
     name: console
     namespace: openshift-console
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  annotations:
+    "release.openshift.io/create-only": 'true'
+  name: helm-chartrepos-view
+subjects:
+  - kind: Group
+    apiGroup: rbac.authorization.k8s.io
+    name: 'system:authenticated'
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: helm-chartrepos-viewer
--- a/manifests/01-helm.yaml
+++ b/manifests/01-helm.yaml
@@ -1,0 +1,10 @@
+apiVersion: helm.openshift.io/v1beta1
+kind: HelmChartRepository
+metadata:
+  annotations:
+    "release.openshift.io/create-only": 'true'
+  name: redhat-helm-repo
+spec:
+  name: Red Hat Helm Charts
+  connectionConfig:
+    url: https://redhat-developer.github.io/redhat-helm-charts


manifests/03-rbac-role-cluster.yaml
```diff
@@ -97,3 +97,16 @@ rules:
...
+metadata:
+  name: helm-chartrepos-viewer
+rules:
+  - apiGroups:
+      - helm.openshift.io
+    resources:
+      - helmchartrepositories
+    verbs:
+      - get
+      - list
```
sbose78, line 112: Looks good to me, all authenticated users should be allowed to 'read' helm chart repositories. That'll naturally drive the content in the developer catalog.

Any advanced configuration needs to be done by the admin.

manifests/05-helm.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: helm.openshift.io/v1beta1
+kind: HelmChartRepository
+metadata:
+  name: redhat-helm-repo
+spec:
+  name: Red Hat Helm Charts
+  connectionConfig:
+    url: https://redhat-developer.github.io/redhat-helm-charts
```
sbose78, line 8: So, 
1. All authenticated users get to see helm chart repos
2. Ootb, they see the default one. https://redhat-developer.github.io/redhat-helm-charts was previously hard-coded in the console config - instead of giving it any special treatment, it is being made just another CR. 👍

sbose78:
@spadgett @christianvogt @rohitkrai03 
FYI
This effectively conforms to _console is pretty kubectl_. User doesn't get any more privileges by using console. Visibility of every resource incl. the default one is driven by the user's privileges.

manifests/05-helm.yaml
jhadvig: > Ootb, they see the default one. https://redhat-developer.github.io/redhat-helm-charts was previously hard-coded in the console config

you mean in the `console-config` CM in `openshift-console` namespace ?

manifests/03-rbac-role-cluster.yaml
jhadvig: Won't `watch` action be needed?

jhadvig (changes requested):
The `05-helm.yaml`should be `01-helm.yaml` no honor the ordering we currently do.

Will the `HelmChartRepository` CRD be available on the cluster by default ?

manifests/03-rbac-role-cluster.yaml
pedjak: no, because the operator does not watch for changes. The permissions are needed for console `/api/helm/charts/index.yaml` endpoiint - it retrieves the list of all `HelmChartRepository` CRs from the cluster, and use them to retrieve individual `index.yaml` files that get combined at the end.

manifests/05-helm.yaml
pedjak: HelmChartRepository is cluster scoped.

pedjak:
> Will the HelmChartRepository CRD be available on the cluster by default ?

@jhadvig  - yes - it is already deployed as [the part console-operator payload](https://github.com/openshift/console-operator/blob/master/Dockerfile.rhel7#L18)

manifests/04-rbac-rolebinding-cluster.yaml
```diff
@@ -54,3 +54,16 @@ subjects:
...
     namespace: openshift-console
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: helm-chartrepos-view
+subjects:
+  - kind: Group
+    apiGroup: rbac.authorization.k8s.io
+    name: 'system:authenticated'
```
spadgett, line 65: I believe you're need a PR like https://github.com/openshift/origin/pull/23975 to get this to pass CI.

jhadvig:
https://github.com/openshift/console-operator/pull/458#discussion_r476505062 will need to be delivered before merging.

manifests/04-rbac-rolebinding-cluster.yaml
pedjak: Pushed in https://github.com/openshift/origin/pull/25485

jhadvig (approved):
/lgtm
/approve

PR #459, opened 2020-08-24
Bug 1870514: Configuring custom certificate for default console route

jhadvig:
When cluster user wants to define custom TLS secret for the default route he should just define the secret's name in the console-operator config. The issue is that we cannot have two routes(default, custom) with the same hostname.
I thing we should update the API [comment](https://github.com/openshift/api/blob/master/operator/v1/types_console.go#L51-L59) with this use-case.

Also we should probably backport this fix to 4.5

/assign @spadgett

### DIFF
--- a/pkg/console/controllers/route/controller.go
+++ b/pkg/console/controllers/route/controller.go
@@ -174,7 +174,16 @@ func (c *RouteSyncController) removeRoute(routeName string) error {
 }
 
 func (c *RouteSyncController) SyncDefaultRoute(operatorConfig *operatorsv1.Console) (*routev1.Route, string, error) {
-	requiredDefaultRoute := routesub.DefaultRoute(operatorConfig)
+	customTLSSecret, configErr := c.GetDefaultRouteTLSSecret(operatorConfig)
+	if configErr != nil {
+		return nil, "InvalidDefaultRouteConfig", configErr
+	}
+	customTLSCert, secretValidationErr := ValidateCustomCertSecret(customTLSSecret)
+	if secretValidationErr != nil {
+		return nil, "InvalidCustomTLSSecret", secretValidationErr
+	}
+
+	requiredDefaultRoute := routesub.DefaultRoute(operatorConfig, customTLSCert)
 
 	defaultRoute, _, defaultRouteError := routesub.ApplyRoute(c.routeClient, c.recorder, requiredDefaultRoute)
 	if defaultRouteError != nil {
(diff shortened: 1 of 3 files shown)


pkg/console/controllers/route/controller.go
```diff
@@ -174,7 +174,16 @@ func (c *RouteSyncController) removeRoute(routeName string) error {
...
+	customTLSSecret, configErr := c.ValidateDefaultRouteConfig(operatorConfig)
+	if configErr != nil {
+		return nil, "InvalidDefaultRouteConfig", configErr
+	}
+	customTLSCert, secretValidationErr := ValidateCustomCertSecret(customTLSSecret)
+	if secretValidationErr != nil {
+		return nil, "InvalidCustomTLSSecret", secretValidationErr
+	}
+
+	requiredDefaultRoute := routesub.DefaultRoute(operatorConfig, customTLSCert)
```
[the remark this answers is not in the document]
jhadvig: > Is this unconditionally applying the custom certificate to the default route?

No the custom cert will be applied to the default route only if the `spec.route.secret.name` is set (`spec.route.hostname` shall not be set in this case) and pointing to a secret in `openshift-config` namespace.

When cluster-admin sets the `spec.route.hostname` with a duplicate of the default hostname, nothing should happen, since a custom route with dup. hostname wont be admitted.

pkg/console/controllers/route/controller.go
```diff
@@ -224,6 +233,24 @@ func (c *RouteSyncController) SyncCustomRoute(operatorConfig *operatorsv1.Consol
 	return customRoute, "", customRouteError
 }
 
+func (c *RouteSyncController) ValidateDefaultRouteConfig(operatorConfig *operatorsv1.Console) (*corev1.Secret, error) {
```
spadgett, line 236: Maybe `GetDefaultRouteTLSConfig`. This isn't really validating anything just returning config, specifically TLS config.

pkg/console/controllers/route/controller.go
```diff
@@ -174,7 +174,16 @@ func (c *RouteSyncController) removeRoute(routeName string) error {
 }
 
 func (c *RouteSyncController) SyncDefaultRoute(operatorConfig *operatorsv1.Console) (*routev1.Route, string, error) {
-	requiredDefaultRoute := routesub.DefaultRoute(operatorConfig)
+	customTLSSecret, configErr := c.ValidateDefaultRouteConfig(operatorConfig)
+	if configErr != nil {
+		return nil, "InvalidDefaultRouteConfig", configErr
```
spadgett, line 179: It's not necessarily invalid. It means we couldn't get the secret, right? It be that the name is wrong, but it could be due to other errors as well?

pkg/console/controllers/route/controller.go
```diff
@@ -224,6 +233,24 @@ func (c *RouteSyncController) SyncCustomRoute(operatorConfig *operatorsv1.Consol
...
+	}
+
+	secret, secretErr := c.secretClient.Secrets(api.OpenShiftConfigNamespace).Get(c.ctx, operatorConfig.Spec.Route.Secret.Name, metav1.GetOptions{})
+	if secretErr != nil {
+		return nil, fmt.Errorf("failed to GET default route TLS secret: %s", secretErr)
+	}
+	return secret, nil
+}
+
 func (c *RouteSyncController) ValidateCustomRouteConfig(operatorConfig *operatorsv1.Console) (*corev1.Secret, error) {
```
spadgett, line 254: Existing code, but this method is overloaded. It both gets and validates, which makes it harder to understand. We can leave this for now, but we might want to break this up into separate methods that each do one thing in the future.

PR #461, opened 2020-08-25
Bug 1862377: Notify user that the CLI binaries downloaded from console might be in different version then the server version

jhadvig:
Making this change as a suggested solution of this issue, based on https://bugzilla.redhat.com/show_bug.cgi?id=1862377#c7

/assign @spadgett

### DIFF
--- a/pkg/console/controllers/clidownloads/controller.go
+++ b/pkg/console/controllers/clidownloads/controller.go
@@ -199,6 +199,8 @@ func PlatformBasedOCConsoleCLIDownloads(host, cliDownloadsName string) *v1.Conso
 			Description: `With the OpenShift command line interface, you can create applications and manage OpenShift projects from a terminal.
 
 The oc binary offers the same capabilities as the kubectl binary, but it is further extended to natively support OpenShift Container Platform features.
+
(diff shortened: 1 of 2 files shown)


pkg/console/controllers/clidownloads/controller.go
```diff
@@ -199,6 +199,8 @@ func PlatformBasedOCConsoleCLIDownloads(host, cliDownloadsName string) *v1.Conso
 			Description: `With the OpenShift command line interface, you can create applications and manage OpenShift projects from a terminal.
 
 The oc binary offers the same capabilities as the kubectl binary, but it is further extended to natively support OpenShift Container Platform features.
+
+Note: The downloaded binary will report a different version based on the source code revision system rather than the server version.
```
spadgett, line 203: I feel like this will raise more questions than it answer. Can we reassign the bug to the cli component to see if we can't solve the underlying problem described here:

https://bugzilla.redhat.com/show_bug.cgi?id=1862377#c6

jhadvig: @spadgett was thinking about this, but since it's a know issue that we know about for a couple of releases (based on Lukes comment) I'm not sure if there is anything the CLI team will do about it.

PR #462, opened 2020-08-28
manifests: move downloads deployment to use python3

mfojtik:


spadgett:
Thanks @mfojtik. See also #463.

cc @yselkowitz

yselkowitz:
The script does need a few tweaks to work with python3, which (hopefully) are covered in my PR.  Maybe close this one and focus on that?

wking:
/close

Replaced by #462.

PR #464, opened 2020-08-28
Bug 1859331: Fix downloads index

yselkowitz:
The variable root is left over from the previous for loop, it shouldn't
be used again here.

/assign @wking


jhadvig (approved):
/retest

PR #470, opened 2020-09-18
Bug 1880307: Bump k8s vendors to 1.18.9. to fix bug in reflector not recovering from 'Too large resource version'

jhadvig:
Due to the bump I had to implement `GetObjectMeta()` for the operator client.

/assign @spadgett


jhadvig:
@spadgett PTAL

spadgett (approved):
/lgtm

PR #478, opened 2020-11-04
CONSOLE-2376: Console plugins

jhadvig:
Console-operator related changes based on the [enhancement doc](https://github.com/openshift/enhancements/pull/441/files).
/assign @spadgett


jhadvig:
@spadgett I've tested the PR with thew [API change](https://github.com/openshift/api/pull/764/files#diff-2d91c76ad45d9598ff5cf4b5a091f9641507a937859fca78eadcece3e6cae9e9).
I've put the plugins field into the `spec.customization.plugins`, but we can also put it directly on the `spec` (thought that would make a bit more sense to have it in the `customization` field).
After the field with and item is set on the operator config, the console-config ConfigMap is updated with it and the `console` pods are redeployed.

Updated operator config:
Screenshot 2020-11-05 at 11 55 10

Updated console-config CM:
Screenshot 2020-11-05 at 11 42 57


PTAL

jhadvig:
just FYI the TC's are failing cause I've removed the hardcoded API changes in the `vendor/`

pkg/console/operator/sync_v400.go
```diff
@@ -329,12 +331,18 @@ func (co *consoleOperator) SyncConfigMap(
 		}
 	}
 
+	pluginsList, pluginsListErr := co.consolePluginClient.List(ctx, metav1.ListOptions{})
+	if pluginsListErr != nil {
+		return nil, false, "FailedGetConsolePlugins", pluginsListErr
+	}
+	enabledPlugins := getEnabledPlugins(operatorConfig, pluginsList.Items)
```
vojtechszocs, line 338: :+1: This means the ConfigMap will contain an intersection of enabled plugins (operator config) and available plugins (`ConsolePlugin` custom resources).

@spadgett This shouldn't interfere with enabling a dynamic plugin as part of operator install flow, right? Console UI can start loading a plugin only when its assets are deployed and available via Service, which means only after the operator install process completes.

@jhadvig Will the ConfigMap be updated upon a change of either operator config or `ConsolePlugin` custom resources?

So if I understand correctly, operator config tracks all plugins marked as enabled (covers operator install flow and plugin admin page) and the ConfigMap tracks enabled plugins which should be loaded in the UI.

test/e2e/plugins_test.go
```diff
@@ -0,0 +1,115 @@
...
+		ObjectMeta: v1.ObjectMeta{
+			Name: pluginName,
+		},
+		Spec: consolev1.ConsolePluginSpec{
+			DisplayName: "TestPlugin",
+			Service: consolev1.ConsolePluginService{
+				Name:      "test-plugin-service-name",
+				Namespace: "test-plugin-service-namespace",
+				Port:      8443,
+				Manifest:  "/manifest",
```
vojtechszocs, line 57: Please update this field to `BasePath` according to https://github.com/openshift/api/pull/764#discussion_r533645627

vojtechszocs (changes requested):

pkg/console/operator/sync_v400.go
jhadvig: @vojtechszocs 
> This shouldn't interfere with enabling a dynamic plugin as part of operator install flow, right?

No it shouldn't. It's an mechanism that prevents console from trying to load plugins that are not present on the cluster as a `ConsolePlugin` CR

> Will the ConfigMap be updated upon a change of either operator config or ConsolePlugin custom resources?

Yes. Both changes should redeploy the console pods

>  So if I understand correctly, operator config tracks all plugins marked as enabled (covers operator install flow and plugin admin page) and the ConfigMap tracks enabled plugins which should be loaded in the UI.

Yes. operator config tracks all the enabled plugins, but since we can not guarantee that are of these plugins are available on the cluster (since the operator config can be updated manually as well) operator will check if a corresponding `ConsolePlugin` exists on the cluster. If so then it will put int into the `console-config.yaml`.

pkg/console/operator/operator.go
```diff
@@ -136,10 +143,11 @@ func NewConsoleOperator(
 			configV1Informers.OAuths().Informer(),
 		).WithFilteredEventsInformers( // console resources
 		targetNameFilter,
-		deployments.Informer(),
-		routes.Informer(),
+		deploymentInformer.Informer(),
+		routeInformer.Informer(),
 		serviceInformer.Informer(),
 		oauthClients.Informer(),
+		consolePluginInformer.Informer(),
```
stlaz, line 150: are you sure? The events from these informers are filtered so that only changes to resources of name `api.OpenShiftConsoleName` are caught.

The way I understand the API is that there are many objects of the `ConsolePlugin` type that you care about, but I might be wrong.

pkg/console/operator/operator.go
```diff
@@ -110,9 +116,10 @@ func NewConsoleOperator(
 		serviceClient:    corev1Client,
 		deploymentClient: deploymentClient,
 		// openshift
-		routeClient:   routev1Client,
-		oauthClient:   oauthv1Client,
-		versionGetter: versionGetter,
+		routeClient:         routev1Client,
+		oauthClient:         oauthv1Client,
+		consolePluginClient: consolePluginClient,
```
stlaz, line 121: you only appear to be doing a `List()` over console plugins, you should use a lister (it caches the data so that you don't need to do live requests to the API server)

pkg/console/operator/sync_v400.go
```diff
@@ -501,3 +509,38 @@ func getConsoleURL(route *routev1.Route) (string, error) {
 	}
 	return util.HTTPS(host), nil
 }
+
+// return plugins that are:
+// - enabled in the console-operator config
+// - created and available as a CR on the cluster
+func getEnabledPlugins(operatorConfig *operatorv1.Console, pluginsList []v1.ConsolePlugin) []string {
+	availablePluginsNames := []string{}
```
stlaz, line 517: use the power of sets instead of having to write this on your own https://github.com/kubernetes/kubernetes/blob/7d72ddca6e9a4e9cfa481b137bd0546be6a16cad/vendor/k8s.io/apimachinery/pkg/util/sets/string.go#L30

stlaz (changes requested):
QoL changes, both for the API server and the developer

pkg/console/operator/operator.go
jhadvig: @stlaz right, so we dont need to register the `consolePluginInformer` here ?

jhadvig:
@spadgett the PR is ready to review.
The logic is following:
When admin adds a list of ConsolePlugins into the console-operator config's `spec.plugins` field, only ConsolePlugins that are available on the cluster, meaning those that are not wont get saved. After that console-operator will fetch all the available and enabled ConsolePlugins CRs and creates a `map[string]string`, where the key is the name of the ConsolePlugin CR and the value is the endpoint to which the request from console will be proxied. The map is passed into the `console-config.yaml`.
I know that originally we wanted to pass only the list of enabled plugins into the `console-config.yaml`, but during development that I've realised that it would be easier just to assemble the endpoints by the console-operator and just pass them to the console in the config, cause:
- it removes duplicate API calls for fetching the ConsolePlugin CRs (console-operator when checking if the CRs are available and console itself when assembling the endpoints)
- it will  make local testing easier for the devs that will rewrite the static plugin into a dynamic one, since they can start the bridge with passing the `--plugins <pluginName>=<pluginEndpoint>` flag and point it to a locally running server on their host for serving assets
PTAL

pkg/console/operator/operator.go
```diff
@@ -136,8 +140,8 @@ func NewConsoleOperator(
 			configV1Informers.OAuths().Informer(),
 		).WithFilteredEventsInformers( // console resources
 		targetNameFilter,
-		deployments.Informer(),
-		routes.Informer(),
+		deploymentInformer.Informer(),
+		routeInformer.Informer(),
 		serviceInformer.Informer(),
 		oauthClients.Informer(),
 	).WithFilteredEventsInformers(
```
stlaz, line 147: `.WithInformers(consolePluginInformer)` to sync on console plugin events?

pkg/console/operator/sync_v400.go
```diff
@@ -501,3 +506,43 @@ func getConsoleURL(route *routev1.Route) (string, error) {
...
+		}
+		pluginsEndpointMap[pluginName] = getServiceHostname(plugin)
+	}
+	return pluginsEndpointMap
+}
+
+// return plugins that are:
+// - enabled in the console-operator config
+// - created and available as a CR on the cluster
+func getEnabledPlugins(operatorConfig *operatorv1.Console, pluginsList []*v1alpha1.ConsolePlugin) []string {
```
stlaz, line 525: this never gets called?

stlaz (changes requested):
looks like there's some dead code and the controller won't sync on console plugin events, otherwise looks good

jhadvig:
@stlaz thanks for review, comments addressed.

/hold for approvals
/assign @ahardin-rh @yapei @sferich888

yapei:
LGTM

pkg/console/operator/sync_v400.go
```diff
@@ -501,3 +504,24 @@ func getConsoleURL(route *routev1.Route) (string, error) {
 	}
 	return util.HTTPS(host), nil
 }
+
+func (co *consoleOperator) GetPluginsEndpointMap(enabledPluginsNames []string) map[string]string {
+	pluginsEndpointMap := map[string]string{}
+	for _, pluginName := range enabledPluginsNames {
+		plugin, err := co.consolePluginLister.Get(pluginName)
+		if err != nil {
+			klog.Errorf("failed to set service endpoint for %q plugin: %v", pluginName, err)
```
stlaz, line 513: you may want to consider going "Progressing: true" in this case

pkg/console/subresource/consoleserver/config_builder.go
```diff
@@ -267,3 +274,10 @@ func (b *ConsoleServerCLIConfigBuilder) providers() Providers {
 	}
 	return Providers{}
 }
+
+func (b *ConsoleServerCLIConfigBuilder) plugins() map[string]string {
+	if len(b.pluginsList) > 0 {
+		return b.pluginsList
+	}
+	return nil
```
stlaz, line 282: why do you avoid returning the "initialized, but no keys" path?

stlaz:
/lgtm
/hold
from what I've been told, you don't merge API before all related PRs get lgtm, so /holding for API merge

spadgett (approved):
/lgtm
/hold cancel

spadgett:
/hold

We probably need to delete the CI project or wait a few hours or this will keep failing. The record set didn't get cleaned up from a previous run.

```
level=error msg=Error: [ERR]: Error building changeset: InvalidChangeBatch: [Tried to create resource record set [name='api.ci-op-qlxhpr6t-4e531.origin-ci-int-aws.dev.rhcloud.com.', type='A'] but it already exists]
level=error msg=	status code: 400, request id: ac72bd30-3bee-4862-b034-4e30be6edc4c
level=error
level=error msg=  on ../tmp/openshift-install-882015745/route53/base.tf line 36, in resource "aws_route53_record" "api_external_alias":
level=error msg=  36: resource "aws_route53_record" "api_external_alias" {
level=error
level=error
level=fatal msg=failed to fetch Cluster: failed to generate asset "Cluster": failed to create cluster: failed to apply Terraform: failed to complete the change
error: failed to execute wrapped command: exit status 1 
```

PR #479, opened 2020-11-05
Add namespace and role resources for user settings

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4755

**Analysis / Root cause**:
Based on the [User Settings for OpenShift Console enhancement proposal](https://github.com/openshift/enhancements/blob/master/enhancements/console/user-settings.md) we want to create a `ConfigMap` for each individual user in the namespace `openshift-console-user-settings`

We create a `ConfigMap` with a service account from the console bridge, see PR https://github.com/openshift/console/pull/7095. For this we need that the namespace, role and rolebinding which is added in this PR.

**Solution Description**: 
* Add Namespace `openshift-console-user-settings`
* Add Role and RoleBinding `console-user-settings-role-operator` to allow the ServiceAccount to get, list, create, update and delete Role, RoleBinding and ConfigMap in this new namespace.

### DIFF
--- a/manifests/03-rbac-role-ns-openshift-console-user-settings.yaml
+++ b/manifests/03-rbac-role-ns-openshift-console-user-settings.yaml
@@ -1,0 +1,49 @@
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: console-user-settings-admin
+  namespace: openshift-console-user-settings
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+rules:
+- apiGroups:
+  - rbac.authorization.k8s.io/v1
+  resources:
+  - roles
+  - rolebindings
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
+  - delete
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
+  - delete
+  - watch
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: console-user-settings-admin
+  namespace: openshift-console-user-settings
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+roleRef:
+  kind: Role
+  name: console-user-settings-admin
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+  - kind: ServiceAccount
+    name: console
+    namespace: openshift-console
--- a/manifests/02-namespace.yaml
+++ b/manifests/02-namespace.yaml
@@ -17,3 +17,11 @@ metadata:
     openshift.io/node-selector: ""
   labels:
     openshift.io/cluster-monitoring: "true"
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-console-user-settings
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+    openshift.io/node-selector: ""


jerolimov:
@spadgett Can you take a look or assign this issue to someone who can review this change and https://github.com/openshift/console/pull/7095?

manifests/03-rbac-role-ns-openshift-console-user-settings.yaml
```diff
@@ -0,0 +1,45 @@
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: console-user-settings-role-operator
```
jhadvig, line 4: Lets drop the `-operator` suffix. It's a bit confusing since there is actually no operator that would be handling the `openshift-console-user-settings` namespace.
Also most of the time we are not adding the `-role` suffix as well since the resource is `Role` itself.
In case we want a name that would reflect the role, I would rather go with `console-user-settings-admin`

manifests/03-rbac-role-ns-openshift-console-user-settings.yaml
```diff
@@ -0,0 +1,45 @@
...
+  namespace: openshift-console-user-settings
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+rules:
+- apiGroups:
+  - rbac.authorization.k8s.io/v1
+  resources:
+  - role
```
jhadvig, line 12: Should be plural
```suggestion
  - roles
```

manifests/03-rbac-role-ns-openshift-console-user-settings.yaml
```diff
@@ -0,0 +1,45 @@
...
+  - rolebinding
```
jhadvig, line 13: Should be plural
```suggestion
  - rolebindings
```

manifests/03-rbac-role-ns-openshift-console-user-settings.yaml
```diff
@@ -0,0 +1,45 @@
...
+  - get
+  - list
+  - create
+  - update
+  - delete
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
```
jhadvig, line 24: Lets add `patch` and `watch`

manifests/03-rbac-role-ns-openshift-console-user-settings.yaml
```diff
@@ -0,0 +1,45 @@
...
+  namespace: openshift-console-user-settings
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+rules:
+- apiGroups:
+  - rbac.authorization.k8s.io/v1
+  resources:
+  - role
+  - rolebinding
+  verbs:
```
jhadvig, line 14: Lets add patch and watch

jhadvig (changes requested):

jerolimov:
Thanks @jhadvig for reviewing this. Applied all your recommendations :+1:

jerolimov:
Rebased and added a new label:

> console.openshift.io/user-settings: "true"

/retest

jhadvig:
@jerolimov noticed you added the `console.openshift.io/user-setting` label to all the resources. Havent found that label documented anywhere, can you please provide any info about it ?

jerolimov:
> @jerolimov noticed you added the `console.openshift.io/user-setting` label to all the resources. Havent found that label documented anywhere, can you please provide any info about it ?

@jhadvig You're right :+1: , based on [this discussion in slack](https://coreos.slack.com/archives/C01BGR3MJ1J/p1605827554007000?thread_ts=1605808630.005100&cid=C01BGR3MJ1J) I removed the labels again.

jhadvig (approved):
/lgtm
/approve

PR #480, opened 2020-11-10
Bump openshift/api to get QuickStarts changes

jhadvig:
The QuickStart CRD will be created via [Dockerfile](https://github.com/openshift/console-operator/blob/master/Dockerfile.rhel7#L16)
Once we have a `openshift/quick-starts` repo with all the [supported quickstarts](https://github.com/openshift/enhancements/blob/master/enhancements/console/quick-starts.md#implementation-details) will update the console-operator, so it creates all of the upon installation.
/assign @spadgett


spadgett (approved):
/lgtm

spadgett:
cc @andrewballantyne @christianvogt 

This adds the QuickStart CRD.

PR #483, opened 2020-11-23
Bug 1894216: Improve OpenShift Console availability

jhadvig:
Based on the bug description, we should avoid several minute outage in case of node failure pods are rescheduled to a healthy zone. For that reason we should use `RequiredDuringSchedulingIgnoredDuringExecution`, due to the [docs](https://kubernetes.io/docs/tutorials/stateful-application/zookeeper/#tolerating-node-failure).
Also 3 console replicas are recommended, although not sure if that's necessary.

/assign @spadgett

### DIFF
--- a/pkg/console/subresource/deployment/deployment.go
+++ b/pkg/console/subresource/deployment/deployment.go
@@ -115,7 +115,7 @@ func DefaultDeployment(operatorConfig *operatorv1.Console, cm *corev1.ConfigMap,
 									LabelSelector: &metav1.LabelSelector{
 										MatchLabels: util.SharedLabels(),
 									},
-									TopologyKey: "kubernetes.io/hostname",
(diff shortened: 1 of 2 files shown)


spadgett:
@jhadvig It doesn't look like this adds zone anti-affinity requested in the Bugzilla.

@smarterclayton @derekwaynecarr Any recommendations on what to set here? See https://bugzilla.redhat.com/show_bug.cgi?id=1894216

jhadvig:
@spadgett was following these docs:
- https://kubernetes.io/docs/tutorials/stateful-application/zookeeper/#tolerating-node-failure
- https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity

csrwng:
lgtm

PR #489, opened 2020-11-30
Bump(api) to get HelmChartRepository changes

pedjak:
Field `disabled` was added and needed by a few new UI functionalities.
Need to bump openshift/api because HelmChartRepository CRD is delivered within the operator payload


jhadvig (approved):
/lgtm
/approve

PR #493, opened 2020-12-06
add 'single-node-production-edge' annotations to CVO manifests.

osherdp:
This adds annotations for the single-node-production-edge cluster profile. There's a growing requirement from several customers to enable creation of single-node (not high-available) Openshift clusters.
In stage one (following openshift/enhancements#504) there should be no implication on components logic.
In the next stage, the component's behavior will match a non high-availability profile if the customer is specifically interested in one.
This PR is separate from the 'single-node-developer' work, which will implement a different behavior and is currently on another stage of implementation.

For more info, please refer to the enhancement link and participate in the discussion.

### DIFF
--- a/manifests/0000_90_console-operator_01_prometheusrbac.yaml
+++ b/manifests/0000_90_console-operator_01_prometheusrbac.yaml
@@ -8,6 +8,7 @@ metadata:
     include.release.openshift.io/ibm-cloud-managed: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
+    include.release.openshift.io/single-node-production-edge: "true"
 rules:
   - apiGroups:
       - ""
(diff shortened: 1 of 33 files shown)


jhadvig (changes requested):
Thanks for this PR.
Could you please also add the annotation to the [/quickstarts](https://github.com/openshift/console-operator/tree/master/quickstarts) and also update it's [Readme](https://github.com/openshift/console-operator/blob/master/quickstarts/README.md) with the new annotation.
Thanks

spadgett (commented):
Can you note that this annotation is required in the quickstarts README here?

https://github.com/openshift/console-operator/tree/master/quickstarts#contributing-quick-starts

osherdp:
@spadgett already did :)

osherdp:
Shifting strategy based on https://github.com/openshift/enhancements/pull/560

PR #499, opened 2021-01-19
Bug 1917759: Dont panic after setting plugin that does not exists to the console-operator config

jhadvig:
When we fail to fetch any of the set plugins in the console-operator config, dont set an endpoint for it and just skip it.
If the plugins will be created additionally, informer will trigger a next sync loop.

also update e2e test for adding available and unavailable plugin names to the console-operator config

/assign @spadgett


spadgett (approved):
/lgtm

PR #500, opened 2021-01-26
CONSOLE-2366: Move the operator's controllers to lib-go controller factory

jhadvig:
All operator's controllers should be utilising the controller factory from the `library-go`,  to get rid of the boilerplates that our controllers currently contain and are redundant.

/assign @spadgett @stlaz


pkg/console/controllers/route/controller.go
```diff
@@ -92,27 +81,25 @@ func NewRouteSyncController(
...
-		queue:        workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), controllerName),
-		ctx:          ctx,
+		resourceSyncer: resourceSyncer,
 	}
 
-	operatorClient.Informer().AddEventHandler(ctrl.newEventHandler())
-	operatorConfigInformer.Informer().AddEventHandler(ctrl.newEventHandler())
-	routeInformer.Informer().AddEventHandler(ctrl.newEventHandler())
+	configNameFilter := util.NamesFilter(api.ConfigResourceName)
+	targetNameFilter := util.NamesFilter(api.OpenShiftConsoleName)
```
stlaz, line 88: were these filtered before?

stlaz (commented):
the changes look good, I am only not sure where the config and target namefilters come from, but I assume they are correct

pkg/console/controllers/route/controller.go
jhadvig: @stlaz it looks like no

pkg/console/controllers/clidownloads/controller.go
```diff
@@ -84,28 +77,27 @@ func NewCLIDownloadsSyncController(
...
-	ctrl.cachesToSync = append(ctrl.cachesToSync,
-		operatorClient.Informer().HasSynced,
-		operatorConfigInformer.Informer().HasSynced,
-		consoleCLIDownloadsInformers.Informer().HasSynced,
-		routesInformers.Informer().HasSynced,
-	)
-
-	return ctrl
+	configNameFilter := controllersutil.NamesFilter(api.ConfigResourceName)
+	targetNameFilter := controllersutil.NamesFilter(api.OpenShiftConsoleName)
```
stlaz, line 84: wrong name?

pkg/console/controllers/clidownloads/controller.go
```diff
@@ -136,13 +128,13 @@ func (c *CLIDownloadsSyncController) sync() error {
 
 	statusHandler := status.NewStatusHandler(c.operatorClient)
 	ocConsoleCLIDownloads := PlatformBasedOCConsoleCLIDownloads(host, api.OCCLIDownloadsCustomResourceName)
-	_, ocCLIDownloadsErrReason, ocCLIDownloadsErr := ApplyCLIDownloads(c.consoleCliDownloadsClient, ocConsoleCLIDownloads, c.ctx)
+	_, ocCLIDownloadsErrReason, ocCLIDownloadsErr := ApplyCLIDownloads(c.consoleCliDownloadsClient, ocConsoleCLIDownloads, ctx)
```
stlaz, line 131: context should be the first parameter: https://golang.org/pkg/context/

pkg/console/controllers/clidownloads/controller.go
```diff
@@ -136,13 +128,13 @@ func (c *CLIDownloadsSyncController) sync() error {
...
 	statusHandler.AddCondition(status.HandleDegraded("OCDownloadsSync", ocCLIDownloadsErrReason, ocCLIDownloadsErr))
 	if ocCLIDownloadsErr != nil {
 		return statusHandler.FlushAndReturn(ocCLIDownloadsErr)
 	}
 
-	_, odoCLIDownloadsErrReason, odoCLIDownloadsErr := ApplyCLIDownloads(c.consoleCliDownloadsClient, ODOConsoleCLIDownloads(), c.ctx)
+	_, odoCLIDownloadsErrReason, odoCLIDownloadsErr := ApplyCLIDownloads(c.consoleCliDownloadsClient, ODOConsoleCLIDownloads(), ctx)
```
stlaz, line 137: ctx ditto

pkg/console/controllers/resourcesyncdestination/controller.go
```diff
@@ -27,118 +26,103 @@ const (
...
 func NewResourceSyncDestinationController(
+	operatorClient v1helpers.OperatorClient,
 	// operatorconfig
 	operatorConfigClient operatorclientv1.ConsoleInterface,
 	operatorConfigInformer operatorinformersv1.ConsoleInformer,
-	// configmap
-	corev1Client coreclientv1.CoreV1Interface,
-	configMapInformer coreinformersv1.ConfigMapInformer,
+	// openshift managed
+	managedCoreV1 coreinformerv1.Interface,
```
stlaz, line 42: nit: `managedCoreInformer` to match the other variable names

pkg/console/controllers/resourcesyncdestination/controller.go
```diff
@@ -27,118 +26,103 @@ const (
...
-
-	return ctrl
+	configNameFilter := util.NamesFilter(api.ConfigResourceName)
+
+	return factory.New().
+		WithFilteredEventsInformers( // configs
+			configNameFilter,
+			operatorConfigInformer.Informer(),
+		).WithFilteredEventsInformers(
+		util.NamesFilter(api.DefaultIngressCertConfigMapName),
```
stlaz, line 63: this does not seem to be the only CM that gets synced in this controller

pkg/console/controllers/resourcesyncdestination/controller.go
```diff
@@ -27,118 +26,103 @@ const (
...
-		DeleteFunc: func(obj interface{}) { c.queue.Add(controllerWorkQueueKey) },
-	}
+func (c *ResourceSyncDestinationController) syncIngressCert(ctx context.Context) error {
+	// sync: 'default-ingress-cert' configmap
+	// from: 'openshift-config-managed' namespace
+	// to:   'openshift-console' namespace
+	return c.resourceSyncer.SyncConfigMap(
+		resourcesynccontroller.ResourceLocation{Name: api.DefaultIngressCertConfigMapName, Namespace: api.OpenShiftConsoleNamespace},
+		resourcesynccontroller.ResourceLocation{Name: api.DefaultIngressCertConfigMapName, Namespace: api.OpenShiftConfigManagedNamespace},
+	)
```
stlaz, line 127: These appear to be static, you can just move them out to the starter, just like 
https://github.com/openshift/cluster-authentication-operator/blob/cf1e1a6c79db7cf29e7de2a90ecdc458bc13059c/pkg/operator/starter.go#L198-L204

```go
	// add syncing for the OAuth metadata ConfigMap
	if err := operatorCtx.resourceSyncController.SyncConfigMap(
		resourcesynccontroller.ResourceLocation{Namespace: "openshift-config-managed", Name: "oauth-openshift"},
		resourcesynccontroller.ResourceLocation{Namespace: "openshift-authentication", Name: "v4-0-config-system-metadata"},
	); err != nil {
		return err
	}
```

pkg/console/controllers/route/controller.go
```diff
@@ -92,27 +81,25 @@ func NewRouteSyncController(
...
-		routeInformer.Informer().HasSynced,
-	)
-
-	return ctrl
+	return factory.New().
+		WithFilteredEventsInformers( // configs
+			configNameFilter,
+			operatorConfigInformer.Informer(),
+		).WithFilteredEventsInformers( // console resources
+		targetNameFilter,
```
stlaz, line 95: there appear to be at least 2 different routes handled in this controller

pkg/console/controllers/route/controller.go
```diff
@@ -92,27 +81,25 @@ func NewRouteSyncController(
...
-		operatorConfigInformer.Informer().HasSynced,
-		routeInformer.Informer().HasSynced,
-	)
-
-	return ctrl
+	return factory.New().
+		WithFilteredEventsInformers( // configs
+			configNameFilter,
+			operatorConfigInformer.Informer(),
+		).WithFilteredEventsInformers( // console resources
```
stlaz, line 94: shouldn't you react to changes of the CMs and secrets, too?

pkg/console/controllers/service/controller.go
```diff
@@ -80,31 +74,28 @@ func NewServiceSyncController(
...
-
-	ctrl.cachesToSync = append(ctrl.cachesToSync,
-		operatorClient.Informer().HasSynced,
-		operatorConfigInformer.Informer().HasSynced,
-		serviceInformer.Informer().HasSynced,
-	)
-
-	return ctrl
+	configNameFilter := util.NamesFilter(api.ConfigResourceName)
+	targetNameFilter := util.NamesFilter(api.OpenShiftConsoleName)
```
stlaz, line 81: this is managing 2 different services, correct?

pkg/console/starter/starter.go
```diff
@@ -317,34 +308,21 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
...
-	go consoleRouteController.Run(1, ctx.Done())
-	go resourceSyncDestinationController.Run(1, ctx.Done())
-	go cliDownloadsController.Run(1, ctx.Done())
 	// go staleConditionsController.Run(1, ctx.Done())
 
 	<-ctx.Done()
 	return fmt.Errorf("stopped")
 }
 
-// startResourceSyncing should start syncing process of all secrets and configmaps that need to be synced.
```
stlaz, line 335: should've stayed here

stlaz (changes requested):

pkg/console/controllers/resourcesyncdestination/controller.go
jhadvig: there is another custom-logo CM, but it's name is set by the admin on the console-operator config, and we read it from there. Since the name is arbitrary we can't set the informer.

pkg/console/controllers/route/controller.go
jhadvig: yes, but not the Secrets name, since those names are arbitrary 👍

jhadvig:
@stlaz thanks for review, comments addressed and replied to a few comments.
Removed the `resourcesyncdestination/controller.go` since it was not necessary.

PTAL

pkg/console/controllers/resourcesyncdestination/controller.go
stlaz: in that case, you need to watch all in the given NS, see `WithNamespaceInformer()`

stlaz (approved):
Looks good to me now. I hope the informer event filtering is fixed after we've discussed it with Jakub and after I pointed out the issues on a couple of places throughout the PR.

The rule of thumb is "the operator needs to watch for all objects on its input and output".

I've also asked Jakub to try the excercise of removing the `openshift-console` NS to see if the operator is capable of pushing all necessary objects back into it. That should be a good springboard to saying "it's all good and working" in case some of the controllers are untested to reacting to changes.

I'm leaving the final LGTM to the members of the console team after Jakub's confirmed all is fine and well after the above excercise :slightly_smiling_face:

pkg/console/controllers/resourcesyncdestination/controller.go
```diff
@@ -1,144 +0,0 @@
...
-		return nil
-	case operatorsv1.Removed:
-		klog.V(4).Infoln("console is in an removed state: removing synced default-ingress-cert configmap")
-		return c.removeDefaultIngressCertConfigMap()
-	default:
-		return fmt.Errorf("unknown state: %v", operatorConfig.Spec.ManagementState)
-	}
-
-	return err
-}
```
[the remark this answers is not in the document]
jhadvig: nope :-/

jhadvig:
@stlaz I've addressed additional comments.

> I've also asked Jakub to try the excercise of removing the openshift-console NS to see if the operator is capable of pushing all necessary objects back into it. That should be a good springboard to saying "it's all good and working" in case some of the controllers are untested to reacting to changes.

tested this case and all got back to the working state 👍 

@spadgett PTAL

spadgett:
for approvals

@ahardin-rh This only refactors the code and has no doc impact.

spadgett:
@yapei This just needs a regression test of the console operator and does not add any new features.

spadgett:
I love how much unnecessary code we're able to remove with this change 👍

yapei:
QE did regression testing of console-operator and no issues found

spadgett (approved):
/lgtm

PR #504, opened 2021-02-08
RetryOnConflict should be fetching latest version of console-operator config

jhadvig:
Fixing occasional flakes, since inside the `RetryOnConflict` loop we should be fetching latest version of the console-operator config.

/assign @spadgett


spadgett (approved):
/lgtm

PR #507, opened 2021-02-10
Bump polling interval when checking for resource unavailability

jhadvig:
Due to the small polling interval the `RemovedTest` has been flaking, since we didnt get all the operator sync loop enough between testing if all the resources are removed and not recreated.

/assign @spadgett


jhadvig:
flaky test
/retest

spadgett (approved):
/lgtm

PR #508, opened 2021-02-17
CONSOLE-2526: Update console operator for single-node clusters

jhadvig:
Console-operator updates:
- number of replicas and anti-affinity of `console` and `downloads` deployments should determined based on the infrastructure configs status `infrastructureTopology` field. If set to `HighlyAvailable` is set, then number of replicas are set to `2` and anti-affinity  are set. In case of `SingleReplica`, only `1` pod replica is set and no anti-affinity rules are set.
- `downloads` deployment is removed from the `/manifest` and is now managed by console-operator as a separate controller, so it does reflect changes in the infrastructure config. A `legacy_manifests` dir has been created where the original manifest was moved, in case we need it in the future.
- add unit
- code refactoring + cleanUp

Note: needed to test the change manually by creating a single-node cluster. Since the `infrastructureTopology` status field is immutable and set only upon the cluster provisioning.

/assign @spadgett


jhadvig:
/hold
for approvals

@spadgett the PR is ready for review

pkg/console/controllers/downloadsdeployment/controller.go
```diff
@@ -0,0 +1,139 @@
...
+
+	updatedOperatorConfig := operatorConfig.DeepCopy()
+	requiredDownloadsDeployment := deploymentsub.DefaultDownloadsDeployment(operatorConfig, infrastructureConfig)
+
+	_, _, applyErr := resourceapply.ApplyDeployment(
+		c.deploymentClient,
+		controllerContext.Recorder(),
+		requiredDownloadsDeployment,
+		resourcemerge.ExpectedDeploymentGeneration(requiredDownloadsDeployment, updatedOperatorConfig.Status.Generations),
+	)
```
stlaz, line 132: Should there be more detailed conditions describing the status of this deployment? Is this done elsewhere?

pkg/console/controllers/downloadsdeployment/controller.go
```diff
@@ -0,0 +1,139 @@
...
+			operatorConfigInformer.Informer(),
+			configV1Informers.Infrastructures().Informer(),
+		).WithFilteredEventsInformers( // downloads deployment
+		downloadsNameFilter,
+		deploymentInformer.Informer(),
+	).WithSync(ctrl.Sync).
+		ToController("ConsoleDownloadsDeploymentSyncController", recorder.WithComponentSuffix("console-downloads-deployment-controller"))
+}
+
+func (c *DownloadsDeploymentSyncController) Sync(ctx context.Context, controllerContext factory.SyncContext) error {
```
stlaz, line 86: is this copied from somewhere?

stlaz (changes requested):
you should split the changes related to the new deployment controller and to the single-node cluster into separate commits

legacy_manifests/cluster-downloads-deployment.yaml
```diff
@@ -0,0 +1,347 @@
+kind: Deployment
```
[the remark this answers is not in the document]
jhadvig: wanted to keep the original yaml just in case .... 
but on the other hand we have git, will remove 👍

pkg/console/controllers/downloadsdeployment/controller.go
jhadvig: well all the controllers have the similar `Sync()` method. not sure why you are asking?

pkg/console/controllers/downloadsdeployment/controller.go
jhadvig: yes, when syncing `console` deployment.
https://github.com/openshift/console-operator/blob/master/pkg/console/operator/sync_v400.go#L247

pkg/console/subresource/deployment/deployment.go
```diff
@@ -157,6 +130,60 @@ func DefaultDeployment(operatorConfig *operatorv1.Console, cm *corev1.ConfigMap,
...
+							LivenessProbe:  defaultDownloadsProbe(),
+							Command:        []string{"/bin/sh"},
+							Resources:      containerResourceRequirements("10m", "50Mi"),
+							Args:           downloadsContainerArgs(),
+						},
+					},
+				},
+			},
+		},
+	}
```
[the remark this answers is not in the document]
jhadvig: After talking to @stlaz we decided to address this change globally in terms of the whole console-operator.

pkg/console/subresource/deployment/deployment.go
```diff
@@ -450,3 +548,118 @@ func customLogoVolume() volumeConfig {
 		path:        "/var/logo/",
 		isConfigMap: true}
 }
+
+func downloadsContainerArgs() []string {
```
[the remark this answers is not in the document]
jhadvig: After talking to @stlaz we decided to address this change globally in terms of the whole console-operator.

jhadvig:
> you should split the changes related to the new deployment controller and to the single-node cluster into separate commits

the new downloads deployment controller is there for the purpose of addressing the single-node cluster functionality.

pkg/console/subresource/deployment/deployment.go
```diff
@@ -184,6 +211,64 @@ func LogDeploymentAnnotationChanges(client appsclientv1.DeploymentsGetter, updat
...
+			Key:               "node.kubernetes.io/not-reachable",
+			Operator:          corev1.TolerationOpExists,
+			Effect:            corev1.TaintEffectNoExecute,
+			TolerationSeconds: &tolerationSeconds,
+		},
+	}
+}
+
+func affinity(infrastructureConfig *configv1.Infrastructure) *corev1.Affinity {
+	if infrastructureConfig.Status.InfrastructureTopology == configv1.SingleReplicaTopologyMode {
```
spadgett, line 246: I believe the two deployments should be looking at different properties. Since console itself has a label selector for master nodes, it should look at `controlPlaneTopology`. The downloads deployment runs on any Linux node, so it should look for `infrastructureTopology`.

Makefile
```diff
@@ -36,7 +36,7 @@ test-unit:
 .PHONY: test-unit
 
 test-e2e:
-	KUBERNETES_CONFIG=${KUBECONFIG} go test -timeout 30m -v ./test/e2e/
+	KUBERNETES_CONFIG=${KUBECONFIG} go test -timeout 30m -v ./test/e2e/ -run TestUnmanaged
```
spadgett, line 39: Was this intentionally committed?

spadgett (commented):
Thanks, @jhadvig 👍  Have you been able to test an upgrade scenario? Is the console operator able to take over the existing downloads deployment from an older version?

spadgett:
@romfreiman This is the PR adding single node support for console

@wking FYI, this changes how we create the downloads deployment

Makefile
jhadvig: yes, `git add .` .... will remove

pkg/api/api.go
```diff
@@ -41,4 +43,6 @@ const (
 	RedirectContainerPort       = 8444
 	RedirectContainerTargetPort = RedirectContainerPort
 	ConsoleServingCertName      = "console-serving-cert"
+	DownloadsPortName           = "http"
+	DownloadsPort               = 8080
```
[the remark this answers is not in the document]
jhadvig: Would rather do it as followUp

manifests/07-downloads-deployment.yaml
```diff
@@ -1,177 +0,0 @@
-apiVersion: apps/v1
```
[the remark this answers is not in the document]
jhadvig: No we dont need to remove them since they are static.
If it's ok, I would rather keep it simple and create additional card for it, since the the PR is pretty big even now.

pkg/console/subresource/deployment/deployment.go
jhadvig: @spadgett  update, but thinking if it makes sense to keep the anti-affinity rules for the `downloads` deployment?

jhadvig:
@spadgett I've addressed your comments. Also added a comments to those that are unresolved. PTAL

spadgett (approved):
/lgtm

PR #509, opened 2021-02-22
Bug 1931760: Non-unified reference to OCS in quickstarts

rexagod:
Fixes (a part of): https://bugzilla.redhat.com/show_bug.cgi?id=1930109


rexagod:
image

rexagod:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_console-operator/509/pull-ci-openshift-console-operator-master-e2e-aws-console/1363932895818813440#1:build-log.txt%3A531

jhadvig (changes requested):
@rexagod thanks for the fix. Should we also remove the `®` character from the `displayName` fields in other ConsoleQuickStarts CRs as well?
https://github.com/openshift/console-operator/blob/master/quickstarts/explore-pipeline-quickstart.yaml#L11
https://github.com/openshift/console-operator/blob/master/quickstarts/install-serverless.yaml#L17

rexagod:
While that would go outside the scope of this issue, I'm personally +1 for it. Should I ping Eran, Yuval, and Sam and take their feedbacks as well?

jhadvig:
> While that would go outside the scope of this BZ, I'm personally +1 for it. Should I ping Eran, Yuval, and Sam and take their feedbacks as well?

@rexagod I would just fix it, since its a two extra oneliners.

spadgett:
We should be consistent in all of the quick starts whatever we decide.

@openshift/quickstart-ux-content

quickstarts/install-serverless.yaml
```diff
@@ -14,15 +14,15 @@ spec:
...
   description: >-
     Install the OpenShift Serverless Operator to deploy stateless,
     event-trigger-based applications.
   introduction: >-
-    Red Hat® OpenShift® Serverless lets you run stateless, serverless workloads
+    Red Hat® OpenShift Serverless lets you run stateless, serverless workloads
```
rexagod, line 25: This is the only ® remaining this repository after this commit.

rexagod (commented):
This commit (https://github.com/openshift/console-operator/pull/509/commits/a91fd15805d8317b332b2d2d14b7466990c379cd) aims to make the spelling consistent across this repository.

Based on the aforementioned team's review, we can choose to incorporate or drop this commit.

bmignano:
I believe when we wrote the first quick start (Serverless) someone from the docs team had reached out to legal and they were in favor of marking the first appearance of trademarked terms in each quick start. Agree that we should be consistent across all OCS tours, but I'd defer to @abigaeljamie here in terms of the specifics!

abigaeljamie:
Thanks, Brie! 

I'd align with Legal's advice for all quick starts so that we have consistency across the board: Mark the first appearance of trademarked terms in each quick start. So if this is the first mention of the term in this quick start, we'd keep the trademark, but we can drop it in all later mentions.

rexagod:
I followed up on this issue with @abigaeljamie who contacted Max (the original person who consulted Legal about using trademarks) and he gave the following input:

> Legal's guidance at the time was that marking trademarked terms on their first appearance in quick start body text (say, an intro) would be "good."

I've made changes based on this in my latest commit (https://github.com/openshift/console-operator/pull/509/commits/5d2002818ac7ac2056d3920a73f9a68b32d65424).

rexagod:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_console-operator/509/pull-ci-openshift-console-operator-master-e2e-agnostic-upgrade/1364679832411049984#1:build-log.txt%3A79

jhadvig:
Based on comment @abigaeljamie we should have the trademark logo only on the `displayName` property, right? All other occurrences of the trademark logo after the "OpenShift" should be dropped. @rexagod please update the PR accordingly.
Also please dont forget to squash into a single commit.

rexagod:
Quoting @abigaeljamie:

> I'm going to touch base with Legal again to see if they still align with that guideline, but for the time being, we can follow that. So we won't include the trademark on the tile—just in the body text in the first mention of the product name.

I made these changes based on this. Changing the `displayName` was my first thought as well, but it's explicitly stated here that the change needs to be in the body text and not the title. Maybe we should wait for her opinion on this, or I can make your suggested changes if you want that.

jhadvig:
@rexagod ok, let's go with the suggestion and just put the trademark on the with occurrence in the text, would go with `introduction` property here.

@rexagod could you please go and change this across all the QuickStarts in this repository? I seethere are couple of inconsistencies, eg:
- remove the trademark after "Red Hat" word in https://github.com/openshift/console-operator/blob/master/quickstarts/install-serverless.yaml#L25
- add the trademark after https://github.com/openshift/console-operator/blob/master/quickstarts/sample-application-quickstart.yaml#L16

please dont forget to squash. thanks

jhadvig (approved):
/lgtm
/approve

Thanks @rexagod for this fix 👍

abigaeljamie:
> Quoting @abigaeljamie:
> 
> > I'm going to touch base with Legal again to see if they still align with that guideline, but for the time being, we can follow that. So we won't include the trademark on the tile—just in the body text in the first mention of the product name.
> 
> I made these changes based on this. Changing the `displayName` was my first thought as well, but it's explicitly stated here that the change needs to be in the body text and not the title. Maybe we should wait for her opinion on this, or I can make your suggested changes if you want that.

I was able to touch base with Erin in legal. She said that trademarks are not used in titles, but they should be used in the body text upon first use of the product name. 

Because tiles are titles, we can omit the trademarks there and instead use them in the body text upon first mention of the product name. So the advice hasn't changed since Max checked with her 👍

PR #510, opened 2021-03-02
Add e2e test to check correct deployment replicas, based on the Infrastructure config TopologyMode

jhadvig:
adding e2e test to check the `console` && `downloads` deployments replicas, based on the Infrastructure config, since the release team will by adding and enabling single-node cluster prow job to this repo.

/assign @spadgett

### DIFF
--- a/pkg/console/subresource/deployment/deployment_test.go
+++ b/pkg/console/subresource/deployment/deployment_test.go
@@ -554,7 +554,7 @@ func TestReplicas(t *testing.T) {
 			want: 1,
 		},
 		{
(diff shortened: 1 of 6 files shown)


pkg/console/subresource/deployment/deployment_test.go
```diff
@@ -567,7 +567,7 @@ func TestReplicas(t *testing.T) {
 	}
```
omertuc, line 567: Two of the tests above this comment have the same name

jhadvig:
@omertuc I've addressed the comments. Thanks! :)

PR #511, opened 2021-03-05
Bug 1935541: Check for error when generating default and user-defined config for the console-config configmap

jhadvig:
/assign @spadgett 
/cherry-pick release-4.7


spadgett (approved):
/lgtm

PR #513, opened 2021-03-09
Add in get subscriptions to console service account

JPinkney:
This PR adds in additional permissions that are needed for https://github.com/openshift/console/pull/8342. We are using it to check if the web-terminal operator is installed on the cluster.

Related issue: https://issues.redhat.com/browse/WTO-20

Signed-off-by: Josh Pinkney <joshpinkney@gmail.com>

### DIFF
--- a/manifests/03-rbac-role-cluster.yaml
+++ b/manifests/03-rbac-role-cluster.yaml
@@ -105,6 +105,14 @@ rules:
   - validatingwebhookconfigurations
   verbs:
   - get
+- apiGroups:
+  - operators.coreos.com
+  resources:
+  - subscriptions
+  resourceNames:
+  - web-terminal
+  verbs:
+  - get
 ---
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1


jhadvig (changes requested):
@JPinkney you are adding the permission to the `console-operator` SA. Based on the [console change](https://github.com/openshift/console/pull/8342) you linked, the request is done by the console itself, right? In that case the permission should be added to `03-rbac-role-ns-console.yaml` file

manifests/03-rbac-role-cluster.yaml
```diff
@@ -105,6 +105,12 @@ rules:
   - validatingwebhookconfigurations
   verbs:
   - get
+- apiGroups:
+  - operators.coreos.com
+  resources:
+  - subscriptions
+  verbs:
+  - get
```
spadgett, line 113: It's better to restrict the permission to the single subscription name / namespace we need if it's a well-known value.

JPinkney:
Maybe there's something I'm missing but moving my changes from `03-rbac-role-cluster.yaml` to `03-rbac-role-ns-console.yaml` yields:
```
Failed to check if the web-terminal operator is installed: subscriptions.operators.coreos.com "web-terminal" is forbidden: User "system:serviceaccount:openshift-console:console" cannot get resource "subscriptions" in API group "operators.coreos.com" in the namespace "openshift-operators"
```

I'm trying to add the required permissions to the `console` service account not the `console-operator` one

jhadvig:
@JPinkney sorry, you are right, overlooked that the `console` ClusterRole is in the `03-rbac-role-cluster.yaml`.

manifests/03-rbac-role-cluster.yaml
jhadvig: yes, please add `resourceNames` field with appropriate name of the subscription instance.

JPinkney: I've added in the resourceNames field

jhadvig (changes requested):
@JPinkney please squash so we can merge this one 👍

jhadvig (approved):
/lgtm

PR #517, opened 2021-03-18
CONSOLE-2792: Bump openshift/api dependecy in console-operator to get CustomDomain API for Ingress config

jhadvig:
/assign @spadgett

PR #519, opened 2021-03-23
Bump openshift/api to get project access role cuztomization changes

rottencandy:
Issue: https://issues.redhat.com/browse/ODC-5447

Bump openshift/api to get project access role customization changes from https://github.com/openshift/api/pull/857


spadgett:
We also have #517 open. We should probably consolidate these into one PR.

/cc @jhadvig

jhadvig (approved):
/lgtm

PR #521, opened 2021-04-06
Bug 1945326: Resync controllers every minute

jhadvig:
We should be periodically resyncing each controller.
/assign @spadgett


spadgett (approved):
/lgtm

PR #522, opened 2021-04-06
CONSOLE-2793: Implement console-operator changes to consume new CustomDomains API

jhadvig:
Implementing new custom routes functionally to use new [ingresses config route configuration API](https://github.com/openshift/api/pull/852), for `console` and `downloads` routes.
Console-operator already contains custom route customisation, for the `console` route only.

In case the `console` custom route would be setup in both ingress config and console-operator config, then the new ingress config custom route configuration will take precedent.

To make the it easier to read the route and service configuration for any given case Im adding the go bindata, that contain those configurations and would be used per specify use-case.

PR is also adding additional e2e tests.

Story: https://issues.redhat.com/browse/CONSOLE-2793

@awgreene @spadgett PTAL


pkg/console/subresource/route/route.go
```diff
@@ -35,6 +36,137 @@ type CustomTLSCert struct {
...
+		customRoute.hostname = string(componentRouteSpec.Hostname)
+		if len(componentRouteSpec.ServingCertKeyPairSecret.Name) != 0 {
+			customRoute.secretName = componentRouteSpec.ServingCertKeyPairSecret.Name
+		}
+		isIngressConfigCustomHostnameSet = true
+	}
+
+	// Legacy behaviour, operatorConfig only configures custom hostname for "console" route.
+	// The custom route hostname doesn't have to be set if admin just wants to set the custom
+	// TLS for the default route.
```
spadgett, line 90: I'm a little unclear on this comment. Are both these scenarios supported in the new API? (Setting hostname without certs or setting certs without hostname?)

test/e2e/custom_url_test.go
```diff
@@ -19,90 +20,307 @@ import (
...
+	DefaultRouteName          string
+	CustomRouteName           string
+	CustomRouteHostname       string
+	CustomRouteHostnamePrefix string
+	LegacySetup               bool
+	SkipRouteCheck            bool
+	CustomTLSSecretName       string
+}
+
+func (tc *testCaseConfig) setup(t *testing.T, client *framework.ClientSet) {
```
spadgett, line 47: Thank you for all the tests 👍

spadgett (commented):
Love how much code we are able to remove with the bindata change 👍 

for approvals

yanpzhan:
I tested the scenario when add custom console and downloads route in ingress config with custom domain/cert, the new route could access successfully.
Will do more test later.

yanpzhan:
After test against the pr, the main feature works, here is the test case:  https://polarion.engineering.redhat.com/polarion/redirect/project/OSE/workitem?id=OCP-41393 
could you check if there is other checkpoint need to be tested? @jhadvig thanks!

jhadvig:
@yanpzhan TC incorporates all the cases we need to check. LGTM 👍

jhadvig:
Rebased PR since there was a conflict in the `go.sum`. No other changes where done.

pkg/console/subresource/route/route.go
jhadvig: @spadgett it's up to each component how they implement the logic. Since we already had the functionality I've kept it for the console-operator.

bindata/routes/console-custom-route.yaml
```diff
@@ -0,0 +1,21 @@
...
+  to:
+    kind: Service
+    name: console
+    weight: 100
+  port:
+    targetPort: https
+  tls:
+    termination: reencrypt
+    insecureEdgeTerminationPolicy: Redirect
+  wildcardPolicy: None
```
awgreene, line 21: Nit: newline.

bindata/routes/console-redirect-route.yaml
```diff
@@ -0,0 +1,22 @@
...
+  to:
+    kind: Service
+    name: console-redirect
+    weight: 100
+  port:
+    targetPort: custom-route-redirect
+  tls:
+    termination: edge
+    insecureEdgeTerminationPolicy: Redirect
+  wildcardPolicy: None
```
awgreene, line 22: Nit: newline

bindata/routes/console-route.yaml
```diff
@@ -0,0 +1,20 @@
...
+  to:
+    kind: Service
+    name: console
+    weight: 100
+  port:
+    targetPort: https
+  tls:
+    termination: reencrypt
+    insecureEdgeTerminationPolicy: Redirect
+  wildcardPolicy: None
```
awgreene, line 20: Nit: newline

bindata/routes/downloads-custom-route.yaml
```diff
@@ -0,0 +1,18 @@
...
+spec:
+  tls:
+    termination: edge
+    insecureEdgeTerminationPolicy: Redirect
+  port:
+    targetPort: http
+  to:
+    kind: Service
+    name: downloads
+  wildcardPolicy: None
```
awgreene, line 18: Nit: newline

bindata/routes/downloads-route.yaml
```diff
@@ -1,12 +1,10 @@
...
+# The 'downloads' route will be pointing to the 'downloads' service. 
 apiVersion: route.openshift.io/v1
 kind: Route
 metadata:
   namespace: openshift-console
   name: downloads
-  annotations:
-    include.release.openshift.io/ibm-cloud-managed: "true"
-    include.release.openshift.io/self-managed-high-availability: "true"
-    include.release.openshift.io/single-node-developer: "true"
```
awgreene, line 9: Why aren't routes not included in these releases?

bindata/services/console-service.yaml
```diff
@@ -0,0 +1,22 @@
...
+  ports:
+    - name: https
+      protocol: TCP
+      port: 443
+      targetPort: 8443
+  selector:
+    app: console
+    component: ui
+  type: ClusterIP
+  sessionAffinity: None
```
awgreene, line 22: Nit: newline

bindata/services/downloads-service.yaml
```diff
@@ -16,3 +14,5 @@ spec:
   selector:
     app: console
     component: downloads
+  type: ClusterIP
+  sessionAffinity: None
```
awgreene, line 18: Nit: newline

bindata/services/downloads-service.yaml
```diff
@@ -1,12 +1,10 @@
+# Default 'downloads' service manifest.
+# The 'downloads' route will be pointing to the 'downloads' deployment. 
 apiVersion: v1
 kind: Service
 metadata:
   namespace: openshift-console
   name: downloads
```
sferich888, line 7: Should this be renamed to be downoads custom? to make it more clear that this is the customized service downloads, should someone try and use these files as an example of how to use this feature.

pkg/console/subresource/configmap/configmap.go
```diff
@@ -110,6 +109,13 @@ func DefaultConfigMap(
 	return configMap, willMergeConfigOverrides, nil
 }
 
+func isCustomRoute(activeRoute *routev1.Route) bool {
+	if activeRoute.GetName() == api.OpenShiftConsoleRouteName {
+		return false
+	}
+	return true
+}
```
awgreene, line 117:
```suggestion
	return activeRoute.GetName() == api.OpenShiftConsoleRouteName 
```

sferich888 (approved):
I think this looks good; but I wonder if the examples could be clearer; to better show the usecaes that we are adding with this. 

- How to alter the console URL
- How to alter the console URL + connection type (redirect vs re-encrypt)
- How to alter the downloads URL

IMO we need to clearly show how we can do all 3 with the examples (and possibly others I am leaving out); so that this can be properly documented. 

cc: @sjstout

pkg/console/controllers/route/controller_test.go
```diff
@@ -200,29 +200,3 @@ func TestValidateCustomCertSecret(t *testing.T) {
 		})
 	}
 }
-
-func TestGetDefaultRouteHost(t *testing.T) {
```
awgreene, line 204: Could you explain the decision to remove these unit tests?

pkg/console/subresource/route/route.go
```diff
@@ -35,6 +36,137 @@ type CustomTLSCert struct {
...
+type RouteControllerSpec struct {
+	hostname   string
+	secretName string
+}
+
+func getComponentRouteSpec(ingressConfig *configv1.Ingress, componentName string) *configv1.ComponentRouteSpec {
+	for i, componentRoute := range ingressConfig.Spec.ComponentRoutes {
+		if componentRoute.Name == componentName && componentRoute.Namespace == api.OpenShiftConsoleNamespace {
+			componentRoute := ingressConfig.Spec.ComponentRoutes[i].DeepCopy()
+			return componentRoute
```
awgreene, line 55:
```suggestion
			return ingressConfig.Spec.ComponentRoutes[i].DeepCopy()
```

pkg/console/subresource/route/route.go
```diff
@@ -35,6 +36,137 @@ type CustomTLSCert struct {
...
+		}
+	}
+	return nil
+}
+
+func getComponentRouteStatus(ingressConfig *configv1.Ingress, componentName string) *configv1.ComponentRouteStatus {
+	for i, componentRoute := range ingressConfig.Status.ComponentRoutes {
+		if componentRoute.Name == componentName && componentRoute.Namespace == api.OpenShiftConsoleNamespace {
+			componentRoute := ingressConfig.Status.ComponentRoutes[i].DeepCopy()
+			return componentRoute
```
awgreene, line 65:
```suggestion
return ingressConfig.Status.ComponentRoutes[i].DeepCopy()
```

pkg/console/subresource/route/route.go
```diff
@@ -35,6 +36,137 @@ type CustomTLSCert struct {
...
+	defaultRoute := RouteControllerSpec{
+		hostname: GetDefaultRouteHost(routeName, ingressConfig),
+	}
+	var customRoute RouteControllerSpec
+	var isIngressConfigCustomHostnameSet bool
+
+	// Custom hostname in ingress config takes precedent over console operator's config
+	componentRouteSpec := getComponentRouteSpec(ingressConfig, routeName)
+	if componentRouteSpec != nil {
+		customRoute.hostname = string(componentRouteSpec.Hostname)
```
awgreene, line 81: You may want to validate the hostname.

pkg/console/subresource/route/route.go
```diff
@@ -35,6 +36,137 @@ type CustomTLSCert struct {
...
+		if len(componentRouteSpec.ServingCertKeyPairSecret.Name) != 0 {
```
awgreene, line 82:
```suggestion
		if componentRouteSpec.ServingCertKeyPairSecret.Name != "" {
```

pkg/console/subresource/route/route_test.go
```diff
@@ -22,290 +20,6 @@ const (
 	tlsCertificate = "certificate"
 )
 
-func TestDefaultRoute(t *testing.T) {
```
awgreene, line 25: Could you explain the decision to remove these tests rather than update them?

pkg/console/controllers/route/controller_test.go
awgreene: It seems like the functionality may be checked in [these e2e tests](https://github.com/openshift/console-operator/pull/522/files#diff-5da9aaf9399e87599aa868fc1e2e371845bdde346e5642913e4c3b212d8fda5fR47), but I appreciate fail fast unit tests.

test/e2e/custom_url_test.go
awgreene: +1

awgreene (changes requested):
Great work @jhadvig, I have a few nits and 2 questions:

-  Does the console operator write to the status of the ingresses.config.openshift.io/cluster resource? Admins should be able to inspect the status of the ingress config resource to notify admins that the route is configurable: https://github.com/openshift/enhancements/blob/master/enhancements/ingress/custom-route-configuration.md#user-stories
- I believe there is still work to be done so Console trusts the configured secret for the oauth-openshift service (this can be done later  but should be done at some point),

bindata/routes/downloads-route.yaml
awgreene: I had not noticed that this file was moved from the manifests dir. CVO is no longer managing this resource.

pkg/console/controllers/route/controller_test.go
jhadvig: Moved the function to a different pkg and forgot to move the tests... will update!

pkg/console/subresource/route/route_test.go
jhadvig: since we moved from assembling the route manifest via function, to using the bindata I thought that it doesnt make sense to have those tests anymore.

pkg/console/subresource/route/route.go
jhadvig: All the validation hostname + secret is done in the route controller. This is just setting an internal config.

jhadvig:
> Does the console operator write to the status of the ingresses.config.openshift.io/cluster resource? Admins should be able to inspect the status of the ingress config resource to notify admins that the route is configurable.

Current implementation is not writing the status of the ingresses.config.openshift.io/cluster resource. Haven't found any helpers for setting/editing the `ComponentRouteStatus` in the library-go, so thought that we are still missing that functionality.

jhadvig:
@sferich888  @awgreene Addressed comments + answered questions.
PTAL

bindata/routes/console-custom-route.yaml
```diff
@@ -0,0 +1,23 @@
...
+
```
awgreene, line 23: Weird - GitHub is still complaining about a missing newline.

manifests/03-rbac-role-cluster.yaml
```diff
@@ -35,6 +35,17 @@ rules:
...
+  - config.openshift.io
+  resources:
+  - ingresses/status
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
+  - delete
```
awgreene, line 48: You should get this for free by adding a consuming user in the `status.componentroute`, can be addressed when the controller is updated to write to status.

spadgett (approved):
/lgtm

awgreene (commented):
/lgtm

PR #523, opened 2021-04-08
Add management workload annotations

cynepco3hahue:
In support of the workload partitioning feature
(openshift/enhancements#703), we need to add
annotations to all management pods and namespaces so they can be
properly identified and assigned to segregated management cores on
clusters configured to do so.

Signed-off-by: Artyom Lukianov <alukiano@redhat.com>

### DIFF
--- a/manifests/02-namespace.yaml
+++ b/manifests/02-namespace.yaml
@@ -7,6 +7,7 @@ metadata:
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
     openshift.io/node-selector: ""
+    workload.openshift.io/allowed: "management"
 ---
 apiVersion: v1
 kind: Namespace
@@ -17,6 +18,7 @@ metadata:
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
     openshift.io/node-selector: ""
+    workload.openshift.io/allowed: "management"
   labels:
     openshift.io/cluster-monitoring: "true"
 ---
@@ -29,3 +31,4 @@ metadata:
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
     openshift.io/node-selector: ""
+    workload.openshift.io/allowed: "management"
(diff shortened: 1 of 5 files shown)


pkg/console/subresource/deployment/deployment.go
```diff
@@ -74,6 +76,7 @@ func DefaultDeployment(operatorConfig *operatorv1.Console, cm *corev1.ConfigMap,
 		infrastructureConfigResourceVersionAnnotation:        infrastructureConfig.GetResourceVersion(),
 		secretResourceVersionAnnotation:                      sec.GetResourceVersion(),
 		consoleImageAnnotation:                               util.GetImageEnv("CONSOLE_IMAGE"),
+		workloadManagementAnnotation:                         workloadManagementAnnotationValue,
```
jhadvig, line 79: This will set the annotation on the `Deployment`. From the enhancement I was under impression that the `workload.openshift.io/management` should be set only on pods. Is that right ?

jhadvig (commented):
@cynepco3hahue thanks for PR. Added couple of questions before we can merge.

manifests/02-namespace.yaml
```diff
@@ -29,3 +31,4 @@ metadata:
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
     openshift.io/node-selector: ""
+    workload.openshift.io/allowed: "management"
```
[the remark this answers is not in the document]
dhellmann: We're going to add a CI job to require all of our namespaces to have the annotation, and it would be easier to add the value here than track the special case in the CI test job definition.

Including the annotation won't trigger any extra behaviors if there are never any pods in the namespace.

pkg/console/subresource/deployment/deployment.go
dhellmann: You are correct that the machinery managing the workloads will only look at the pod. There's a comment on line 81 about tracking annotations on the deployment in case they change. The same annotation struct created on line 70 and used on line 83 is also used in the pod spec on line 107. Do we need the 2 sets of annotations to be the same in order to manage the update in the way that comment on line 81 implies?

manifests/02-namespace.yaml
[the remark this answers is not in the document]
jhadvig: Thanks for the clearing things up!

pkg/console/subresource/deployment/deployment.go
jhadvig: > Do we need the 2 sets of annotations to be the same in order to manage the update in the way that comment on line 81 implies?

I would rather have two sets of annotations. One `deploymentAnnotations` for deployment and second `podsAnnotations` for it's pods, where the pods annotation would include also the new annotation.

cynepco3hahue: done

jhadvig (approved):
/lgtm

PR #528, opened 2021-04-14
Export quickStarts data to console-config ConfigMap

sahil143:
fixes: https://issues.redhat.com/browse/ODC-5566

Export Quick start disabled data to console config map

Based on the enhancement proposal https://github.com/openshift/enhancements/pull/715

### DIFF
--- a/pkg/console/subresource/consoleserver/config_builder_test.go
+++ b/pkg/console/subresource/consoleserver/config_builder_test.go
@@ -700,6 +700,37 @@ plugins:
   plugin2: plugin2_url
 `,
 		},
+		{
+			name: "Config builder should handle inputs for quick starts options",
+			input: func() ([]byte, error) {
+				b := &ConsoleServerCLIConfigBuilder{}
+				b.QuickStarts(v1.QuickStarts{
+					Disabled: []string{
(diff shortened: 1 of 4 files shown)


jerolimov (approved):
Also if I have no lgtm persmissions here, this looks good to me. :+1: 

I verified this on a cluster (`launch openshift/console-operator#528`) and the data was copied successfully from the `cluster` `Console` to the `console-config` `ConfigMap`.

pkg/console/subresource/consoleserver/config_builder_test.go
```diff
@@ -321,6 +321,39 @@ func TestConsoleServerCLIConfigBuilder(t *testing.T) {
 				Providers: Providers{},
 			},
 		},
+		{
+			name: "Config builder should handle quick starts options",
```
jerolimov, line 325: @sahil143 this works fine, just one idea: there is another test section below which verifies the input against some YAML output. If you like you can add another test there because the final YAML is what the console bridge could parse later. So this verifies the contract between both projects a little better.

jhadvig (approved):
/lgtm
/approve

PR #531, opened 2021-04-20
Bug 1948524: Update operator's status with downloads deployment generation && pull route health check into a standalone controller

jhadvig:
Remove the `ResyncEvery(time.Minute)` from most of the controllers, so unnecessary sync loops are not triggered. Instead I've created a `HealthCheckController` that should contain all the health check that we have (currently only for `console` route), and will add in the future. This controller sync is triggered periodically, every minute.

/assign @spadgett

### DIFF
--- a/pkg/console/controllers/clidownloads/controller.go
+++ b/pkg/console/controllers/clidownloads/controller.go
@@ -4,7 +4,6 @@ import (
 	// standard lib
 	"context"
 	"fmt"
-	"time"
 
 	// kube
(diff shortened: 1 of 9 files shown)


spadgett (approved):
/lgtm

spadgett (approved):
/lgtm

spadgett (approved):
/lgtm

spadgett (approved):
/lgtm

spadgett (approved):
/lgtm

PR #534, opened 2021-04-21
update to non-deprecated NewCommand method

pweil-:
Update to non-deprecated cmd method with TODO context.

```go
// NewCommand returns a new command that a caller must set the Use and Descriptions on.  It wires default log, profiling,
// leader election and other "normal" behaviors.
// Deprecated: Use the NewCommandWithContext instead, this is here to be less disturbing for existing usages.
func (c *ControllerCommandConfig) NewCommand() *cobra.Command {
	return c.NewCommandWithContext(context.TODO())

}
```

### DIFF
--- a/pkg/cmd/operator/cmd.go
+++ b/pkg/cmd/operator/cmd.go
@@ -1,6 +1,8 @@
 package operator
 
 import (
+	// golang
+	"context"
 	// 3rd party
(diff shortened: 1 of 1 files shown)


jhadvig (approved):
/lgtm
/approve

PR #535, opened 2021-04-23
Bug 1947794: Pull HelmChartRepository CRD v1

pedjak:
Up to now HelmChartRepository was defined using CRD v1beta1 api, which is going to be deprecated in OCP 4.8


jhadvig (approved):
/lgtm
/approve
/retest

PR #543, opened 2021-05-12
CONSOLE-2768: Update the console public configmap manifests to use bindata

florkbr:
First of a few PRs to update the console public configmap to be pulled from bindata.

https://issues.redhat.com/browse/CONSOLE-2768


bindata/configmaps/console-public-configmap.yaml
```diff
@@ -1,3 +1,5 @@
+# This configmap 'console-public' manifest is used to expose the console URL
+# to all authenticated users
```
jhadvig, line 2: Since we are moving this manifest from the `/manifest` (which would be created the CVO) and instead will be created and handled by console-operator, we should remove all the `include.release.openshift.io` and `release.openshift.io` annotations.

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -743,10 +743,19 @@ func TestDefaultPublicConfigMap(t *testing.T) {
...
 			want: &corev1.ConfigMap{
-				TypeMeta: metav1.TypeMeta{},
+				TypeMeta: metav1.TypeMeta{
+					Kind:       "ConfigMap",
+					APIVersion: "v1",
+				},
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      api.OpenShiftConsolePublicConfigMapName,
 					Namespace: api.OpenShiftConfigManagedNamespace,
+					Annotations: map[string]string{
```
jhadvig, line 753: We can drop these, since we are moving the manifest to be handled by console-operator

jhadvig (changes requested):
@florkbr thanks for the PR! We can remove the annotations from the CM now that console-operator is responsible for it.

jhadvig (approved):
/lgtm
/approve

PR #544, opened 2021-05-13
Bug 1959696: Add warning log message when custom domain is configured on console-operator config

jhadvig:
If the console-operator config is being configured with custom domain for the `console` route, warn user that this type of configuration is deprecated and navigate him where the configuration should be moved and how does it should look like.

/assign @spadgett

cc'ing @TheRealJon @florkbr


spadgett (approved):
/lgtm

PR #546, opened 2021-05-13
Bug 1960324: quickstarts: remove "spec.version"

vrutkovs:
This key is not declared in CRD, so it makes CVO re-apply this manifest on every sync cycle


spadgett (approved):
/lgtm

Thank you

jhadvig (approved):
/lgtm

PR #548, opened 2021-05-20
Bug 1962698: Console-operator can not create resource console-public configmap in the openshift-config-managed namespace

jhadvig:
console-operator doesnt have RBAc to create `console-public` CM in the `openshift-config-managed` NS, thats why we had it in the `/manifest` folder, so CVO would create it.

Just FYI, sin the `resourceapply.ApplyConfigMap()` is merging labels and annotations we dont need to have the `release.openshift.io` have in the bindata of the `console-public` CM. 

/assigne @spadgett 

@florkbr fyi

### DIFF
--- a/manifests/03-rbac-role-ns-openshift-config-managed.yaml
+++ b/manifests/03-rbac-role-ns-openshift-config-managed.yaml
@@ -15,6 +15,10 @@ rules:
   verbs:
   - get
   - list
+  # We cannot restrict create or deletecollection requests by resourceName.
+  # For create, this limitation is because the object name is not known at authorization time.
+  # Check: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources
+  - create
   - watch
 - apiGroups:
   - ""
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -58,16 +58,15 @@ func (co *consoleOperator) sync_v400(ctx context.Context, controllerContext fact
 		routeName = api.OpenshiftConsoleCustomRouteName
 	}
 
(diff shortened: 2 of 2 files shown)


jhadvig:
@spadgett other option is to add `create` to the RBAC

spadgett:
@jhadvig I would probably just add `create` to the RBAC if we want to use bindata for the configmap.

We need to determine why the operator didn't go to degraded when the create failed and fix that issue as well.

deads2k:
/hold

please address the fact that the operator fails to go degraded when it cannot accomplish its goals before fixing the permissions issue.

jhadvig:
@spadgett I've addressed @deads2k comment. We have been missing setting of the Degraded conditions when updating  console config and also console-public CM.
PTAL

spadgett:
> @spadgett I've addressed @deads2k comment. We have been missing setting of the Degraded conditions when updating console config and also console-public CM.

@jhadvig Thanks, can you check the error handling elsewhere to make sure this was just a one-off and not a systemic problem? Also I believe @deads2k is requesting that we have a PR without the `create` RBAC change to see the operator go degraded and fail CI.

spadgett:
Looks good, the operator now goes degraded when commenting out the RBAC change 👍 

> Operator degraded (ConsolePublicConfigMap_FailedApply): ConsolePublicConfigMapDegraded: configmaps is forbidden: User "system:serviceaccount:openshift-console-operator:console-operator" cannot create resource "configmaps" in API group "" in the namespace "openshift-config-managed"

jhadvig:
@spadgett removed additional commit which was commenting out the `create` verb for the console-operator RBAC.

spadgett (approved):
/lgtm
/hold cancel

PR #551, opened 2021-06-01
Bug 1966480: Console-operator's controllers are passed resourceSyncer which is not used

jhadvig:
Removing all the unnecessary resourceSyncer instances passed to operator's controllers.
Originally this refactoring PR should be bigger, but since we are too close to release I've decide to address other naming and structural changes in 4.9

/assign @spadgett @florkbr


spadgett (approved):
/lgtm

PR #552, opened 2021-06-07
Bug 1952405: Console operator should report Available:False when it's route is not accessible

jhadvig:
We haven't been reporting `Available: False` when the console route was not accessible.

/assign @spadgett


spadgett:
Code change looks good, but you have a typo in the commit message (operator is misspelled).

jhadvig:
@spadgett comment addressed :)
PTAL

spadgett (approved):
/lgtm

PR #553, opened 2021-06-10
Bug 1954148: Implement console-operator updating Ingress operator config status with ComponentRoutes conditions

jhadvig:
Addressing per discussion with @awgreene, PTAL

FYI some methods have been borrowed from https://github.com/openshift/cluster-authentication-operator/pull/430/files, since the helpers for updating the status are still not available in library-go.

/assign @spadgett

### DIFF
--- a/manifests/03-rbac-role-cluster.yaml
+++ b/manifests/03-rbac-role-cluster.yaml
@@ -38,6 +38,16 @@ rules:
 - apiGroups:
   - config.openshift.io
   resources:
+  - ingresses/status
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
+- apiGroups:
+  - config.openshift.io
+  resources:
   - proxies
   verbs:
   - get
--- a/pkg/console/controllers/route/controller.go
+++ b/pkg/console/controllers/route/controller.go
@@ -12,13 +12,16 @@ import (
 
 	// k8s
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	coreinformersv1 "k8s.io/client-go/informers/core/v1"
 	coreclientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 
 	// openshift
+	configv1 "github.com/openshift/api/config/v1"
 	operatorsv1 "github.com/openshift/api/operator/v1"
 	routev1 "github.com/openshift/api/route/v1"
 	configclientv1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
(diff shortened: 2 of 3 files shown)


manifests/03-rbac-role-cluster.yaml
```diff
@@ -35,6 +35,16 @@ rules:
...
   - watch
+- apiGroups:
+  - config.openshift.io
+  resources:
+  - ingresses/status
+  verbs:
+  - get
+  - list
+  - watch
+  - create
```
awgreene, line 46: I'm not convinced that this is needed but it's probably fine.

pkg/console/controllers/route/controller.go
```diff
@@ -159,6 +163,124 @@ func (c *RouteSyncController) Sync(ctx context.Context, controllerContext factor
...
+				Name:      c.routeName,
+				Namespace: api.OpenShiftConsoleNamespace},
+		},
+	}
+	conditions := checkErrorsConfiguringCustomRoute(customRouteErrors)
+
+	componentRoute.Conditions = ensureDefaultConditions(conditions)
+	_, err := c.updateComponentRouteStatus(ctx, componentRoute)
+
+	fmt.Printf("\nERROR - %q\n", err)
```
awgreene, line 188: It seems like you typically use `klog`.

pkg/console/controllers/route/controller.go
```diff
@@ -159,6 +163,124 @@ func (c *RouteSyncController) Sync(ctx context.Context, controllerContext factor
 	return statusHandler.FlushAndReturn(defaultRouteErr)
 }
 
+func (c *RouteSyncController) updateIngressConfigStatus(ctx context.Context, routeConfig *routesub.RouteConfig, customRouteErrors []error) error {
+	componentRoute := &configv1.ComponentRouteStatus{
+		Namespace:        api.OpenShiftConsoleNamespace,
+		Name:             c.routeName,
+		DefaultHostname:  configv1.Hostname(routeConfig.GetDefaultRouteHostname()),
+		CurrentHostnames: []configv1.Hostname{configv1.Hostname(routeConfig.GetCustomRouteHostname())},
```
awgreene, line 171: Seems like current hostname could be set to an empty string if [this defer statement](https://github.com/openshift/console-operator/pull/553/files#diff-af15b762c545564ccd44f8bf5c8592e4233e4836241f99b05bd73f8109c181daR323) is executed after the custom route fails validation.

pkg/console/controllers/route/controller.go
```diff
@@ -144,6 +147,7 @@ func (c *RouteSyncController) Sync(ctx context.Context, controllerContext factor
 	// route into inaccessible state.
 	_, customRouteErrReason, customRouteErr := c.SyncCustomRoute(ctx, routeConfig, controllerContext)
 	statusHandler.AddConditions(status.HandleProgressingOrDegraded("CustomRouteSync", customRouteErrReason, customRouteErr))
+	// if routeConfig.IsCustomHostnameSet()
```
awgreene, line 150:
```suggestion
```

awgreene (changes requested):
A few comments but it looks good otherwise.

yanpzhan:
@jhadvig Seems the pr is ready for testing. I checked on cluster launched against the pr, after add custom route for "downloads" , the new created custom route could be accessed. In the ingress config yaml, there is condition:
    - conditions:
        - lastTransitionTime: '2022-01-21T03:32:54Z'
          message: 'Error Configuring custom route: [<nil>]'
          reason: CustomRouteError
          status: 'True'
          type: Degraded
        - lastTransitionTime: '2022-01-21T03:32:54Z'
          message: 'Error Configuring custom route: [<nil>]'
          reason: CustomRouteError
          status: 'False'
          type: Progressing
Check the whole yaml in http://pastebin.test.redhat.com/1023063 , is the condition correct?

PR #557, opened 2021-06-16
Bug 1975714: Add policy-group label to the openshift-console namespace manifest

jhadvig:
Adding the `network.openshift.io/policy-group` label based.
@sleshchenko FYI

/assign @spadgett

### DIFF
--- a/manifests/02-namespace.yaml
+++ b/manifests/02-namespace.yaml
@@ -8,6 +8,8 @@ metadata:
     include.release.openshift.io/single-node-developer: "true"
     openshift.io/node-selector: ""
     workload.openshift.io/allowed: "management"
+  labels:
+    network.openshift.io/policy-group: "console"
 ---
 apiVersion: v1
 kind: Namespace


manifests/02-namespace.yaml
```diff
@@ -8,6 +8,7 @@ metadata:
     include.release.openshift.io/single-node-developer: "true"
     openshift.io/node-selector: ""
     workload.openshift.io/allowed: "management"
+    network.openshift.io/policy-group: "console"
```
sleshchenko, line 11: Well, I'm checking on mobile but AFAIS you add annotation but label is needed on our end

jhadvig:
@sleshchenko comment addressed. PTAL

sleshchenko (approved):
LGTM

sleshchenko:
Please also backport to OS 4.7, since it blocks installing WTO on DevSandbox https://issues.redhat.com/browse/CRT-938

spadgett (approved):
/lgtm
/hold
@jhadvig Can you attach this to a Bugzilla? Then we can remove the hold.

PR #550, opened 2021-05-25
CONSOLE-2768: Update default and download deployments to use bindata

florkbr:
Tech debt - migrates our inline manifests for the default console deployment and the download deployment to Go bindata while substituting in necessary variables. This will make it clearer where to modify the base deployments in the console operator (versus looking through Go code). See `bindata/deployments` in the diff.

https://issues.redhat.com/browse/CONSOLE-2768

PR #558, opened 2021-06-24
Bug 1975605: Schedule downloads pods on master nodes

jhadvig:
Adding the `"node-role.kubernetes.io/master": "",` nodeSelector for the downloads pods.

/assign @spadgett 

@florkbr FYI this will need to ported to your https://github.com/openshift/console-operator/pull/550 PR afterwards.


spadgett:
/hold

I don't think we should always force the downloads to the master nodes... We should look at what other operators are doing. I think the right fix is probably to make the node selector configurable.

jhadvig:
Closing since we are migrating this issue into a [jira story ](https://issues.redhat.com/browse/CONSOLE-2903?filter=-2)

PR #559, opened 2021-06-24
Adds ODF install tour

vbnrh:
For 4.9 release, OCS operator will be replaced with the ODF at a higher level. Hence we would want to make sure that new install quickstart accurately represent the workflows for ODF operator.

This commit adds quickstart for installing the ODF operator and setting up the storage cluster.

Signed-off-by: Vineet Badrinath <vbadrina@redhat.com>

### DIFF
--- a/quickstarts/install-odf.yaml
+++ b/quickstarts/install-odf.yaml
@@ -1,0 +1,92 @@
+apiVersion: console.openshift.io/v1
+kind: ConsoleQuickStart
+metadata:
+  name: odf-install-tour
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+spec:
+  displayName: Install the OpenShift Data Foundation
+  durationMinutes: 5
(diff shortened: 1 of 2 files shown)


quickstarts/install-odf.yaml
```diff
@@ -1,31 +1,31 @@
 apiVersion: console.openshift.io/v1
 kind: ConsoleQuickStart
 metadata:
-  name: ocs-install-tour
+  name: odf-install-tour
```
spadgett, line 4: I don't believe anything will delete the old `ocs-install-tour` quick start if you change the name. The simplest would be to keep the old `metadata.name` and just update the to reference ODF. (Only the display name should be shown in the UI.)

spadgett (changes requested):

spadgett:
@openshift/quickstart-ux-content

quickstarts/install-odf.yaml
vbnrh: Makes sense. Ideally I thought the operator may be reconciling to see which resources were available and deploying them.

quickstarts/install-odf.yaml
```diff
@@ -7,25 +7,25 @@ metadata:
...
   tasks:
-  - title: Install the OpenShift Container Storage operator
+  - title: Install the OpenShift Data Foundation
     description: |- 
       The OperatorHub is where you can find a catalog of available Operators to install on your cluster.
       
-      To install OpenShift Container Storage, follow these steps:
+      To install OpenShift Data Foundation, follow these steps:
 
       1. Enter the administrator perspective: In the main navigation, click the dropdown menu and select Administrator.
```
serenamarie125, line 22: I'd suggest using this, which will enable discovery of the perspective switcher.  @beanh66 i'm not sure if this wording works, so you may want to verify!

Enter the administrator perspective: In the main navigation, click the [perspective switcher]{{highlight qs-perspective-switcher}} and select Administrator.

quickstarts/install-odf.yaml
```diff
@@ -7,25 +7,25 @@ metadata:
...
 
       1. In the navigation menu, select **Operators > OperatorHub**
```
serenamarie125, line 24: I'd suggest using the following for discoverability of the nav section:

In the navigation menu, select [Operators]{{highlight qs-nav-operators}} > OperatorHub

quickstarts/install-odf.yaml
```diff
@@ -34,41 +34,40 @@ spec:
...
       failedTaskHelp: This task isn’t verified yet. Try the task again.
     summary:
-      success: Great work! You installed the OpenShift Container Storage operator.
+      success: Great work! You installed the OpenShift Data Foundation operator.
       failed: Try the steps again.
-  conclusion: Congratulations! The OpenShift Container Storage operator is ready to use. To learn how you can manage your storage space effectively, take the Getting Started With OpenShift Container Storage quick start.
+  conclusion: Congratulations! The OpenShift Data Foundation operator is ready to use. To learn how you can manage your storage space effectively, take the Getting Started With OpenShift Data Foundation quick start.
   nextQuickStart:
-    - "getting-started-ocs"
+    - "getting-started-odf"
```
serenamarie125, line 73: I don't see this quick start you are linking to.  Should it be part of this PR as well?

serenamarie125 (changes requested):
Requesting a few changes and have some questions.  Thanks for pinging me for a review!

I'd assume this should be called odf-install-tour ?

quickstarts/install-odf.yaml
christianvogt: Would it not be installed by the operator?

vbnrh: @serenamarie125 It would be installed and managed by the odf operator.

spadgett:
If we want to rename the resource, I believe we can now add a `release.openshift.io/delete: "true"` annotation to the old manifest so it's removed. We'd need to test that it's implemented. See https://github.com/openshift/enhancements/blob/ce4d303db807622687159eb9d3248285a003fabb/enhancements/update/object-removal-manifest-annotation.md

jhadvig (commented):
So I dont think that we can just rename the CVO managed resources. As @spadgett mentioned we should rather add the `release.openshift.io/delete: "true"` annotation to the old manifest so it's removed and create the new ODF install tour.

vbnrh:
@jhadvig @spadgett I have added the annotation to the older manifest and updated the PR. 

Please take a look

quickstarts/install-odf.yaml
vbnrh: Updated and added the proper perspective switchers

quickstarts/install-odf.yaml
vbnrh: I've added the annotation to the older manifest.

quickstarts/install-odf.yaml
vbnrh: Added

quickstarts/install-odf.yaml
vbnrh: Added

yuvalgalanti (approved):

spadgett (approved):
/lgtm

PR #560, opened 2021-06-24
Bug 1975379: Use hard requirement for anti-affinity rules on both console's deployments

jhadvig:
Use `requiredDuringSchedulingIgnoredDuringExecution` for anti-affinity rules on both console's deployments.
The label selectors for console deployment will be:
`"app": "console", "component": "ui"`
The label selectors for downloads deployment will be:
`"app": "console", "component": "downloads"`

/assign @spadgett

### DIFF
--- a/pkg/console/subresource/deployment/deployment.go
+++ b/pkg/console/subresource/deployment/deployment.go
@@ -257,15 +257,13 @@ func consolePodAffinity(infrastructureConfig *configv1.Infrastructure) *corev1.A
 	}
 	return &corev1.Affinity{
 		PodAntiAffinity: &corev1.PodAntiAffinity{
-			PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{{
-				Weight: 100,
-				PodAffinityTerm: corev1.PodAffinityTerm{
-					LabelSelector: &metav1.LabelSelector{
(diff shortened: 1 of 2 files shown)


spadgett (approved):
/lgtm

spadgett:
Hm, looks like this prevents console rollout:

```yaml
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2021-06-24T16:59:02Z"
    message: '0/6 nodes are available: 3 node(s) didn''t match Pod''s node affinity/selector,
      3 node(s) didn''t match pod affinity/anti-affinity rules, 3 node(s) didn''t
      match pod anti-affinity rules.'
    reason: Unschedulable
    status: "False"
    type: PodScheduled
  phase: Pending
```

/lgtm cancel

spadgett:
I have a feeling the problem is with rolling deployments since we'll have more than the normal number of replicas. If there are ever more than 3, they'll fail to schedule. We probably need to look at max surge and max unavailable if we want to make this a hard requirement.

jhadvig:
@spadgett I've added additional commit that update podAntiAffinity rules and explicitly specifies deployment strategy and its `maxSurge` and `maxUnavailability` 🤞

spadgett (approved):
/lgtm

PR #564, opened 2021-07-14
fix: Update the Node.js quickstart example repo.

lholmquist:
This updates the example repo for the Node.js quickstart.  The current repo has been archived and moved.  This PR points to the correct repo that we, the nodeshift, are still maintaining


spadgett (approved):
/lgtm
/ok-to-test

jhadvig (approved):
/retest

PR #565, opened 2021-07-15
Bug 1980531: Add release.openshift.io/delete annotation to consoleLink CRDs

jhadvig:


spadgett:
Looks like we'll need to update the console e2e tests which expect the items we're removing to be there. cc @kdoberst

kdoberst:
> Looks like we'll need to update the console e2e tests which expect the items we're removing to be there. cc @kdoberst

@spadgett I'm assuming the console e2e tests are tests that cover both the console-operator and the console code via the browser.   With the above changes, the tests are going to fail because the links are missing.  But once the code for https://github.com/openshift/console/pull/9495 is merged in, the links will display because the console code will be providing the links and the tests (in theory) will start passing.  

The changes proposed in https://github.com/openshift/console/pull/9495 aren't dependent on these console-operator changes.  There is code to ignore these two specific help links if they happen to be passed from the console-operator.  So here is what I'm thinking ... once https://github.com/openshift/console/pull/9495 is merged in, re-run the tests for this PR and see if they all pass.  Thoughts?

jhadvig:
> once openshift/console#9495 is merged in, re-run the tests for this PR and see if they all pass. Thoughts?

👍

spadgett:
> The changes proposed in [openshift/console#9495](https://github.com/openshift/console/pull/9495) aren't dependent on these console-operator changes. There is code to ignore these two specific help links if they happen to be passed from the console-operator. So here is what I'm thinking ... once [openshift/console#9495](https://github.com/openshift/console/pull/9495) is merged in, re-run the tests for this PR and see if they all pass. Thoughts?

My only hesitation is that the checks to ignore the help links won't be needed with this change, so we'd want a follow up PR to remove those. It's going to be 3 PRs either way. It might be simpler to

1. Update the console tests in one PR
2. Merge the console-operator PR that removes the ConsoleLink resources
3. Merge the console PR that adds the links back

jhadvig:
> 2. Merge the console-operator PR that removes the ConsoleLink resources

Not 100% sure if CVO will remove the ConsoleLink CRs on update. But I thing that the annotation is for the upgrade case, so the when updating cluster from 4.8. to 4.9. it will remove the manifest.

kdoberst:
@jhadvig @spadgett My concern is that we don't set up a situation where the tests start failing for other PRs and that we don't accidentally set up a situation where the links are showing up as duplicates (once from the operator and once from the new console code).   So how about this approach:

1.  Close the current PR for adding the translated links in the console code.  A new PR is created to modify the console to not show any help links coming from the console.  In this same PR the tests are modified to verify that there are no help links.
2. Merge the changes for this PR.  
3.  Put in the code to add the translation link into the console code, add that ability for the operator to add links and finally modify the tests to verify there are the help links without duplicates.  It is at this point that we verify the manifests were removed.

The down side is that there is an issue getting step 3 in, then there is a risk that 4.9 will be released without the help links or having to reverse steps 1 and 2.

jhadvig:
@kdoberst I think that if we go with the actions as @spadgett suggested in his [comment](https://github.com/openshift/console-operator/pull/565#issuecomment-881450811) we should be OK :)

kdoberst:
Opened a new PR and modified an existing PR to complete the fix for Bug 1980531.

This PR should be merged in after https://github.com/openshift/console/pull/9702 which removes a test and allow tests on this PR to pass.

As soon as this PR is merged, then https://github.com/openshift/console/pull/9495 should be merged.

spadgett (approved):
/lgtm

PR #566, opened 2021-07-15
Bug 1975379: Only use `kubernetes.io/hostname` for workload anti-affinity

jhadvig:


mandre (changes requested):
This fixes deployment on OpenStack. Please fix the unit tests and let's merge ASAP.

spadgett (approved):
Code LGTM, but we need to fix the unit tests

spadgett (approved):
/lgtm

mdbooth (commented):
I don't want it to hold up this patch, but the use of `timezone` in the subject confused me! It's an availability zone/failure domain.

spadgett:
Yeah I noticed that too. It really should be zone not timezone.

mdbooth:
I would suggest that we should actually entirely remove topology from the scheduling requirements here. From the original bug report:

```
It seems like console pods sometimes are scheduled to run on the same master node.
This cause issues for our customers by not be able to reach console during certain times while we patch our cluster.
How can we force console pods to not run on same master nodes?
```

The only requirement here is that the console is not impacted during cluster upgrades, which is covered by `kubernetes.io/hostname`. The `topology.kubernetes.io` labels are intended for user workloads where you want to, e.g. spread critical workloads across your cloud provider's failure domain (physical host/rack/dc; it's actually not well defined what they mean). For the OpenShift control plane we do this by scheduling the control plane VMs appropriately, so individual control plane services shouldn't be second guessing that.

My immediate priority, though, is to un-break the OpenStack jobs, so if that's landing this patch then land away! However, I recommend you remove this entirely. We also need to get our CI voting here!

spadgett:
/hold

It's late for @jhadvig, but I can push an update to change the commit message and remove the `topology.kubernetes.io` labels if you are able to review @mdbooth @mandre

spadgett:
Updated, @mdbooth PTAL

mandre:
Thanks for updating the patch. I personally didn't have major concerns for the soft anti-affinity for availability zones, however the commit message bothered me a bit. Glad it's now fixed. Let's unblock openstack CI.
/lgtm
/retest

jhadvig (commented):
@spadgett thanks for updating the PR 👍

mdbooth (approved):
/lgtm

PR #569, opened 2021-07-19
Support External control plane topology

csrwng:
In an external control plane topology (ROKS, Hypershift), there are no
master nodes. Deployments created by this operator should not include a
masters node selector.


jhadvig:
The PR looks good. @csrwng I've [sent you PR with commit](https://github.com/csrwng/console-operator/pull/1) thats adding unit test coverage. PTAL

/retest

jhadvig (approved):
/lgtm
/approve

PR #570, opened 2021-07-30
Bug 1988401: Add console RBAC for listing operators

jhadvig:
We need to give `console` SA additional rights for listing operators, check https://github.com/openshift/console/pull/9142#issuecomment-889340172

Story: https://issues.redhat.com/browse/CONSOLE-2458

/assign @spadgett


jhadvig:
Closing per discussion with @spadgett

PR #573, opened 2021-08-12
Bug 1986129: Bump openshift/api to add missing 'include.release.openshift.io/single-node-developer' annotation to the ConsolePlugin CRD

jhadvig:
/assign @spadgett


jhadvig:
@spadgett this one will need to by backported manually, since the commit hash will be different for 4.9 and 4.8

spadgett (approved):
/lgtm

PR #576, opened 2021-08-20
CONSOLE-2837: Enable plugins by default when migrating from static to dynamic

jhadvig:
Adding functionality for enable plugins by default when migrating from static to dynamic.
When operator is in `Managed` state it will check for available `ConsolePlugins` CRs on the cluster. if any of the available ConsolePlugin CR's name is not present in the console-operator configs `console.openshift.io/migrated-plugins` annotation array, it will be enabled by the operator and added to the `console.openshift.io/migrated-plugins` annotation array.
When the operator is in `Removed` state it will remove all the ConsolePlugins from its `Spec.Plugins` field and the `console.openshift.io/migrated-plugins` annotation will be set to empty array, so when the operator is back to the `Managed` state all the available ConsolePlugins will be enabled.

/assign @TheRealJon @spadgett

### DIFF
--- a/pkg/console/controllers/plugins/controller.go
+++ b/pkg/console/controllers/plugins/controller.go
@@ -26,7 +26,7 @@ import (
 )
 
 const (
-	migratedPluginsAnnotation = "console.openshift.io/migrated-plugins"
+	MigratedPluginsAnnotation = "console.openshift.io/migrated-plugins"
 )
 
 type PluginsMigrationController struct {
@@ -125,7 +125,7 @@ func (c *PluginsMigrationController) GetAvailablePluginsName() ([]string, error)
 }
 
 func GetMigratedPlugins(operatorConfig *operatorv1.Console) ([]string, error) {
-	migratedPluginsAnnotation, migratedPluginsAnnotationExists := operatorConfig.Annotations[migratedPluginsAnnotation]
+	migratedPluginsAnnotation, migratedPluginsAnnotationExists := operatorConfig.Annotations[MigratedPluginsAnnotation]
 	migratedPluginsArray := []string{}
 	if migratedPluginsAnnotationExists {
 		err := json.Unmarshal([]byte(migratedPluginsAnnotation), &migratedPluginsArray)
@@ -170,7 +170,7 @@ func (c *PluginsMigrationController) UpdateOperatorConfig(
 	if err != nil {
 		return err
 	}
-	updatedOperatorConfig.Annotations[migratedPluginsAnnotation] = string(marshaledPluginsToAnnotate)
+	updatedOperatorConfig.Annotations[MigratedPluginsAnnotation] = string(marshaledPluginsToAnnotate)
 
 	_, err = c.operatorConfigClient.Update(ctx, updatedOperatorConfig, metav1.UpdateOptions{})
 	return err
@@ -183,7 +183,7 @@ func (c *PluginsMigrationController) removeMigratedPlugins(ctx context.Context, 
 	if err != nil {
 		return err
 	}
-	updatedOperatorConfig.Annotations[migratedPluginsAnnotation] = string(marshaledPluginsToAnnotate)
+	updatedOperatorConfig.Annotations[MigratedPluginsAnnotation] = string(marshaledPluginsToAnnotate)
 	_, err = c.operatorConfigClient.Update(ctx, updatedOperatorConfig, metav1.UpdateOptions{})
 	return nil
 }
--- a/pkg/console/controllers/plugins/controller.go
+++ b/pkg/console/controllers/plugins/controller.go
@@ -1,0 +1,189 @@
+package plugins
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	// k8s
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/klog/v2"
+
+	// openshift
+	operatorv1 "github.com/openshift/api/operator/v1"
+	consoleinformersv1alpha1 "github.com/openshift/client-go/console/informers/externalversions/console/v1alpha1"
+	listerv1alpha1 "github.com/openshift/client-go/console/listers/console/v1alpha1"
+	operatorclientv1 "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+
+	// console-operator
+	"github.com/openshift/console-operator/pkg/api"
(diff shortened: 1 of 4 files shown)


pkg/console/controllers/plugins/controller.go
```diff
@@ -0,0 +1,189 @@
...
+
+	switch operatorConfig.Spec.ManagementState {
+	case operatorv1.Managed:
+		klog.V(4).Infof("console-operator is in a managed state: migrating available plugins")
+	case operatorv1.Unmanaged:
+		klog.V(4).Infof("console-operator is in an unmanaged state: skipping available plugins migration")
+		return nil
+	case operatorv1.Removed:
+		klog.V(4).Infof("console-operator is in a removed state: removing migrated plugins")
+		return c.removeMigratedPlugins(ctx, operatorConfig)
```
spadgett, line 82: It's unclear to me why we're clearing enabled plugins when we put the console into removed state. I don't think we'd want to change any of the console config, just stop the console. If I set the console to removed then back to managed, I'd expect it to have the same config as before.

pkg/console/controllers/plugins/controller.go
```diff
@@ -0,0 +1,189 @@
...
+
+func GetPluginsToEnable(
+	operatorConfig *operatorv1.Console,
+	availablePluginsName []string,
+	migratedPluginsArray []string,
+) []string {
+	var pluginsToEnable []string
+	for _, availablePluginName := range availablePluginsName {
+		if !util.ContainsString(migratedPluginsArray, availablePluginName) && !util.ContainsString(operatorConfig.Spec.Plugins, availablePluginName) {
+			pluginsToEnable = append(pluginsToEnable, availablePluginName)
```
spadgett, line 148: We don't want to enable all plugins. We only want to enable ones from a well-defined allow list (e.g., `["kubevirt"]`)

pkg/console/controllers/plugins/controller.go
```diff
@@ -0,0 +1,189 @@
...
+	err = c.UpdateOperatorConfig(ctx, operatorConfig, pluginsToEnable, migratedPluginsArray)
+	statusHandler.AddConditions(status.HandleProgressingOrDegraded("AvailablePluginMigration", "FailedOperatorConfigUpdate", err))
+	if err != nil {
+		return statusHandler.FlushAndReturn(err)
+	}
+
+	return statusHandler.FlushAndReturn(err)
+}
+
+func (c *PluginsMigrationController) GetAvailablePluginsName() ([]string, error) {
```
spadgett, line 115:
```suggestion
func (c *PluginsMigrationController) GetAvailablePluginNames() ([]string, error) {
```

pkg/console/controllers/plugins/controller.go
```diff
@@ -0,0 +1,189 @@
...
+	availablePlugins, err := c.consolePluginLister.List(labels.Everything())
+	if err != nil {
+		return nil, err
+	}
+	var availablePluginsName []string
```
spadgett, line 120:
```suggestion
	var availablePluginNames []string
```

pkg/console/controllers/plugins/controller.go
```diff
@@ -0,0 +1,189 @@
...
+		return statusHandler.FlushAndReturn(err)
+	}
+
+	pluginsToEnable := GetPluginsToEnable(operatorConfig, availablePluginsName, migratedPluginsArray)
+
+	if len(pluginsToEnable) == 0 {
+		return statusHandler.FlushAndReturn(nil)
+	}
+
+	err = c.UpdateOperatorConfig(ctx, operatorConfig, pluginsToEnable, migratedPluginsArray)
```
spadgett, line 106: I think we want to add the well-known plugins to the migrated plugins array even if we didn't enable them. If kubevirt isn't already installed before the console upgrade, we don't want to automatically enable it when it's installed later. The use case we're trying to avoid is disabling the kubevirt UI when it was previously enabled in 4.8.

So

1. If `kubevirt` ConsolePlugin exists, automatically enable it.
2. Add `kubevrit` to the list of migrated plugins (even if the ConsolePlugin wasn't found).

I'm assuming here the kubevirt operator with the plugin must be updated before upgrading the cluster. We'll need to check with the kubevirt team. cc @glekner @vojtechszocs

pkg/console/controllers/plugins/controller.go
```diff
@@ -0,0 +1,189 @@
...
+	case operatorv1.Removed:
+		klog.V(4).Infof("console-operator is in a removed state: removing migrated plugins")
+		return c.removeMigratedPlugins(ctx, operatorConfig)
+	default:
+		return fmt.Errorf("unknown state: %v", operatorConfig.Spec.ManagementState)
+	}
+
+	statusHandler := status.NewStatusHandler(c.operatorClient)
+
+	availablePluginsName, err := c.GetAvailablePluginsName()
```
spadgett, line 89: Honestly, I think this should be drive off the installed operator rather than the availability of the plugin. Anyone who can create ConsolePlugin resources can create one with the name `kubevrit`. We want to check instead that the CNV operator is installed in `openshift-cnv`.

test/e2e/plugins_test.go
```diff
@@ -79,6 +67,53 @@ func TestAddPlugins(t *testing.T) {
 	}
 }
 
+func TestPluginsMigration(t *testing.T) {
+	expectedMigratedPluginsAnnotation := "['test-plugin','test-plugin-2']"
```
spadgett, line 71: Again, we only want to migrate plugins from an allow list, so we'd want to check that these plugins are *not* migrated.

spadgett (changes requested):
Thanks, @jhadvig. Note that we only want to automatically enable plugins in a well-defined allow list. We'll need to coordinate with the CNV team to learn the following:

1. Are we definitely moving kubevirt to a dynamic plugin in 4.9? (Otherwise we'd want to hold this PR.)
2. What is the plugin name for CNV?
3. Will older versions of the CNV operator without the plugin will block cluster upgrade to 4.9?

pkg/console/controllers/plugins/controller.go
jhadvig: > We want to check instead that the CNV operator is installed in openshift-cnv.

so wondering if we want check if the CNV's CSV is installed in the `openshift-cnv` together with the ConsolePlugin's name?

jhadvig:
@spadgett thanks for the review.
Wasnt aware that the `kubevirt` plugin should be the single plugin available in 4.9 (maybe).
If so I will create the a list that would containe:
- ConsolePlugin name 
- Namespace with a particular CSV name

eg.

```yaml
-  csv: 
      name: "kubevirt-hyperconverged-operator.v4.8.0"
      namespace: "openshift-cnv"
   pluginName: "kubevirt"
```

also not really sure in this case how we wanna handle CSV versioning ? or we dont ?

christianvogt:
I think we should be listing _all_ our static plugins today so that when a dynamic plugin gets installed to override the static UI, the dynamic plugin will automatically be enabled.

PR #577, opened 2021-08-27
OCS Quick Start should not be shown unless you have proper privileges

yozaam:
Fixes: https://issues.redhat.com/browse/ODC-6241

Added the described accessReviewResources inside OCS quickstart -spec

since this quick start should only be available if you can install operators


yozaam:
sorry, I have messed up something in git made a new PR https://github.com/openshift/console-operator/pull/578

PR #578, opened 2021-08-30
Bug 2000440: OCS Quick Start should not be shown unless you have proper privileges

yozaam:
Fixes: https://issues.redhat.com/browse/ODC-6241 ( relates to  https://issues.redhat.com/browse/ODC-6161 )

Added the described accessReviewResources inside OCS quickstart -spec

since this quick start should only be available if you can install operators

## doubt
I am seeing the operator install option in `KUBEADMIN` login
image

but not seeing the quickstart, is this expected from the given `accessReviewResources:` 
image


christianvogt:
Replying to @spadgett comment https://github.com/openshift/console-operator/pull/577#issuecomment-907453011

Why do they need the ability to create a namespace? They should also be able to install the operator into an existing namespace. Likely the admin would have this permission anyways though.

spadgett:
@christianvogt The quick start as written has you creating and installing into namespace `openshift-storage`. (I believe this is also required by the operator.)

spadgett:
(That is if the follow the steps and take the defaults in the form it will create the namespace.)

christianvogt:
I should have read the quick start before I commented :) Indeed if that's how the quick starts are written then this makes sense.

spadgett (approved):
/lgtm

PR #579, opened 2021-09-01
helm/metrics: add a new gauge metric to monitor helm releases

zonggen:
Adds a `helm_chart_release_health_status` gauge metric to monitor the
status of the current helm releases.

To query the total number of healthy releases: `count (helm_chart_release_health_status == 1) by (chartName, chartVersion, releaseName)`.
To query the total number of unhealthy releases: `count (helm_chart_release_health_status == 0) by (chartName, chartVersion, releaseName)`.

From console UI:
image

From Prometheus UI:
image

Closes: https://issues.redhat.com/browse/HELM-224
Signed-off-by: Allen Bai <abai@redhat.com>


manifests/03-rbac-role-cluster.yaml
```diff
@@ -80,6 +80,9 @@ rules:
   - create
   - update
   - delete
+- apiGroups: [""]
```
zonggen, line 83: This is required to run `helm list --all-namespaces` with Helm SDK. 

Error without it: `metric helm_chart_release_health_status unhandled: list: failed to list: secrets is forbidden: User "system:serviceaccount:openshift-console-operator:console-operator" cannot list resource "secrets" in API group "" at the cluster scope`

pkg/helm/metrics/metrics.go
```diff
@@ -0,0 +1,83 @@
...
+	} else {
+		listAction := action.NewList(actionConfig)
+		releases, err := listAction.Run()
+		if err != nil {
+			klog.Errorf("metric helm_chart_release_health_status unhandled: %v", err)
+			return
+		}
+
+		for _, release := range releases {
+			if status := release.Info.Status.String(); status == "deployed" {
```
dperaza4dustbit, line 44: So only deploy is healthy, what about the other states. I would say if it is unknown or failed then is 0 then is 1 for all other states

pkg/helm/metrics/metrics.go
```diff
@@ -0,0 +1,83 @@
...
+				klog.V(4).Infof("metric helm_chart_release_health_status 1: %s %s %s", release.Name, release.Chart.Metadata.Name, release.Chart.Metadata.Version)
+				helmChartReleaseHealthStatus.WithLabelValues(release.Name, release.Chart.Metadata.Name, release.Chart.Metadata.Version).Set(1)
```
dperaza4dustbit, line 46: the only diff between this and the line below is the 0 or 1. I would do healthStatus := 1 then change to 0 if status unknown or failed. Then make a single call to klog and to helmChartReleaseHealthStatus passing healthStatus

dperaza4dustbit (commented):
All those vendor files came from the helm sdk reference?

zonggen:
The vendor files came from after I did `go mod vendor` followed by `go mod tidy`.

pkg/helm/metrics/metrics.go
zonggen: Updated

zonggen:
@jhadvig @spadgett Gentle reminder: could you take a look at this PR?

jhadvig:
@zonggen I see that the e2e `TestMetricsEndpoint` TC is failing:

```
=== RUN   TestMetricsEndpoint
    ....
    metrics_test.go:71: searching for helm_chart_release_health_status in metrics data...
    metrics_test.go:74: did not find helm_chart_release_health_status
    metrics_test.go:47: waiting for cleanup to reach settled state...
--- FAIL: TestMetricsEndpoint (6.27s)

```

Should be fixed in the PR so it's passing.

zonggen:
@jhadvig Could you take a second look? The test case failed because the `helm_chart_release_health_status` was not initialized util there's a Helm release in the cluster. Added a logic to initialize it to 0 with no labels.

Reference: https://prometheus.io/docs/practices/instrumentation/#avoid-missing-metrics

pkg/helm/metrics/metrics.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+		klog.V(4).Infof("metric helm_chart_release_health_status %d: %s %s %s", healthStatus, release.Name, release.Chart.Metadata.Name, release.Chart.Metadata.Version)
+		helmChartReleaseHealthStatus.WithLabelValues(release.Name, release.Chart.Metadata.Name, release.Chart.Metadata.Version).Set(float64(healthStatus))
+	}
+}
+
+// Reference: https://github.com/helm/helm/issues/7430#issuecomment-620489002
+func getActionConfig() (*action.Configuration, error) {
+	actionConfig := new(action.Configuration)
+	var kubeConfig *genericclioptions.ConfigFlags
```
jhadvig, line 65: Dont think kubeconfig is the right name
```suggestion
	var configFlags *genericclioptions.ConfigFlags
```

jhadvig: also why do we need to initialize the configFlags, since on line in 72 you are setting it ? cant we jusst set it directly ?

jhadvig (changes requested):
@zonggen thanks for the PR. Adding couple of comments.

manifests/03-rbac-role-cluster.yaml
```diff
@@ -80,6 +80,14 @@ rules:
   - create
   - update
   - delete
+- apiGroups:
+  - ""
+  resources:
+  - secrets
```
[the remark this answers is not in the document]
zonggen: The `helm list --all-namespaces` action from `metrics.go` requires this particular permission. If I remove this it would produce: 
```
metrics.go:40] metric helm_chart_release_health_status unhandled: list: failed to list: secrets is forbidden: User "system:serviceaccount:openshift-console-operator:console-operator" cannot list resource "secrets" in API group "" at the cluster scope
```

pkg/console/operator/sync_v400.go
```diff
@@ -218,6 +220,7 @@ func (co *consoleOperator) GetActiveRouteInfo(ctx context.Context, activeRouteNa
 func (co *consoleOperator) SyncConsoleConfig(ctx context.Context, consoleConfig *configv1.Console, consoleURL string) (*configv1.Console, error) {
 	oldURL := consoleConfig.Status.ConsoleURL
 	metrics.HandleConsoleURL(oldURL, consoleURL)
+	helmmetrics.HandleHelmChartReleaseHealthStatus()
```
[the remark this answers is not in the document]
zonggen: Agreed, will move it to the main `sync_v400`.  The original intent was to piggyback on the `console_url` metric. Also willing to add a new controller but that controller will only consist of this one line `helmmetrics.HandleHelmChartReleaseHealthStatus()`.

pkg/helm/metrics/metrics.go
zonggen: Updated :)

pkg/helm/metrics/metrics.go
```diff
@@ -0,0 +1,90 @@
+package metrics
```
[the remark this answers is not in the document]
zonggen: Moved to `pkg/console/metrics/helm_metrics.go`

zonggen:
Thanks for reviewing! Waiting for the tests before another look.

zonggen:
@jhadvig PR is ready for another look but the `ci/prow/e2e-aws-operator` test kept failing for other non-related reasons.  Do I keep retesting until it succeeds? Thanks :)

manifests/03-rbac-role-cluster.yaml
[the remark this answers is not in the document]
spadgett: This is pretty concerning because console-operator is effectively running as root on the cluster with this change. Is there any way to add the metrics without needing this permission?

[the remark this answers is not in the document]
jwforres: Yeah this is a hard no. We won't escalate the console operator to have read on Secrets cluster-wide.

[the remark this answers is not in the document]
dperaza4dustbit: Understand, we are looking for alternative options to mitigate. It won't be a full solution if we have to use the users identity on the console side, but that is what we will pivot towards. For future reference @spadgett and @jwforres, If we had a way to filter Allow Rules by Resource Type like for example "type": "helm.sh/release.v1", would that work to reduce the access escalation? Trying to figure out if I propose a change in Kubernetes RBAC upstream on this.

PR #580, opened 2021-09-02
Bug 1994443: console-operator should report Available=true when at least available replica exists

jhadvig:
When setting console-operator Available condition it should be enough to check if there is at least one available replica.

/assign @spadgett


spadgett (approved):
/lgtm

PR #582, opened 2021-09-03
Bug 1989055: Revoke usage of Default Ingress Cert for console route healthcheck

jhadvig:
How route health check we need to use `DefaultIngressCert` not the `OAuthServingCertConfigMapName`.
This is related to failing health checks in the https://bugzilla.redhat.com/show_bug.cgi?id=1989055#c7

Also removing dead code from route controller.

/assign @stlaz @spadgett 

@florkbr FYI


spadgett (approved):
/lgtm

PR #584, opened 2021-09-06
Bug 2001523: Set Available condition to False when console-operator in Removed state

jhadvig:
We should report `Available` condition set to `False` when operator is in `Removed` state.

/assign @spadgett

### DIFF
--- a/pkg/console/operator/operator.go
+++ b/pkg/console/operator/operator.go
@@ -30,6 +30,7 @@ import (
 	routesinformersv1 "github.com/openshift/client-go/route/informers/externalversions/route/v1"
 	"github.com/openshift/console-operator/pkg/api"
 	"github.com/openshift/console-operator/pkg/console/controllers/util"
+	consolestatus "github.com/openshift/console-operator/pkg/console/status"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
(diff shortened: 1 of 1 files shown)


spadgett:
@wking Can you confirm this is the right behavior?

We should confirm setting "Available: false" doesn't block upgrades when console is removed.

wking:
> We should confirm setting "Available: false" doesn't block upgrades when console is removed.

Setting `Available: False` will absolutely [block updates][1], and fire off `critical` alerts, and all sorts of bad things.  Please don't do that ;).  openshift/api#1000 is my in-flight effort to clarify these condition semantics.  Applied to this case, if the user asked you to remove your operand, and you have successfully removed it, then you are `Available: True`, meaning "everything my component is configured to be doing, it's successfully doing".

[1]: https://github.com/openshift/enhancements/blob/98f33c569d9d5f86f91b12cfa8a1ce86ece72066/dev-guide/cluster-version-operator/user/reconciliation.md#ClusterOperator

spadgett:
> Setting `Available: False` will absolutely [block updates](https://github.com/openshift/enhancements/blob/98f33c569d9d5f86f91b12cfa8a1ce86ece72066/dev-guide/cluster-version-operator/user/reconciliation.md#ClusterOperator), and fire off `critical` alerts, and all sorts of bad things. Please don't do that ;). [openshift/api#1000](https://github.com/openshift/api/pull/1000) is my in-flight effort to clarify these condition semantics. Applied to this case, if the user asked you to remove your operand, and you have successfully removed it, then you are `Available: True`, meaning "everything my component is configured to be doing, it's successfully doing".

Thanks, @wking. Closing the PR.

/close

PR #585, opened 2021-09-06
Bug 2001240: Remove SimpleHTTP 'server' response header value

jhadvig:
This fix will set the `server` value to `""` since otherwise we would need to override the whole SimpleHTTP module. That would be an overkill due to that fact that we are inlining the server code in the containers spec.

response will look:
Screenshot 2021-09-06 at 16 14 18

/assign @spadgett 
@rludva FYI

### DIFF
--- a/pkg/console/subresource/deployment/deployment.go
+++ b/pkg/console/subresource/deployment/deployment.go
@@ -631,7 +631,10 @@ class Thread(threading.Thread):
 		self.start()
 
 	def run(self):
-		httpd = http.server.HTTPServer(addr, http.server.SimpleHTTPRequestHandler, False)
+		server = http.server.SimpleHTTPRequestHandler
+		server.server_version = "OpenShift Downloads Server"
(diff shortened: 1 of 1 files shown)


pkg/console/subresource/deployment/deployment.go
```diff
@@ -631,7 +631,10 @@ class Thread(threading.Thread):
 		self.start()
 
 	def run(self):
-		httpd = http.server.HTTPServer(addr, http.server.SimpleHTTPRequestHandler, False)
+		server = http.server.SimpleHTTPRequestHandler
+		server.server_version = ""
```
spadgett, line 635: nit: You might just make this something like "OpenShift Downloads Server" (without any specific version). This doesn't reveal any information about the server runtime, but avoids serving an empty response header.

jhadvig:

spadgett (approved):
/lgtm

PR #590, opened 2021-09-14
Bug 1999314: Resync all controllers periodically

jhadvig:
Resync of the controllers was removed in https://github.com/openshift/console-operator/pull/531. But since it looks like we need it Im adding them back.
Together with https://github.com/openshift/console-operator/pull/580 PR this issue should be solved. Wondering if a minute is the right period here.

/assign @spadgett 
@TheRealJon @florkbr FIY

### DIFF
--- a/pkg/console/controllers/clidownloads/controller.go
+++ b/pkg/console/controllers/clidownloads/controller.go
@@ -4,6 +4,7 @@ import (
 	// standard lib
 	"context"
 	"fmt"
+	"time"
 
 	// kube
 	"k8s.io/apimachinery/pkg/api/equality"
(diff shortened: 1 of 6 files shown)


spadgett:
@jhadvig Was the issue that originally caused us to rollback the resync change fixed?

https://bugzilla.redhat.com/show_bug.cgi?id=1948524

jhadvig:
@spadgett so the issue was with the downloads deployment which was being rolled out every minute. That was cause by missing, which was due to the fact that we have been missing calling method for [updating deployment generation](https://github.com/jhadvig/console-operator/blob/34940460b4721db3fa780d14e3ae26579f5dfa2a/pkg/console/controllers/downloadsdeployment/controller.go#L119) which we added. Unfortunately with that PR I've removed the resync methods call.

Tested the change and there should be no redeployments happening.

spadgett (approved):
/lgtm

PR #589, opened 2021-09-13
[release-4.8] Bug 2003639: Use kubernetes.io/hostname for workload anti-affi… …nity

jhadvig:
Manual backport of:
- https://github.com/openshift/console-operator/pull/560
- https://github.com/openshift/console-operator/pull/566

/assign @spadgett

PR #605, opened 2021-10-25
Bug 2016654: Use hard requirements for antiaffinity rules

jhadvig:
Manual backport of https://github.com/openshift/console-operator/pull/589, stripped of single node cluster work.

/assign @spadgett


yanpzhan:
Tested against the pr, console pods are scheduled on different master nodes. Test passed.

PR #607, opened 2021-11-01
Bug 2018380: Migrate docs links to access.redhat.com

yselkowitz:
In line with efforts currently under way for OCM, we need to update and migrate all docs.openshift.com links to access.redhat.com.

This needs to be considered in tandem with https://github.com/openshift/console/pull/10354

/assign @spadgett﻿


spadgett (approved):
/lgtm

PR #602, opened 2021-10-12
CONSOLE 2919: create OAuth clients on spoke clusters - rebased

florkbr:
Creates oauth clients on spoke clusters using the new managed cluster
controller via a dynamic client.

PR #612, opened 2021-11-23
Folllow on changes to CONSOLE-2919

TheRealJon:
Addresses my feedback to @florkbr 's PR #602 plus a little bit more polishing.

- Create managed cluster oauth cert configmap in openshift-console namespace
- Only set managed cluster config file on console config if the managed cluster config map is present
- Remove all multicluster related resources when no managed clusters exist or the API is not present.
- Simplify some variable and file names
- Get rid of a few magic strings and static values (moved to the api package)
- Update managed cluster controller to stop syncing if no local oauth client is found.
- Update `ManagedClusterController.getValidManagedClusters` to `ManagedClusterController.SyncManagedClusterList` for consistency
- Add new config map volumes and volume mounts for oauth certs to console operator deployment
- Update managed cluster controller to build managed cluster config with oauth data
- Add FeatureGate logic to managed cluster controller


kdoberst:
@TheRealJon It looks good to me, but I'm giving @florkbr a chance to take a look before I tag it.

pkg/console/controllers/managedclusters/controller.go
```diff
@@ -197,20 +213,143 @@ func (c *ManagedClusterController) ValidateManagedClusterClientConfigs(ctx conte
...
+				managedClusterListErrors = append(managedClusterListErrors, fmt.Sprintf("Unable to create oauth client for cluster %v: managed cluster view status unknown", namespace))
+			} else {
+				managedClusterList = append(managedClusterList, namespace)
+			}
+		}
+	}
+
+	errs := []string{}
+	mcas := []*unstructured.Unstructured{}
+	secretString := oauthsub.GetSecretString(localOAuthClient)
```
spadgett, line 264: Ideally we'd have different secrets for each OAuthClient, but OK for a tech preview.

pkg/console/controllers/managedclusters/controller.go
```diff
@@ -197,20 +213,143 @@ func (c *ManagedClusterController) ValidateManagedClusterClientConfigs(ctx conte
...
+	redirects := oauthsub.GetRedirectURIs(localOAuthClient)
+	for _, managedClusterName := range managedClusterList {
+		required := managedclusteractionsub.DefaultCreateOAuthClientAction(managedClusterName, secretString, redirects)
+		opt := metav1.CreateOptions{}
+		mca, err := c.dynamicClient.Resource(api.ManagedClusterActionGroupVersionResource).Namespace(managedClusterName).Create(ctx, required, opt)
+		if err != nil && apierrors.IsAlreadyExists(err) {
+			mca, err = c.dynamicClient.Resource(api.ManagedClusterActionGroupVersionResource).Namespace(managedClusterName).Get(ctx, required.GetName(), metav1.GetOptions{})
```
spadgett, line 271: What happens if the console public hostname changes after the managed cluster OAuth client is created? Is the redirect URL updated?

pkg/console/operator/sync_v400.go
```diff
@@ -241,19 +263,37 @@ func (co *consoleOperator) SyncDeployment(
...
-	}
-	requiredDeployment := deploymentsub.DefaultDeployment(operatorConfig, cm, apiServerCAConfigMaps, serviceCAConfigMap, oauthServingCertConfigMap, trustedCAConfigMap, sec, proxyConfig, infrastructureConfig, canMountCustomLogo, canMountManagedClusterConfig)
+	apiServerCertConfigMaps, apiServerCertConfigMapsErr := co.configMapClient.ConfigMaps(api.OpenShiftConsoleNamespace).List(ctx, metav1.ListOptions{LabelSelector: api.ManagedClusterAPIServerCertName})
+	if apiServerCertConfigMapsErr != nil {
+		klog.Warningf("Unable to list managed cluster API server cert ConfigMaps. Multicluster will not be enabled: %v", apiServerCertConfigMapsErr)
+	}
+
+	oAuthServerCertConfigMaps, oAuthServerCertConfigMapsErr := co.configMapClient.ConfigMaps(api.OpenShiftConsoleNamespace).List(ctx, metav1.ListOptions{LabelSelector: api.ManagedClusterOAuthServerCertName})
+	if apiServerCertConfigMapsErr != nil {
+		klog.Warningf("Unable to list managed cluster API server cert ConfigMaps. Multicluster will not be enabled: %v", oAuthServerCertConfigMapsErr)
```
spadgett, line 281:
```suggestion
		klog.Warningf("Unable to list managed cluster OAuth server cert ConfigMaps. Multicluster will not be enabled: %v", oAuthServerCertConfigMapsErr)
```

spadgett (commented):
Thanks, @TheRealJon

pkg/console/controllers/managedclusters/controller.go
TheRealJon: No, I think we will have to add some special logic for that to patch the Oauth client on the spoke cluster.

spadgett (approved):
/lgtm

Thanks @TheRealJon. Let's open a follow up issue for handling console hostname changes.

TheRealJon:
Created https://issues.redhat.com/browse/CONSOLE-3011 to follow-up on the cluster hostname edge case.

pkg/console/subresource/managedclusteraction/oauth-client.go
```diff
@@ -1,31 +1,29 @@
...
-func DefaultCreateOAuthClient(cr *operatorv1.Console, cn string, sec string, redirects []string) *unstructured.Unstructured {
-	managedClusterAction := CreateOAuthClientStub(cn)
-	withDefaultCreateOAuthClientInfo(managedClusterAction, cn, sec, redirects)
-	return managedClusterAction
-}
-
-func withDefaultCreateOAuthClientInfo(mca *unstructured.Unstructured, cn string, sec string, redirects []string) {
+func DefaultCreateOAuthClientAction(cn string, sec string, redirects []string) *unstructured.Unstructured {
+	mca := CreateOAuthClientStub(cn)
+	unstructured.SetNestedField(mca.Object, api.CreateOAuthClientManagedClusterActionName, "metadata", "name")
```
jhadvig, line 16: We should check for errors that `SetNested...` methods return

pkg/console/subresource/managedclusterview/oauth-client.go
```diff
@@ -1,30 +1,27 @@
...
-func DefaultViewOAuthClient(cr *operatorv1.Console, cn string) *unstructured.Unstructured {
-	managedClusterView := ViewOAuthClientStub(cn)
-	withDefaultViewOAuthClientInfo(managedClusterView, cn)
-	return managedClusterView
-}
-
-func withDefaultViewOAuthClientInfo(mcv *unstructured.Unstructured, cn string) {
+func DefaultOAuthClientView(cn string) *unstructured.Unstructured {
+	mcv := OAuthClientViewStub()
+	unstructured.SetNestedField(mcv.Object, api.OAuthClientManagedClusterViewName, "metadata", "name")
```
jhadvig, line 17: same here for the `SetNested...` methods

jhadvig (commented):
@TheRealJon I see really nice work here! Adding couple of comments.

manifests/03-rbac-role-cluster.yaml
```diff
@@ -160,3 +125,36 @@ rules:
...
+      - action.open-cluster-management.io
+    resources:
+      - managedclusteractions
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+      - update
+      - delete
```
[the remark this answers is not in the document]
TheRealJon: Actually, we probably can just remove the update verb. We shouldn't ever need to update/patch MCAs or MCVs. We would probably just delete and recreate them if anything.

examples/cvo-unmanage-operator.yaml
```diff
@@ -9,4 +9,9 @@ spec:
       name: console-operator
       namespace: openshift-console-operator
       unmanaged: true
-      group: apps/v1
+      group: apps
+    - kind: ClusterRole
```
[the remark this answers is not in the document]
TheRealJon: This was related to a regression that was introduced recently. The override[].group property was ignored prior to a recent update, so it didn't matter what value was there. The updates team recently added validation for that field, which should only contain group, and not group/version. So "apps/v1" was causing the override not to work at all.

TheRealJon:
@jhadvig If it's okay with you, I might wait to address these changes in a follow-on PR after we've merged the feature into master since none of the changes seem to be too impactful.

pkg/console/controllers/managedclusters/controller.go
```diff
@@ -229,18 +229,21 @@ func (c *ManagedClusterController) SyncManagedClusterList(ctx context.Context) (
...
-			mcvName, _ := managedclusterviewsub.GetName(required)
-			mcv, err = c.dynamicClient.Resource(api.ManagedClusterViewGroupVersionResource).Namespace(managedCluster.Name).Get(ctx, mcvName, metav1.GetOptions{})
+		mcv, err := c.dynamicClient.Resource(api.ManagedClusterViewGroupVersionResource).Namespace(managedCluster.Name).Get(ctx, api.OAuthClientManagedClusterViewName, metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			required, err := managedclusterviewsub.DefaultOAuthClientView(managedCluster.Name)
+			if err != nil {
+				errs = append(errs, fmt.Sprintf("Error initializing oauth client ManagedClusterView for cluster %s: %v", managedCluster.Name, err))
+				continue
+			}
+			mcv, err = c.dynamicClient.Resource(api.ManagedClusterViewGroupVersionResource).Namespace(managedCluster.Name).Create(ctx, required, metav1.CreateOptions{})
```
TheRealJon, line 246: Since we have to do error checking on initializing the required MCV, the logic needed to be switched around here. Instead of trying to create the resource, then handling the case where it already exists, we check to see if it already exists, then create it if not.

pkg/console/subresource/managedclusteraction/create_oauth_client.go
```diff
@@ -11,15 +11,18 @@ import (
...
-	return mca
+	err := unstructured.SetNestedField(mca.Object, api.CreateOAuthClientManagedClusterActionName, "metadata", "name")
+	err = unstructured.SetNestedField(mca.Object, cn, "metadata", "namespace")
+	err = unstructured.SetNestedStringMap(mca.Object, util.LabelsForManagedClusterResources(cn), "metadata", "labels")
+	err = unstructured.SetNestedField(mca.Object, sec, "spec", "kube", "template", "secret")
+	err = unstructured.SetNestedField(mca.Object, api.ManagedClusterOAuthClientName, "spec", "kube", "template", "metadata", "name")
+	err = unstructured.SetNestedStringSlice(mca.Object, redirects, "spec", "kube", "template", "redirectURIs")
+	if err != nil {
+		return nil, err
+	}
```
TheRealJon, line 24: I didn't see a need for an early return between each unstructured.SetNested... call, but  we can do that if anyone feels otherwise.

jhadvig: nit:
yes it does not make sense to check for errors after each call, but this way we can loose an error if multiple errors occur. For this case we have `"k8s.io/apimachinery/pkg/util/errors"` package.
```go
	var errors []error

	errors = append(errors, unstructured.SetNestedField(mca.Object, api.CreateOAuthClientManagedClusterActionName, "metadata", "name"))
	errors = append(errors, unstructured.SetNestedField(mca.Object, cn, "metadata", "namespace"))
        ...
	if err != nil {
		return nil, utilerrors.NewAggregate(errors)
	}
```

pkg/console/operator/sync_v400.go
```diff
@@ -241,19 +263,37 @@ func (co *consoleOperator) SyncDeployment(
...
-		klog.Warningf("Unable to list managed cluster API server CA bundle ConfigMaps. Multicluster will not be enabled: %v", clusterCAConfigMapsErr)
-	}
-	requiredDeployment := deploymentsub.DefaultDeployment(operatorConfig, cm, apiServerCAConfigMaps, serviceCAConfigMap, oauthServingCertConfigMap, trustedCAConfigMap, sec, proxyConfig, infrastructureConfig, canMountCustomLogo, canMountManagedClusterConfig)
+	apiServerCertConfigMaps, apiServerCertConfigMapsErr := co.configMapClient.ConfigMaps(api.OpenShiftConsoleNamespace).List(ctx, metav1.ListOptions{LabelSelector: api.ManagedClusterAPIServerCertName})
+	if apiServerCertConfigMapsErr != nil {
+		klog.Warningf("Unable to list managed cluster API server cert ConfigMaps. Multicluster will not be enabled: %v", apiServerCertConfigMapsErr)
+	}
+
+	oAuthServerCertConfigMaps, oAuthServerCertConfigMapsErr := co.configMapClient.ConfigMaps(api.OpenShiftConsoleNamespace).List(ctx, metav1.ListOptions{LabelSelector: api.ManagedClusterOAuthServerCertName})
+	if apiServerCertConfigMapsErr != nil {
```
jhadvig, line 280:
```suggestion
	if oAuthServerCertConfigMapsErr != nil {
```

pkg/console/subresource/configmap/managed_cluster_oauth_server_cert.go
```diff
@@ -0,0 +1,41 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	"github.com/openshift/console-operator/pkg/api"
+	"github.com/openshift/console-operator/pkg/console/subresource/util"
+)
+
+func DefaultManagedClusterOAuthServerCertConfigMap(clusterName string, caBundle string, cr *operatorv1.Console) *corev1.ConfigMap {
```
jhadvig, line 14: I cannot find where this function is used?

pkg/console/subresource/configmap/managed_cluster_oauth_server_cert.go
```diff
@@ -0,0 +1,41 @@
...
+	configMap := ManagedClusterOAuthServerCertConfigMapStub(clusterName)
+
+	if caBundle != "" {
```
jhadvig, line 17: does it make sense to call `DefaultManagedClusterOAuthServerCertConfigMap` function if the `caBundle` is empty?

pkg/console/subresource/managedclusterview/oauth_client.go
```diff
@@ -52,16 +52,16 @@ func GetName(mcv *unstructured.Unstructured) (string, error) {
...
 }
 
-func GetGroupVersionResource() schema.GroupVersionResource {
-	return schema.GroupVersionResource{
-		Group:    "view.open-cluster-management.io",
-		Version:  "v1beta1",
-		Resource: "managedclusterviews",
+func GetOAuthClientSecret(mcv *unstructured.Unstructured) (string, error) {
+	secret, found, err := unstructured.NestedString(mcv.Object, "status", "result", "secret")
+	if err != nil || !found {
```
jhadvig, line 63: we can directly return these values
```suggestion
        return secret, err
```

jhadvig: Also I cant find where this function is being used.

pkg/console/subresource/managedclusterview/oauth_server_cert.go
```diff
@@ -12,22 +12,23 @@ import (
...
-
-func withDefaultViewIngressCertInfo(mcv *unstructured.Unstructured, cn string) {
-	unstructured.SetNestedField(mcv.Object, cn, "metadata", "namespace")
+func DefaultOAuthServerCertView(cr *operatorv1.Console, cn string) (*unstructured.Unstructured, error) {
+	mcv := OAuthServerCertViewStub(cn)
+	err := unstructured.SetNestedField(mcv.Object, api.OAuthServerCertManagedClusterViewName, "metadata", "name")
+	err = unstructured.SetNestedField(mcv.Object, cn, "metadata", "namespace")
+	err = unstructured.SetNestedStringMap(mcv.Object, util.LabelsForManagedClusterResources(cn), "metadata", "labels")
+	if err != nil {
+		return nil, err
```
jhadvig, line 21: we should return aggregated error, check https://github.com/openshift/console-operator/pull/612#discussion_r787959525

jhadvig (commented):
@TheRealJon found a few additional inconsistencies, PTAL

pkg/console/subresource/configmap/managed_cluster_oauth_server_cert.go
TheRealJon: https://github.com/TheRealJon/console-operator/blob/5c44b4725b3c6d3a76ffd517968d1bc6e45c00c1/pkg/console/controllers/managedclusters/controller.go#L346

pkg/console/subresource/configmap/managed_cluster_oauth_server_cert.go
TheRealJon: I'll remove this empty string check. We check for empty string before calling this, so it should never be empty.

pkg/console/subresource/managedclusterview/oauth_client.go
TheRealJon: https://github.com/TheRealJon/console-operator/blob/5c44b4725b3c6d3a76ffd517968d1bc6e45c00c1/pkg/console/controllers/managedclusters/controller.go#L437

pkg/console/subresource/managedclusteraction/create_oauth_client.go
TheRealJon: Wouldn't the `errors` array contain nil values if we do it this way?

TheRealJon: Nevermind, I see that the NewAggregate func handles that.

TheRealJon:
@jhadvig I've addressed your feedback, let me know if this looks good and I'll squash

spadgett:
/override ci/prow/e2e-aws-single-node
/override ci/prow/e2e-aws-operator
/override ci/prow/e2e-agnostic-upgrade

Overriding statuses since this is a PR against a feature branch. We will make sure it passes CI before merging to master.

jhadvig (approved):
/lgtm

TheRealJon:
Closing as all of these changes are included in https://github.com/openshift/console-operator/pull/630

PR #614, opened 2021-12-07
[WIP] Update console operator cluster role rbac manifests

TheRealJon:
Only provide ManagedClusterAction and ManagedClusterView permissions to the console operator when tech preview is enabled.


TheRealJon:
Rolling this into #612

PR #615, opened 2021-12-08
Library go bump

eggfoobar:
Bumping the library-go dependency to latest, this should take advantage of leader election changes for SNO clusters proposed in this [library-go PR](https://github.com/openshift/library-go/pull/1242) and performance improvements in this [PR](https://github.com/openshift/library-go/pull/1235).

Changes:
- updated library-go to use latest `release-4.10` branch
- updated method calls to reflect new method signatures with passed in context
- updated to use new `managementstatecontroller` package instead of old `managementstate` package

PR #616, opened 2021-12-08
Fast forward multi cluster feature branch

TheRealJon:


florkbr:
@TheRealJon the fix for the failing verify job on master was merged in on https://github.com/openshift/console-operator/pull/615. Did you want to include that PR in this fast forward? Or should I open a PR with the fix I had opened in https://github.com/openshift/console-operator/pull/619 against `master-multi-cluster-feature`?

TheRealJon:
@florkbr I'm going to close this PR. I think there is a legit test failure happening and rather than try to fix it here, I'm going to address the issue in #612

PR #613, opened 2021-11-24
CONSOLE-3005: Provide alias for dynamic plugin's proxy

jhadvig:
Story: https://issues.redhat.com/browse/CONSOLE-3005

Changes are based on the [API change PR](https://github.com/openshift/api/pull/1067)

/assign @christianvogt @spadgett

PR #618, opened 2021-12-14
Bug 2032289: Add test case

jotak:
Adding a test case related to https://github.com/openshift/api/pull/1085

Empty fields in ConsolePlugin Proxy , such as CACertificate, should
remain empty and not be transformed to an empty string, so that it isn't validated against regex

### DIFF
--- a/pkg/console/subresource/configmap/configmap_test.go
+++ b/pkg/console/subresource/configmap/configmap_test.go
@@ -742,6 +742,73 @@ nV5cXbp9W1bC12Tc8nnNXn4ypLE2JTQAvyp51zoZ8hQoSnRVx/VCY55Yu+br8gQZ` + "\n" + `
 			},
 		},
 		{
+			name: "Test operator config, with enabledPlugins set, without CA certificate",
+			args: args{
+				operatorConfig: &operatorv1.Console{},
+				consoleConfig:  &configv1.Console{},
+				managedConfig:  &corev1.ConfigMap{},
+				infrastructureConfig: &configv1.Infrastructure{
+					Status: configv1.InfrastructureStatus{
+						APIServerURL: mockAPIServer,
+					},
+				},
+				rt: &routev1.Route{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: api.OpenShiftConsoleName,
+					},
+					Spec: routev1.RouteSpec{
+						Host: host,
+					},
+				},
+				useDefaultCAFile:         true,
+				inactivityTimeoutSeconds: 0,
+				availablePlugins: []*v1alpha1.ConsolePlugin{
+					testPlugins("plugin1", "service1", "service-namespace1"),
+					testPluginsWithProxyNoCA("plugin2", "service2", "service-namespace2"),
+				},
+				managedClusterConfigFile: "",
+			},
+			want: &corev1.ConfigMap{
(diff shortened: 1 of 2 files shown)


jhadvig (commented):
Thanks @jotak for the TC! A change in the the ConsolePlugin proxy is about to land in https://github.com/openshift/api/pull/1067 ... We will need to wait with this PR until then.

jhadvig:
@jotak so https://github.com/openshift/console-operator/pull/613 got merged. PLease rebase on top of master.
For more info check https://issues.redhat.com/browse/OCPPLAN-8045 or https://github.com/openshift/enhancements/blob/master/enhancements/console/dynamic-plugins.md#delivering-plugins

jotak:
thanks @jhadvig 
do i have to switch to go 1.17? looks like i can't build with 1.16

jotak:
@jhadvig to clarify, I get:

```bash
vendor/sigs.k8s.io/json/internal/golang/encoding/json/encode.go:1249:12: sf.IsExported undefined (type reflect.StructField has no field or method IsExported)
vendor/sigs.k8s.io/json/internal/golang/encoding/json/encode.go:1255:18: sf.IsExported undefined (type reflect.StructField has no field or method IsExported)
```

we had the same issue on our operator. AFAIK that's because this lib now uses features of go1.17 that aren't backward compatible. In our case that was bumped as a transitive dependency, we needed to pin the "faulty" dep, `crdoc` in our case: https://github.com/netobserv/network-observability-operator/pull/42/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R91 

note that we stick on go1.16 at the moment until a [ubi8 go-toolset](https://catalog.redhat.com/software/containers/ubi8/go-toolset/5ce8713aac3db925c03774d1) image is released for 1.17 .. maybe it's the same for you?

jotak:
I rebased, but wasn't able to run the test case due to the issue mentioned above

jhadvig:
yes, you have to switch to go-1.17.6

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -615,6 +615,67 @@ nV5cXbp9W1bC12Tc8nnNXn4ypLE2JTQAvyp51zoZ8hQoSnRVx/VCY55Yu+br8gQZ` + "\n" + `
...
+  bindAddress: https://[::]:8443
+  certFile: /var/serving-cert/tls.crt
+  keyFile: /var/serving-cert/tls.key
+providers: {}
+plugins:
+  plugin1: https://serviceName1.serviceNamespace1.svc.cluster.local:8443/
+  plugin2: https://serviceName2.serviceNamespace2.svc.cluster.local:8443/
+proxy:
+  services:
+  - consoleAPIPath: /api/proxy/namespace/proxy-serviceNamespace2/service/proxy-serviceName2:9991/
```
jhadvig, line 677: so the endpoint path has changed, check https://github.com/openshift/enhancements/blob/master/enhancements/console/dynamic-plugins.md#delivering-plugins

jotak:
hey these three tests keep failing but I don't see the relation with my changes, don't know what I can do more

jhadvig:
@jotak from what I see I think those are mostly test flakes.. but agree that there is just too many of them

/retest

PR #621, opened 2022-01-03
Bug 2030574: Console resources are using older 'service.alpha.openshift.io' annotation

jhadvig:
/assign @spadgett


spadgett (approved):
/lgtm

PR #623, opened 2022-01-07
Bug 2037635: Fix setting of custom cert for default route

jhadvig:
Issue was with setting an ingress's `componentRoute` only with TLS secret(not supplying hostname), which should be set for default route. Fixing the issue and adding two e2e tests.

/assign @TheRealJon @spadgett

### DIFF
--- a/pkg/console/controllers/route/controller.go
+++ b/pkg/console/controllers/route/controller.go
@@ -2,8 +2,6 @@ package route
 
 import (
 	"context"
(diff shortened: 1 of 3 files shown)


yapei:
tested the PR and confirmed that it is working as expected, but I will do some more regression testing after the PR is merged

spadgett (approved):
/lgtm

PR #638, opened 2022-02-21
Bug 2055494: console-operator should report Upgradeable False when SAN-less certs are used

jhadvig:
Adding verification for checking if custom certificate contains SAN(Subject Alternative Names) field. If it does not we should be setting `Upgradable` to `False` for console-operator, to prevent 4.10 upgrade. Otherwise the cluster will get broken.

Based on the discussion this fix is only necessary for 4.9

/assign @stlaz


pkg/console/subresource/route/route.go
```diff
@@ -206,21 +207,19 @@ func GetCustomTLS(customCertSecret *corev1.Secret) (*CustomTLSCert, error) {
...
 	key, keyExist := customCertSecret.Data["tls.key"]
 	if !keyExist {
 		return nil, fmt.Errorf("custom cert secret data doesn't contain 'tls.key' entry")
 	}
-
 	privateKeyVerifyErr := privateKeyVerifier(key)
 	if privateKeyVerifyErr != nil {
 		return nil, fmt.Errorf("failed to verify custom key PEM: " + privateKeyVerifyErr.Error())
 	}
+	certificateVerifyErr := certificateVerifier(cert)
```
stlaz, line 218: I think you may want to keep this as is and move it back to where you're checking for certificate existence.

pkg/console/subresource/route/route.go
```diff
@@ -242,6 +241,9 @@ func certificateVerifier(customCert []byte) error {
 	if now.Before(certificate.NotBefore) {
 		return fmt.Errorf("custom TLS certificate is not valid yet")
 	}
+	if !oscrypto.CertHasSAN(certificate) {
```
stlaz, line 244: `certificateVerifier` has no godoc, I am therefore assuming that it can be used generically on any CA and server certs. CA certificates won't contain SANs, though, so you may want to move this somewhere outside, or rename `certificateVerifier` to `serverCertificateVerifier`, or create a wrapping `serverCertificateVerifier`.

test/e2e/custom_url_test.go
```diff
@@ -466,18 +552,10 @@ func getComponentRouteSpec(t *testing.T, client *framework.ClientSet, routeName
 	return componentRouteSpec
 }
 
-func createTLSSecret(t *testing.T, client *framework.ClientSet, tlsSecretName, hostname string) {
+func createTLSSecret(t *testing.T, client *framework.ClientSet, tlsSecretName string) {
 	if tlsSecretName == "" {
 		return
 	}
-	tlsCert, err := crypto.MakeSelfSignedCAConfig(hostname, 1)
```
stlaz, line 473: have you checked whether it's possible in Go to create a certificate w/o SAN?

pkg/console/controllers/route/controller.go
```diff
@@ -128,12 +128,14 @@ func (c *RouteSyncController) Sync(ctx context.Context, controllerContext factor
 	// route into inaccessible state.
 	_, customRouteErrReason, customRouteErr := c.SyncCustomRoute(ctx, routeConfig, controllerContext)
 	statusHandler.AddConditions(status.HandleProgressingOrDegraded("CustomRouteSync", customRouteErrReason, customRouteErr))
+	statusHandler.AddCondition(status.HandleUpgradable("DefaultRouteSync", customRouteErrReason, customRouteErr))
```
stlaz, line 131: Isn't this once again `CustomRouteSync`?

Do you want to distinguish the error types for upgradable and degraded cases?

pkg/console/subresource/route/route.go
jhadvig: created a wrapper function

test/e2e/custom_url_test.go
```diff
@@ -22,14 +22,100 @@ import (
...
 )
 
 const (
 	consoleRouteCustomTLSSecretName   = "console-route-custom-tls"
 	downloadsRouteCustomTLSSecretName = "downloads-route-custom-tls"
 )
 
+const (
+	validCertificate = `-----BEGIN CERTIFICATE-----
+MIIFeDCCA2CgAwIBAgIJALcdZxainrkZMA0GCSqGSIb3DQEBCwUAMFMxCzAJBgNV
```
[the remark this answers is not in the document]
jhadvig: added script into `script/gencert.sh`

pkg/console/controllers/route/controller.go
jhadvig: We are distinguishing between them, since we are calling both:
- `HandleProgressingOrDegraded`, which will set operator's `Degraded` condition to `True` if an error occurs
- `HandleUpgradable`, which will set operator's `Upgradable` condition to `False` if an error occurs

pkg/console/subresource/route/route.go
jhadvig: moved back 👍

test/e2e/custom_url_test.go
jhadvig: so I was checking it but most of the time openssl was used to generate the certificate.

pkg/console/subresource/route/route.go
stlaz: it should be wrapped the other way around, this way you're still checking the cert for SAN while the certificate can be a CA cert.

test/e2e/custom_url_test.go
```diff
@@ -22,14 +22,100 @@ import (
...
-	"github.com/openshift/library-go/pkg/crypto"
 )
 
 const (
 	consoleRouteCustomTLSSecretName   = "console-route-custom-tls"
 	downloadsRouteCustomTLSSecretName = "downloads-route-custom-tls"
 )
 
+const (
+	validCertificate = `-----BEGIN CERTIFICATE-----
```
stlaz, line 33: You should also test the invalid case. I was recently reminded that I actually contributed a script to generate all kinds of certificates you may need, see https://www.github.com/kubernetes/kubernetes/blob/5479306b1326bcbf29ae2ca0694a36555ed1ce35/staging/src/k8s.io/apiserver/pkg/util/webhook/gencerts.sh

stlaz (changes requested):

stlaz:
There seem to be console pod replicas missing for a long time in the tests, I wonder why: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_console-operator/638/pull-ci-openshift-console-operator-release-4.9-e2e-gcp/1511625326034161664/artifacts/e2e-gcp/gather-extra/artifacts/pods/openshift-console-operator_console-operator-6dd4cb767b-nsq5l_console-operator.log
Possibly unrelated.

jhadvig:
@stlaz this is a flake from what I understand. We need to backport to 4.9 https://github.com/openshift/console-operator/pull/646 first though.

stlaz:
/lgtm
I don't like the GCP failures but they really don't seem to be related to the PR

stlaz:
/lgtm cancel
from the logs:
```
Some cluster operators are not ready: console (Upgradeable=False CustomRouteSync::DefaultRouteSync: )}
```
Upgradable is set to false for some reason, and it seems like it could be doing of this PR in fact

jhadvig:
Adding label manually since this is not a cherry-pick but rather a fix for a given release version

PR #642, opened 2022-03-06
Bug 2057696: Console operator should not block installation/upgrade process when set to Removed state

jhadvig:
We havent been setting the right conditions when the console-operator has been set to `Removed` state. The correct ones should be `Available: True` & `Degraded: False`, when no error occurs during the removal process. 

/assign @TheRealJon

### DIFF
--- a/pkg/console/operator/operator.go
+++ b/pkg/console/operator/operator.go
@@ -7,7 +7,7 @@ import (
 	"time"
 
 	// kube
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	corev1 "k8s.io/client-go/informers/core/v1"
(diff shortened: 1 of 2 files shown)


pkg/console/operator/operator.go
```diff
@@ -273,5 +276,16 @@ func (c *consoleOperator) removeConsole(ctx context.Context, recorder events.Rec
...
+	err := utilerrors.FilterOut(utilerrors.NewAggregate(errs), apierrors.IsNotFound)
+
+	if err != nil {
+		statusHandler.AddCondition(consolestatus.HandleAvailable("", "ConsoleRemoved", err))
+		statusHandler.AddCondition(consolestatus.HandleDegraded("", "ConsoleRemoved", err))
+		return err
+	}
+
+	statusHandler.AddCondition(consolestatus.HandleAvailable("", "ConsoleRemoved", nil))
+	statusHandler.AddCondition(consolestatus.HandleDegraded("", "ConsoleRemoved", nil))
```
spadgett, line 289: Do we need to handle upgradeable as well?

jhadvig: Was wondering about that TBH, we probably should.

PR #640, opened 2022-02-22
Bug 2046497: Re-enable TestMetricsEndpoint e2e test case

TheRealJon:
Improve logging and add polling to TestMetricsEndpoint e2e test to prevent flakes

PR #645, opened 2022-04-07
Bug 2046497: Metrics e2e should not fail on first failed polling attempt

jhadvig:
We need to replace `Errorf` with `Logf`. Its my failt, I should have noticed this when reviewing https://github.com/openshift/console-operator/pull/640

```go
// Errorf is equivalent to Logf followed by Fail.
func (c *common) Errorf(format string, args ...interface{}) {
	c.log(fmt.Sprintf(format, args...))
	c.Fail()
}

// Fail marks the function as having failed but continues execution.
func (c *common) Fail() {
    ...
}
```

/assign @TheRealJon


TheRealJon (approved):
/lgtm

PR #650, opened 2022-04-20
Bug 2067155: Modify the operator display name to match it with the name displayed in operatorhub

karthikjeeyar:
Fixes:
https://bugzilla.redhat.com/show_bug.cgi?id=2067155

Description:
Quickstart installation step contains outdated operator name, so users following that step will not be able to find the operator by following the steps.

Solution:
Updated the content with latest operator name.
     `Openshift Pipelines Operator` is changed to `Red Hat Openshift Pipelines`

cc: @serenamarie125

### DIFF
--- a/quickstarts/explore-pipeline-quickstart.yaml
+++ b/quickstarts/explore-pipeline-quickstart.yaml
@@ -7,10 +7,10 @@ metadata:
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
 spec:
-  displayName: Install the OpenShift Pipelines Operator
+  displayName: Install the Red Hat OpenShift Pipelines
   durationMinutes: 10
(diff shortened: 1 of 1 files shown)


quickstarts/explore-pipeline-quickstart.yaml
```diff
@@ -51,18 +50,18 @@ spec:
           Pipelines Operator's status to change from **Installing** to **Succeeded**.
       review:
         instructions: |-
-          #### To verify that the OpenShift Pipelines Operator is installed:
+          #### To verify that the Red Hat OpenShift Pipelines Operator is installed:
           1. From the **Operators** section of the navigation, go to the **Installed Operators** page.
-          2. Verify that the **OpenShift Pipelines Operator** appears in the list of Operators.
+          2. Verify that the **Red Hat OpenShift Pipelines** appears in the list of Operators.
```
jhadvig, line 55: Shouldn't this be?
```suggestion
          2. Verify that the **Red Hat OpenShift Pipelines Operator** appears in the list of Operators.
```

quickstarts/explore-pipeline-quickstart.yaml
```diff
@@ -51,18 +50,18 @@ spec:
...
 
-          In the status column, is the status of the OpenShift Pipelines Operator **Succeeded**?
+          In the status column, is the status of the Red Hat OpenShift Pipelines **Succeeded**?
```
jhadvig, line 57: same as above 
```suggestion
          In the status column, is the status of the Red Hat OpenShift Pipelines Operator **Succeeded**?
```

jhadvig (approved):
@karthikjeeyar thanks for fix! Added two comments about the wording. PTAL

quickstarts/explore-pipeline-quickstart.yaml
karthikjeeyar: This is a manual verification step that users would do to verify if their installation is success or not, so the highlighted text should exactly match the operator display name in the UI. This applies to be below comment as well.


image

karthikjeeyar:
Thanks @jhadvig for the review.  I modified all the instances in this QS which had `Openshift Pipelines Operator` (old name) to `Red Hat Openshift Pipelines` to be consistent. PTAL

jhadvig (approved):
@karthikjeeyar thanks!
/approve
/lgtm

PR #652, opened 2022-05-13
Bug 2086519: AUTH-133: manifests: comply to restricted pod security

s-urbaniak:
Starting from OpenShift 4.11 pod security admission is being activated. In OpenShift the default pod security admission level is going to be `restricted`. This PR fixes workloads from this repository. Concretely, the following violations have been detected:

```
{
  "objectRef": "openshift-console-operator/pods/",
  "pod-security.kubernetes.io/audit-violations": "would violate PodSecurity \"restricted:latest\": runAsNonRoot != true (pod or container \"console-operator\" must set securityContext.runAsNonRoot=true)"
}
```

/cc @stlaz

### DIFF
--- a/bindata/deployments/console-deployment.yaml
+++ b/bindata/deployments/console-deployment.yaml
@@ -27,7 +27,10 @@ spec:
       restartPolicy: Always
       serviceAccountName: console
       schedulerName: default-scheduler
-      securityContext: {}
+      securityContext:
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
       terminationGracePeriodSeconds: 40
       priorityClassName: system-cluster-critical
       containers:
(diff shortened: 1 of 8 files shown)


jhadvig (commented):
@s-urbaniak thanks for the PR. The change looks good, but wondering if the operand's `console` and `downloads` deployments should also include security context?

s-urbaniak:
@jhadvig can you please give me some pointers in the code where I can do the changes? thank you 👍 is there anything in your operator wrt to upgrade logic and operands we have to take care about?

s-urbaniak:
presumably it's just updates in bindata?

s-urbaniak:
@jhadvig updated, ptal

jhadvig (approved):
@s-urbaniak thanks looks good :)
/lgtm
/approve

spadgett:
@s-urbaniak Console has a bunch of YAML examples for workloads like deployments and pods. For example:

https://github.com/openshift/console/blob/master/frontend/public/models/yaml-templates.ts#L198-L221

Should we update the samples to include `securityContext` as well? We try to make sure these run successfully with no changes.

PR #654, opened 2022-05-25
CONSOLE-3162: Implement check for the new i18n annotation for dynamic plugins

jhadvig:
console-operator should be checking for the new `console.openshift.io/default-i18next-namespace` annotation, update the `console-config.yaml` accordingly and redeploy the console server.

For that I've refact the internal consoleserver API.

/assign @spadgett @TheRealJon 

Story: https://issues.redhat.com/browse/CONSOLE-3162

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -54,6 +54,8 @@ const (
 	CreateOAuthClientManagedClusterActionName = "console-create-oauth-client"
 	OAuthServerCertManagedClusterViewName     = "console-oauth-server-cert"
 
+	PluginI18nAnnotation = "console.openshift.io/use-i18n"
+
 	ManagedClusterViewAPIGroup     = "view.open-cluster-management.io"
 	ManagedClusterViewAPIVersion   = "v1beta1"
 	ManagedClusterViewResource     = "managedclusterviews"
--- a/pkg/console/subresource/configmap/configmap_test.go
+++ b/pkg/console/subresource/configmap/configmap_test.go
@@ -663,8 +663,10 @@ managedClusterConfigFile: 'test'
 				useDefaultCAFile:         true,
 				inactivityTimeoutSeconds: 0,
 				availablePlugins: []*v1alpha1.ConsolePlugin{
-					testPlugins("pluginName1", "serviceName1", "serviceNamespace1"),
-					testPluginsWithProxy("pluginName2", "serviceName2", "serviceNamespace2"),
+					testPlugins("plugin1", "service1", "service-namespace1"),
+					testPluginsWithProxy("plugin2", "service2", "service-namespace2"),
+					testPluginsWithI18nAnnotation("plugin3", "service3", "service-namespace3", "false"),
+					testPluginsWithI18nAnnotation("plugin4", "service4", "service-namespace4", "true"),
 				},
 				managedClusterConfigFile: "",
 			},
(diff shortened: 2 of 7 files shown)


jhadvig:
The tests are failing most probably due to the internal API change

pkg/api/api.go
```diff
@@ -54,6 +54,8 @@ const (
 	CreateOAuthClientManagedClusterActionName = "console-create-oauth-client"
 	OAuthServerCertManagedClusterViewName     = "console-oauth-server-cert"
 
+	PluginI18nAnnotation = "console.openshift.io/default-i18next-namespace"
```
spadgett, line 57: I should have made this comment on the enhancement, but I feel like this annotation sounds like the namespace name is meant to be the value. Maybe `console.openshift.io/use-default-i18next-namespace`? Or maybe even `console.openshift.io/use-i18next` to keep it short?

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -697,8 +698,12 @@ servingInfo:
   keyFile: /var/serving-cert/tls.key
 providers: {}
 plugins:
-  pluginName1: https://serviceName1.serviceNamespace1.svc.cluster.local:8443/
-  pluginName2: https://serviceName2.serviceNamespace2.svc.cluster.local:8443/
+  enabled:
+    pluginName1: https://serviceName1.serviceNamespace1.svc.cluster.local:8443/
+    pluginName2: https://serviceName2.serviceNamespace2.svc.cluster.local:8443/
+    pluginName3: https://serviceName3.serviceNamespace3.svc.cluster.local:8443/
```
spadgett, line 704: This is a nit since it's only a test, but these aren't valid plugin names since they have uppercase letters. I'd make it `plugin1`, `plugin2`, `plugin3`. (Same for `serviceName`.)

pkg/console/subresource/configmap/configmap_test.go
```diff
@@ -850,6 +855,12 @@ func testPlugins(pluginName, serviceName, serviceNamespace string) *v1alpha1.Con
 	}
 }
 
+func testPluginsI18nAnnotation(pluginName, serviceName, serviceNamespace string) *v1alpha1.ConsolePlugin {
+	plugin := testPlugins(pluginName, serviceName, serviceNamespace)
+	plugin.SetAnnotations(map[string]string{api.PluginI18nAnnotation: "true"})
```
spadgett, line 860: I would add a test for `"false"` and other values to make sure they're not added as namespaces.

pkg/console/subresource/consoleserver/types.go
```diff
@@ -22,9 +22,14 @@ type Config struct {
...
-	Plugins                  map[string]string `yaml:"plugins,omitempty"`
-	Proxy                    Proxy             `yaml:"proxy,omitempty"`
-	ManagedClusterConfigFile string            `yaml:"managedClusterConfigFile,omitempty"`
+	Plugins                  Plugins `yaml:"plugins,omitempty"`
+	Proxy                    Proxy   `yaml:"proxy,omitempty"`
+	ManagedClusterConfigFile string  `yaml:"managedClusterConfigFile,omitempty"`
+}
+
+type Plugins struct {
+	Enabled    map[string]string `yaml:"enabled,omitempty"`
```
spadgett, line 31: `enabled` sounds like it should be a boolean value. If we're going to break the API, I would change `plugins` to be an array of objects with values like `name`, `url`, `proxy`, `namespaces`, etc. I would be OK with leaving `plugins` as-is and just adding a separate `i18nNamespaces` array prop.

pkg/api/api.go
jhadvig: love the short one 👍

jhadvig:
@spadgett comments addressed. PTAL

spadgett (approved):
/lgtm

jhadvig:
QE Approver: 
/assign @yapei
Docs Approver:
 /assign @opayne1
PX Approver:
/assign @RickJWagner

spadgett (approved):
/lgtm

PR #659, opened 2022-06-21
Bug 2021297: Pass RELEASE_VERSION envar into the console pod

jhadvig:
We should be passing the RELEASE_VERSION envar into the console so we can then use is for [implementing dynamic plugin dependency resolution](https://github.com/openshift/console/pull/11626).

To achieve this we could:
1. Pass the RELEASE_VERSION envar into the console pod, so we so console backend can read it as a envar
2. We could be publishing the RELEASE_VERSION var into the console-config.yaml so the console server will read the variable and add it to `SERVER_FLAGS`. 

Based on [slack conversation](https://coreos.slack.com/archives/C011BL0FEKZ/p1653600473874119?thread_ts=1653599045.025239&cid=C011BL0FEKZ)

/assign @TheRealJon 
@vojtechszocs @spadgett FYI

### DIFF
--- a/pkg/console/subresource/configmap/configmap.go
+++ b/pkg/console/subresource/configmap/configmap.go
@@ -60,6 +60,7 @@ func DefaultConfigMap(
 		OAuthServingCert(useDefaultCAFile).
 		APIServerURL(getApiUrl(infrastructureConfig)).
 		InactivityTimeout(inactivityTimeoutSeconds).
+		ReleaseVersion().
 		ConfigYAML()
 	if err != nil {
 		klog.Errorf("failed to generate default console-config config: %v", err)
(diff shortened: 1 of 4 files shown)


spadgett (commented):
Thanks, Jakub. Can we add some unit tests?

Makefile
```diff
@@ -45,6 +45,7 @@ test: test-unit test-e2e
 .PHONY: test
 
 test-unit:
+    export RELEASE_VERSION ?=0.0.0-0.test-release-version
```
spadgett, line 48: This should be fine, but should we use `t.Setenv` in the tests instead? Then we can test what happens when there is no env var, and it will also let you test without using the `make` command.

spadgett: https://pkg.go.dev/testing#B.Setenv

PR #660, opened 2022-06-21
Bug 2099606: users with view only access should not see Quickstarts that require deploying applications

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/ODC-6034

**Root analysis:**
missing access review from quick starts

**Solution description:**
added access review with proper verb

**GIF:**
TODO

/kind bug


quickstarts/add-healthchecks-quickstart.yaml
```diff
@@ -12,6 +12,19 @@ spec:
...
+    verb: update
+  - group: apps
+    resource: deployments
+    verb: update
+  - group: apps.openshift.io
+    resource: deploymentconfigs
+    verb: update
+  - group: apps
+    resource: statefulsets
+    verb: update
```
spadgett, line 27: I believe this will check if the user can update these resources in all namespaces, which will hide the quick start for users who can only update these resources in certain workspaces. Have you tested with normal users?

This seems to be the case for other access reviews below as well.

quickstarts/add-healthchecks-quickstart.yaml
```diff
@@ -12,6 +12,19 @@ spec:
   description: You just created a sample application. Now, let’s add health checks
     to it.
   prerequisites: [You completed the "Getting started with a sample" quick start.]
+  accessReviewResources:
+  - group: apps
+    resource: daemonsets
+    verb: update
```
spadgett, line 18: Have you confirmed the health check editor doesn't use patch?

spadgett (commented):
I suspect it will be problematic to add access reviews since we don't have a way to check different user permissions in different namespaces. We don't have a good way to handle this today.

quickstarts/add-healthchecks-quickstart.yaml
debsmita1: I verified this on a dev cluster, it hides the quick start for users who can only update the resources in some namespaces.

debsmita1:
Closing this PR after verifying Sam's comment.

PR #662, opened 2022-07-08
Bug 2055494: Distinguish between route conditions

jhadvig:
Since the `RouteController` is used for `console` and `downloads` routes so we need to distinguish between there two routes and their conditions.

/assign @stlaz


jhadvig:
Adding label manually since this fix is only needed for 4.9 release.

jhadvig:
```
  1) Create namespace from install operators
       creates namespace from operator install page:
     CypressError: Timed out retrying after 30050ms: `cy.click()` failed because this element is `disabled`:
`<button aria-haspopup="true" aria-expanded="false" class="pf-c-dropdown__toggle" data-test-id="dropdown-button" type="button" id="dropdown-selectbox" data-test="dropdown-selectbox" disabled="">...</button>`
```

/test e2e-aws-console

PR #667, opened 2022-08-02
Bug 2117142: Update permission for projecthelmchartrepositories with an aggregator role

Kartikey-star:
This PR updates the permission change for Project Helm Chart Repository . The PR aims to aggregate the admin and edit cluster role with ability to perform actions on ProjectHelmChartRepository.

Signed-off-by: Kartikey Mamgain <mamgainkartikey@gmail.com>

### DIFF
--- a/manifests/03-rbac-role-cluster.yaml
+++ b/manifests/03-rbac-role-cluster.yaml
@@ -156,3 +156,24 @@ rules:
       - watch
       - create
       - delete
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: project-helm-chartrepository-editor
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+rules:
+- apiGroups:
+  - helm.openshift.io
+  resources:
+  - projecthelmchartrepositories
+  verbs:
+  - get
+  - list
+  - update
+  - create
+  - watch 
+  - patch
+  - delete


manifests/03-rbac-role-cluster.yaml
```diff
@@ -156,3 +156,41 @@ rules:
...
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+rules:
+- apiGroups:
+  - helm.openshift.io
+  resources:
+  - projecthelmchartrepositories
+  verbs:
+  - get
+  - list
```
spadgett, line 173: You almost always want `watch` if you have `list`.

manifests/03-rbac-role-cluster.yaml
```diff
@@ -156,3 +156,41 @@ rules:
...
+  - update
```
spadgett, line 174: You almost always want `patch` if you have `update`.

manifests/03-rbac-role-cluster.yaml
```diff
@@ -156,3 +156,41 @@ rules:
...
+  - create
+  - delete
```
spadgett, line 193: So only admins can delete? What is the reasoning?

If it's to protect an editor from deleting someone else's repository, they could always edit it to an invalid value so it's not really guarding much. I'd think we'd want one role for both admin and edit that has `watch`, `patch`, and `delete`.

cc @dperaza4dustbit

Kartikey-star: We can add watch and patch to the edit role , but i think the access to delete a repository should be given only to an admin. 
And if we have watch, patch, and delete  access for  both edit and admin then why to have two roles?(Because in that case essentially both are same).

spadgett: Can you expand on why only admins should delete?

If they are the same, we would only have one role.

Kartikey-star: If i am invited to a project by an admin i would expect to be able to get,list,create,watch the project helm chart repositories and not delete the existing repos unless been given the admin access explicitly.

dperaza4dustbit: Yeah this is looking more like we only need one role with two aggregator labels. I see your points Sam on edit and delete king of being married.

Kartikey-star: Made the changes as to have one role which has the access to watch patch and delete.

spadgett (approved):
/lgtm

spadgett:
@Kartikey-star Please open a bug to track this change. It will be required to backport to 4.11.

jhadvig (approved):

PR #669, opened 2022-08-09
CONSOLE-3242: Heterogeneous architecture clusters

jhadvig:
Implementing https://issues.redhat.com/browse/CONSOLE-3242

Implement logic in the console-operator that will scan though all the nodes and build a set of all the architecture types that the cluster nodes run on and pass it to the `console-config.yaml`

### DIFF
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -12,6 +12,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 
 	// openshift
@@ -367,6 +368,12 @@ func (co *consoleOperator) SyncConfigMap(
 		return nil, false, "FailedGetManagedConfig", mcErr
 	}
 
+	nodeList, nodeListErr := co.nodeClient.Nodes().List(ctx, metav1.ListOptions{})
+	if nodeListErr != nil {
+		return nil, false, "FailedListNodes", nodeListErr
+	}
+	nodeArchitectures := getNodeArchitectures(nodeList)
+
 	useDefaultCAFile := false
 	// We are syncing the `oauth-serving-cert` configmap from `openshift-config-managed` to `openshift-console`.
 	// `oauth-serving-cert` is only published in `openshift-config-managed` in OpenShift 4.9.0 and newer.
@@ -405,6 +412,7 @@ func (co *consoleOperator) SyncConfigMap(
 		inactivityTimeoutSeconds,
 		availablePlugins,
 		managedClusterConfigFile,
+		nodeArchitectures,
 	)
 	if err != nil {
 		return nil, false, "FailedConsoleConfigBuilder", err
@@ -592,4 +600,13 @@ func (co *consoleOperator) GetAvailablePlugins(enabledPluginsNames []string) []*
 		availablePlugins = append(availablePlugins, plugin)
 	}
 	return availablePlugins
+}
+
+func getNodeArchitectures(nodes *corev1.NodeList) []string {
+	nodeArchitecturesSet := sets.NewString()
+	for _, node := range nodes.Items {
+		nodeArch := node.Labels[api.NodeArchitectureLabel]
+		nodeArchitecturesSet.Insert(nodeArch)
+	}
+	return nodeArchitecturesSet.List()
 }
--- a/pkg/console/subresource/consoleserver/types.go
+++ b/pkg/console/subresource/consoleserver/types.go
@@ -65,6 +65,7 @@ type ClusterInfo struct {
 	MasterPublicURL     string                `yaml:"masterPublicURL,omitempty"`
 	ControlPlaneToplogy configv1.TopologyMode `yaml:"controlPlaneTopology,omitempty"`
 	ReleaseVersion      string                `yaml:"releaseVersion,omitempty"`
+	NodeArchitectures   []string              `yaml:"nodeArchitectures,omitempty"`
 }
 
 // Auth holds configuration for authenticating with OpenShift. The auth method is assumed to be "openshift".
--- a/manifests/03-rbac-role-cluster.yaml
+++ b/manifests/03-rbac-role-cluster.yaml
@@ -8,6 +8,14 @@ metadata:
     include.release.openshift.io/single-node-developer: "true"
 rules:
   - apiGroups:
+    - ""
+    resources:
+    - nodes
+    verbs:
+    - get
+    - list
+    - watch
+  - apiGroups:
       - oauth.openshift.io
     resources:
       - oauthclients
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -58,6 +58,8 @@ const (
 
 	PluginI18nAnnotation = "console.openshift.io/use-i18n"
 
+	NodeArchitectureLabel = "kubernetes.io/arch"
+
 	ManagedClusterViewAPIGroup     = "view.open-cluster-management.io"
 	ManagedClusterViewAPIVersion   = "v1beta1"
 	ManagedClusterViewResource     = "managedclusterviews"
--- a/pkg/console/operator/operator.go
+++ b/pkg/console/operator/operator.go
@@ -59,6 +59,7 @@ type consoleOperator struct {
 	secretsClient    coreclientv1.SecretsGetter
 	configMapClient  coreclientv1.ConfigMapsGetter
 	serviceClient    coreclientv1.ServicesGetter
+	nodeClient       coreclientv1.NodesGetter
 	deploymentClient appsclientv1.DeploymentsGetter
 	// openshift
 	routeClient   routeclientv1.RoutesGetter
@@ -113,6 +114,7 @@ func NewConsoleOperator(
 		secretsClient:    corev1Client,
 		configMapClient:  corev1Client,
 		serviceClient:    corev1Client,
+		nodeClient:       corev1Client,
 		deploymentClient: deploymentClient,
 		// openshift
 		routeClient:   routev1Client,
@@ -128,6 +130,7 @@ func NewConsoleOperator(
 	configMapInformer := coreV1.ConfigMaps()
 	managedConfigMapInformer := managedCoreV1.ConfigMaps()
 	serviceInformer := coreV1.Services()
+	nodeInformer := coreV1.Nodes()
 	configV1Informers := configInformer.Config().V1()
 	configNameFilter := util.IncludeNamesFilter(api.ConfigResourceName)
 	targetNameFilter := util.IncludeNamesFilter(api.OpenShiftConsoleName)
@@ -148,6 +151,7 @@ func NewConsoleOperator(
 		serviceInformer.Informer(),
 		oauthClients.Informer(),
 	).WithInformers(
+		nodeInformer.Informer(),
 		consolePluginInformer.Informer(),
 	).WithFilteredEventsInformers(
 		util.LabelFilter(map[string]string{"app": "console"}),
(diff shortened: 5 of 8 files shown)


spadgett:
@Prashanth684 @joelanford FYI

pkg/console/operator/sync_v400.go
```diff
@@ -593,3 +601,14 @@ func (co *consoleOperator) GetAvailablePlugins(enabledPluginsNames []string) []*
 	}
 	return availablePlugins
 }
+
+func getSupportedArchitectures(nodes *corev1.NodeList) []string {
```
spadgett, line 605: Is this going to trigger unnecessary rollouts if the order of the nodes changes (since the order in the arch potentially changes too)? When does this get evaluated and how we decide when to rollout a new console version?

jhadvig: @spadgett good catch. Updated the PR to address the node informer and also sort the `nodesArch` array so it wont get triggered if the order changes.

spadgett:
@Prashanth684 To determine the available architectures, we're listing nodes and looking at their architecture labels. Is this the best way to do it?

jhadvig:
@spadgett @Prashanth684 we discussed this change with @jaypoulz and we align on fetching all the nodes, no matter the role, and create the arch list form them.

jaypoulz:
I had a conversation about this with @joelanford and we felt like it was best to make no assumptions about which nodes operators would schedule their workloads on since operators can set any affinity and taint rules they like.

pkg/console/subresource/consoleserver/types.go
```diff
@@ -65,6 +65,7 @@ type ClusterInfo struct {
 	MasterPublicURL     string                `yaml:"masterPublicURL,omitempty"`
 	ControlPlaneToplogy configv1.TopologyMode `yaml:"controlPlaneTopology,omitempty"`
 	ReleaseVersion      string                `yaml:"releaseVersion,omitempty"`
+	NodesArchitecture   []string              `yaml:"nodesArchitecture,omitempty"`
```
spadgett, line 68: nit: I would call this `nodeArchitectures` (here and other places in the PR).

spadgett (commented):
lgtm, just one nit on the name

spadgett:
@joelanford Previously, we were filtering the OLM catalog on both architecture *and* OS. Do we need to look at the node OS here as well?

cc @bparees

spadgett:
https://docs.openshift.com/container-platform/4.11/operators/operator_sdk/osdk-generating-csvs.html#olm-enabling-operator-for-multi-arch_osdk-generating-csvs

bparees:
> @joelanford Previously, we were filtering the OLM catalog on both architecture and OS. Do we need to look at the node OS here as well?

i would think yes though for our first use case we're not going to have anything other than linux nodes being targeted i don't think.

Prashanth684:
> @Prashanth684 To determine the available architectures, we're listing nodes and looking at their architecture labels. Is this the best way to do it?

yes, as of today that is the only way to determine what architectures are available in a cluster.

jhadvig:

jhadvig:
QE Approver: 
/assign @yapei
Docs Approver:
 /assign @opayne1
PX Approver:
 /assign @RickJWagner

pkg/console/operator/sync_v400.go
```diff
@@ -593,3 +602,15 @@ func (co *consoleOperator) GetAvailablePlugins(enabledPluginsNames []string) []*
...
+func getNodeArchitectures(nodes *corev1.NodeList) []string {
+	nodeArchitectures := []string{}
+	for _, node := range nodes.Items {
+		nodeArch := node.Labels[api.NodeArchitectureLabel]
+		if !util.Contains(nodeArchitectures, nodeArch) {
+			nodeArchitectures = append(nodeArchitectures, nodeArch)
+		}
+	}
+	sort.Strings(nodeArchitectures)
+	return nodeArchitectures
```
joelanford, line 615: Nit: It might be slightly more efficient to use [`sets.String`](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#String) here.

```suggestion
	nodeArchitectures := sets.NewString()
	for _, node := range nodes.Items {
		nodeArch := node.Labels[api.NodeArchitectureLabel]
		nodeArchitectures.Insert(nodeArch)
	}
	return nodeArchitectures.List()
```

joelanford (commented):
> @joelanford Previously, we were filtering the OLM catalog on both architecture and OS. Do we need to look at the node OS here as well?


@spadgett I think I'd re-iterate what @bparees said, but just double checking my understanding, which is that today Console assumes "cluster is homogenous," therefore "the set of supported os/arch pairs in the cluster is equal to console process's os/arch", right?

So yeah, instead of making that assumption, I think the OLM catalog filter should be based on querying the node labels.

One question: do we not need to specifically also consider the node OS label? Technically, OLM CSVs support specifying the supported OS.

As an aside, the OLM annotations have always felt a bit off because they require the Arch and OS to be specified separately. Specifically, this example that's actually documented on the page you linked highlights my concern:
```yaml
labels:
  operatorframework.io/arch.s390x: supported
  operatorframework.io/os.zos: supported
  operatorframework.io/os.linux: supported 
  operatorframework.io/arch.amd64: supported 
```

The machine interpretation of that is likely that the following os/arch pairs are supported:
- linux/amd64
- linux/s390x
- zos/amd64
- zos/s390x

A human's (e.g. an operator author's) interpretation could easily be:
- linux/amd64
- zos/s390x

Does Console look at the ` operatorframework.io/os*` CSV annotations today?

jhadvig:
@joelanford thanks for the comment.

> One question: do we not need to specifically also consider the node OS label? Technically, OLM CSVs support specifying the supported OS.

My understanding, based on the discussion that I've had with @jaypoulz, was that the first pass of the feature would only consider filtering based on the Arch. From the OS point of view the cluster should be homogenous. Is this assumption correct?
The [follow up frontend story](https://issues.redhat.com/browse/CONSOLE-3190) should then filter the operator's PackageManifests based on the supported architectures.

Eventually we aim to add the affinity editor, but that part is still to be discussed.

bparees:
> From the OS point of view the cluster should be homogenous. Is this assumption correct?

@Prashanth684 can you answer/confirm this?  tbh i'm a little surprised to see s390 as an OS option since even when running on Z, CoreOS VMs are still used to run the nodes aiui, so effectively the OS is still linux.

spadgett:
> @spadgett I think I'd re-iterate what @bparees said, but just double checking my understanding, which is that today Console assumes "cluster is homogenous," therefore "the set of supported os/arch pairs in the cluster is equal to console process's os/arch", right?

@joelanford That's right.

> Does Console look at the ` operatorframework.io/os*` CSV annotations today?

Yes, we look at both arch and os labels today. Currently, console gets the GOARCH and GOOS values from the environment of the pod it's running in, then filters the OLM catalog for operators that support those values. 

https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-items.tsx#L46-L86

joelanford:
> tbh i'm a little surprised to see s390 as an OS option since even when running on Z, CoreOS VMs are still used to run the nodes aiui, so effectively the OS is still linux.

I was also a little surprised by that example. But I'm also curious about Windows. Don't we have some level of support for Windows nodes/containers? https://www.redhat.com/en/technologies/cloud-computing/openshift/windows-containers-on-red-hat-openshift

> As an aside, the OLM annotations have always felt a bit off because they require the Arch and OS to be specified separately

@bparees this is probably something we can fix with the PO/rukpak/new bundle format stuff. We probably need to let operator authors specify individual os/arch pairs if we think there's going to ever be another OS in the mix.

jhadvig:
@spadgett @joelanford Im a bit confused here. Are there any additional changes needed ?

joelanford:
I have nothing blocking from my side. I'm interested in my question about windows nodes, but it seems reasonable to me to cover that separately if necessary.

pkg/console/operator/sync_v400.go
```diff
@@ -593,3 +601,12 @@ func (co *consoleOperator) GetAvailablePlugins(enabledPluginsNames []string) []*
...
 	return availablePlugins
 }
+
+func getNodeArchitectures(nodes *corev1.NodeList) []string {
+	nodeArchitectures := sets.NewString()
+	for _, node := range nodes.Items {
+		nodeArch := node.Labels[api.NodeArchitectureLabel]
+		nodeArchitectures.Insert(nodeArch)
+	}
+	return nodeArchitectures.List()
```
spadgett, line 611: Will this return the architectures in the same order every time? If not, will it introduce unnecessary rollouts of console?

spadgett:
@jhadvig I wouldn't block this PR on adding OS based on the discussion here, but let's open a follow-up issue to track so we don't forget.

pkg/console/operator/sync_v400.go
jhadvig: sorry, somehow the sorting was removed in my last update... its back now 👍

joelanford: `List` is already ordered, fwiw. https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#String.List

Prashanth684:
lgtm from me

pkg/console/operator/sync_v400.go
```diff
@@ -593,3 +602,14 @@ func (co *consoleOperator) GetAvailablePlugins(enabledPluginsNames []string) []*
...
 }
+
+func getNodeArchitectures(nodes *corev1.NodeList) []string {
+	nodeArchitecturesSet := sets.NewString()
+	for _, node := range nodes.Items {
+		nodeArch := node.Labels[api.NodeArchitectureLabel]
+		nodeArchitecturesSet.Insert(nodeArch)
+	}
+	nodeArchitectures := nodeArchitecturesSet.List()
+	sort.Strings(nodeArchitectures)
```
spadgett, line 613: @jhadvig Per @joelanford this can be removed

https://github.com/openshift/console-operator/pull/669#discussion_r951778015

Sorry for the noise

jhadvig: @spadgett np. Updated :)

spadgett (approved):
/lgtm

jhadvig:
/retest

```
  Logging in as kubeadmin
      Installing operator: "Red Hat OpenShift Pipelines"
      Operator Red Hat OpenShift Pipelines was not yet installed.
      Performing Pipelines post-installation steps
      Verify the CRD's for the "Red Hat OpenShift Pipelines"
{"component":"entrypoint","file":"k8s.io/test-infra/prow/entrypoint/run.go:165","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Process did not finish before 2h0m0s timeout","severity":"error","time":"2022-08-25T22:36:54Z"}
{"component":"entrypoint","file":"k8s.io/test-infra/prow/entrypoint/run.go:255","func":"k8s.io/test-infra/prow/entrypoint.gracefullyTerminate","level":"error","msg":"Process did not exit before 15s grace period","severity":"error","time":"2022-08-25T22:37:09Z"}
{"component":"entrypoint","error":"process timed out","file":"k8s.io/test-infra/prow/entrypoint/run.go:80","func":"k8s.io/test-infra/prow/entrypoint.Options.Run","level":"error","msg":"Error executing test process","severity":"error","time":"2022-08-25T22:37:09Z"}
error: failed to execute wrapped command: exit status 127 
INFO[2022-08-25T22:37:13Z] Step e2e-aws-console-test failed after 2h2m20s. 
INFO[2022-08-25T22:37:13Z] Step phase test failed after 2h2m20s.        
INFO[2022-08-25T22:37:13Z] Running multi-stage phase post  
```

pkg/console/operator/sync_v400.go
```diff
@@ -367,6 +368,12 @@ func (co *consoleOperator) SyncConfigMap(
 		return nil, false, "FailedGetManagedConfig", mcErr
 	}
 
+	nodeList, nodeListErr := co.nodeClient.Nodes().List(ctx, metav1.ListOptions{})
+	if nodeListErr != nil {
+		return nil, false, "FailedListNodes", nodeListErr
+	}
+	nodeArchitectures := getNodeArchitectures(nodeList)
```
Prashanth684, line 375: how often will this be updated? anytime there is an additional node added to the system ?

PR #672, opened 2022-08-15
Bug 2117646: Changing  field on any of routes in the openshift-console namespace wont trigger sync loop

jhadvig:
/assign @TheRealJon

### DIFF
--- a/test/e2e/util.go
+++ b/test/e2e/util.go
@@ -71,20 +71,20 @@ func patchAndCheckService(t *testing.T, client *framework.ClientSet, isOperatorM
 }
 
 func patchAndCheckRoute(t *testing.T, client *framework.ClientSet, isOperatorManaged bool) error {
-	t.Logf("patching TargetPort on the console Route")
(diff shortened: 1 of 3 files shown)


TheRealJon:
Do we need to update or add any tests for this?

TheRealJon (approved):
Just the one question about whether or not we need to update or add any tests along with this change.

/lgtm

jhadvig:
@TheRealJon thanks for comment. Updated the existing test. PTAL

test/e2e/util.go
```diff
@@ -72,7 +72,7 @@ func patchAndCheckService(t *testing.T, client *framework.ClientSet, isOperatorM
 
 func patchAndCheckRoute(t *testing.T, client *framework.ClientSet, isOperatorManaged bool) error {
 	t.Logf("patching TargetPort on the console Route")
```
TheRealJon, line 74: Does this log statement need to be updated as well?

jhadvig: Yes, you are right. Im blind 🙈

PR #677, opened 2022-08-27
CONSOLE-3125: Add cluster filtering by product and ocp version

zherman0:
Add ability to remove clusters from cluster list based on several criteria:
1) ~~Remove all products with *KS values:  "AKS", "EKS", "IKS", "ROKS".  Instead of using regexp of *KS, a list was made based on product name so non-KS values can be added in the future~~
UPDATE: ROKS is allowed.  Switching logic to an allow list of: OpenShift, ROSA, ARO, ROKS, and OpenShiftDedicated
2) Remove all ocp versions less than 4.0
3) This solution does not offer the option of having the console show things as disable, instead, it works with existing filters and only includes managed clusters that pass the criteria.  To get more options on the dropdown, we probably would need to pass the entire set of managedClusters to console and then filter on the frontend side.

### DIFF
--- a/pkg/console/controllers/managedclusters/controller.go
+++ b/pkg/console/controllers/managedclusters/controller.go
@@ -226,10 +226,39 @@ func (c *ManagedClusterController) SyncManagedClusterList(ctx context.Context) (
 			continue
 		}
 
+		// Check the claims if version and product are supported
+		validProduct, validVersion := isSupportedCluster(managedCluster.Status.ClusterClaims)
+		// Omit clusters that have unsupported product name defined in api UnsupportedClusterProducts
+		if !validProduct {
+			klog.V(4).Infof("Skipping managed cluster %q, product is unsupported", clusterName)
+			continue
+		}
+
+		// Omit any clusters with version less than ocp 4.0.0
+		if !validVersion {
+			klog.V(4).Infof("Skipping managed cluster %q, versions prior to openshift 4.0.0 are not supported", clusterName)
+			continue
+		}
 		valid = append(valid, managedCluster)
 	}
 
 	return valid, "", nil
+}
+
+func isSupportedCluster(clusterClaims []clusterv1.ManagedClusterClaim) (bool, bool) {
+	isValidProduct := false
+	isValidVersion := false
+	for _, claim := range clusterClaims {
+		if claim.Name == api.ManagedClusterClaimProductAnnotation && util.SliceContains(api.SupportedClusterProducts, claim.Value) {
+			isValidProduct = true
+			continue
+		}
+		if claim.Name == api.ManagedClusterClaimVersionAnnotation && util.IsSupportedVersion(claim.Value) {
+			isValidVersion = true
+			continue
+		}
+	}
+	return isValidProduct, isValidVersion
 }
 
 func (c *ManagedClusterController) SyncOAuthClientManagedClusterViews(ctx context.Context, operatorConfig *operatorv1.Console, managedClusters []clusterv1.ManagedCluster) ([]*unstructured.Unstructured, string, error) {
--- a/pkg/console/controllers/managedclusters/controller_test.go
+++ b/pkg/console/controllers/managedclusters/controller_test.go
@@ -1,0 +1,84 @@
+package managedcluster
+
+import (
+	"testing"
+
+	"github.com/go-test/deep"
+
+	v1 "github.com/open-cluster-management/api/cluster/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func generateClusters(t *testing.T) []v1.ManagedCluster {
+
+	clusters := []v1.ManagedCluster{}
+
+	cluster1 := v1.ManagedCluster{}
+	cluster1.ObjectMeta = metav1.ObjectMeta{Name: "ProductAKSVer412"}
+	cluster1.Status = v1.ManagedClusterStatus{
+		ClusterClaims: []v1.ManagedClusterClaim{
+			{Name: "product.open-cluster-management.io", Value: "AKS"},
+			{Name: "version.openshift.io", Value: "4.12.0-ec.1"},
+		},
+	}
+
+	clusters = append(clusters, cluster1)
+
+	cluster2 := v1.ManagedCluster{}
+	cluster2.ObjectMeta = metav1.ObjectMeta{Name: "ProductOpenshiftVer49"}
+	cluster2.Status = v1.ManagedClusterStatus{
+		ClusterClaims: []v1.ManagedClusterClaim{
+			{Name: "product.open-cluster-management.io", Value: "OpenShift"},
+			{Name: "version.openshift.io", Value: "4.9.0-ec.1"},
+		},
+	}
+	clusters = append(clusters, cluster2)
+
+	cluster3 := v1.ManagedCluster{}
+	cluster3.ObjectMeta = metav1.ObjectMeta{Name: "ProductOpenshiftVer39"}
+	cluster3.Status = v1.ManagedClusterStatus{
+		ClusterClaims: []v1.ManagedClusterClaim{
+			{Name: "product.open-cluster-management.io", Value: "OpenShift"},
+			{Name: "version.openshift.io", Value: "3.9"},
+		},
+	}
+	clusters = append(clusters, cluster3)
+
+	cluster4 := v1.ManagedCluster{}
+	cluster4.ObjectMeta = metav1.ObjectMeta{Name: "ProductEKSVer311"}
+	cluster4.Status = v1.ManagedClusterStatus{
+		ClusterClaims: []v1.ManagedClusterClaim{
+			{Name: "product.open-cluster-management.io", Value: "EKS"},
+			{Name: "version.openshift.io", Value: "3.11.rcbeata"},
+		},
+	}
+	clusters = append(clusters, cluster4)
+
+	return clusters
+}
+
+func TestProductAndVersionCheckForClusterList(t *testing.T) {
+	expectedCluster := []string{"ProductOpenshiftVer49"}
+	clusters := generateClusters(t)
+	validClusters := []string{}
+	if len(clusters) < 1 {
+		t.Fatalf("No clusters are defined")
+	}
+	for _, managedCluster := range clusters {
+		clusterName := managedCluster.GetName()
+		validProduct, validVersion := isSupportedCluster(managedCluster.Status.ClusterClaims)
+		if !validProduct {
+			t.Logf("Skipping managed cluster %q, product is unsupported", clusterName)
+			continue
+		}
+		if !validVersion {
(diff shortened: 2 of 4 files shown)


KevinFCormier (approved):

jhadvig (changes requested):
@zherman0 thanks for the PR.
Adding couple of comments. Also I would recommend to wrap the added logic from the controller into a function so its easier to test + add a unit test for it the wrapper function.

pkg/console/controllers/util/util.go
```diff
@@ -37,3 +41,26 @@ func LabelFilter(labels map[string]string) factory.EventFilterFunc {
...
+	for _, v := range s {
+		if v == value {
+			return true
+		}
+	}
+	return false
+}
+
+//Return map of ClusterClaims on a managedCluster resource
+type clusterClaimsMap = map[string]string
```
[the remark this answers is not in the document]
zherman0: This is just one of the conventions that is used with go.  I could argue for or against but I am fine simplifying this.

pkg/console/controllers/util/util.go
```diff
@@ -37,3 +41,26 @@ func LabelFilter(labels map[string]string) factory.EventFilterFunc {
 		return true
 	}
 }
+
+// contains checks if a string is present in a slice
+type sliceName []string
+
+func SliceContains(s sliceName, value string) bool {
```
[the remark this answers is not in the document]
zherman0: Just a go convention, I will simplify.

pkg/console/controllers/managedclusters/controller.go
```diff
@@ -226,6 +226,21 @@ func (c *ManagedClusterController) SyncManagedClusterList(ctx context.Context) (
...
+
+		// Omit clusters that have the following product name
+		omittedProductCluster := []string{"AKS", "EKS", "IKS", "ROKS"}
+		if s, ok := claims["product.open-cluster-management.io"]; ok && util.SliceContains(omittedProductCluster, s) {
+			klog.V(4).Infof("Skipping managed cluster %v, based on vendor label, cluster unsupported", clusterName)
+			continue
+		}
+
+		// Omit any clusters with version less than ocp 4.0
+		if s, ok := claims["version.openshift.io"]; ok && s < "4.0" {
```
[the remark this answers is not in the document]
zherman0: I did test this logic and because the string "3" is < the string "4" is does work.  However I will look at your alternative since it does not really make since to use math operators on a string.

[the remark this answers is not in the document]
zherman0: So researching this, both methods are valid and both use the lexical order so they are essentially the same AFAIK.  I will leave the code as is for readability, if that works for you.

pkg/console/controllers/managedclusters/controller_test.go
```diff
@@ -0,0 +1,122 @@
...
+			continue
+		}
+		validClusters = append(validClusters, clusterName)
+	}
+
+	if diff := deep.Equal(validClusters, expectedCluster); diff != nil {
+		t.Error(diff)
+		t.Fatalf("\n---------------\nValid clusters are %v, while expected clusters are %v\n-----------", validClusters, expectedCluster)
+	}
+	t.Logf(">>>> Numer of Valid Clusters %v------- EOT\n", validClusters)
```
jhadvig, line 93: Why is this log needed?

pkg/console/controllers/managedclusters/controller_test.go
```diff
@@ -0,0 +1,122 @@
...
+			t.Logf("Skipping managed cluster %q with version %q, versions prior to openshift 4.0 not supported", clusterName, claims["version.openshift.io"])
+			continue
+		}
+		validClusters = append(validClusters, clusterName)
+	}
+	if diff := deep.Equal(validClusters, expectedCluster); diff != nil {
+		t.Error(diff)
+		t.Fatalf("\n---------------\nValid clusters count was %v\n-----------", len(validClusters))
+	}
+	t.Logf(">>>> Numer of Valid Clusters %v------- EOT\n", validClusters)
```
jhadvig, line 120: Why is this log needed?

pkg/console/controllers/managedclusters/controller_test.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func TestVersionCheckForClusterList(t *testing.T) {
+	expectedCluster := []string{}
+	expectedCluster = append(expectedCluster, "ProductAKSVer412")
+	expectedCluster = append(expectedCluster, "ProductOpenshiftVer49")
+	clusters := generateClusters(t)
+	validClusters := []string{}
+	if len(clusters) < 1 {
+		t.Fatalf("\n---------------\nNo Valid clusters, count was %v\n-----------", len(clusters))
```
jhadvig, line 104: Please make the error more propitiate

pkg/console/controllers/managedclusters/controller.go
```diff
@@ -226,12 +226,37 @@ func (c *ManagedClusterController) SyncManagedClusterList(ctx context.Context) (
...
+		}
 		valid = append(valid, managedCluster)
 	}
 
 	return valid, "", nil
 }
 
+func VersionCheckForClusterList(claims map[string]string) bool {
+	s, ok := claims["version.openshift.io"]
+	return ok && s < "4.0"
```
jhadvig, line 252: Again, I would rather use a proper solution for comparing version, when plane string coparision.
This package we already vendor in the operator https://github.com/openshift/console-operator/blob/master/go.mod#L7

pkg/console/controllers/managedclusters/controller_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+		if VersionCheckForClusterList(claims) {
+			t.Logf("Skipping managed cluster %q with version %q, versions prior to openshift 4.0 not supported", clusterName, claims["version.openshift.io"])
+			continue
+		}
+		validClusters = append(validClusters, clusterName)
+	}
+	if diff := deep.Equal(validClusters, expectedCluster); diff != nil {
+		t.Error(diff)
+		t.Fatalf("\n---------------\nValid clusters count was %v\n-----------", len(validClusters))
```
jhadvig, line 118: Please make the error more propitiate

pkg/console/controllers/managedclusters/controller_test.go
```diff
@@ -0,0 +1,122 @@
...
+		claims := util.ClaimsMap(&managedCluster)
+
+		if VersionCheckForClusterList(claims) {
+			t.Logf("Skipping managed cluster %q with version %q, versions prior to openshift 4.0 not supported", clusterName, claims["version.openshift.io"])
+			continue
+		}
+		validClusters = append(validClusters, clusterName)
+	}
+	if diff := deep.Equal(validClusters, expectedCluster); diff != nil {
+		t.Error(diff)
```
jhadvig, line 117: When this function is called the program will fail, so no other calls will get executed.

pkg/console/controllers/managedclusters/controller_test.go
```diff
@@ -0,0 +1,122 @@
...
+		omittedProductCluster := []string{"AKS", "EKS", "IKS", "ROKS"}
+		if ProductCheckForClusterList(claims, omittedProductCluster) {
+			t.Logf(">>>>> Skipping managed cluster %q, product %q is unsupported", clusterName, claims["product.open-cluster-management.io"])
+			continue
+		}
+		validClusters = append(validClusters, clusterName)
+	}
+
+	if diff := deep.Equal(validClusters, expectedCluster); diff != nil {
+		t.Error(diff)
```
jhadvig, line 90: When this function is called the program will fail, so no other calls will get executed.

jhadvig (changes requested):
Thanks @zherman0, adding some additional comments.
I would advice to have just a single function in the `controller.go` that would take the `managedCluster` and check if its supported or not. That way it will be easier to test it as a unit, rather then writing the whole wrapped logic again in the tests.
Also would advice work with the Claim object, then flattening it.

pkg/console/controllers/managedclusters/controller.go
zherman0: ok, switched to Compare function

zherman0:
@jhadvig - made your suggested changes; they were all very good points.  I did keep the logic as 2 functions since they are different in what they are checking even though we are using the claims object.  Also, it is more readable IMO in the controller especially since the other checks in there go incrementally through the client config.
Not sure what you mean to use "the Claim object, then flattening it", but we can discuss further. Thanks.

jhadvig (commented):
@zherman0 thanks for the update.
Couple of additional comments.

> I did keep the logic as 2 functions since they are different in what they are checking even though we are using the claims object. Also, it is more readable IMO in the controller especially since the other checks in there go incrementally through the client config.

I dont think thats true since we are validating the same cluster, based on two of his claims.

```go
// in controller.go
func isSupportedCluster(clusterClaims []clusterv1.ManagedClusterClaim) bool {
	isValidProduct := false
	isValidVersion := false
	for _, claim := range clusterClaims {
		if claim.Name == api.ManagedClusterClaimProductAnnotation && util.SliceContains(api.UnsupportedProducts, claim.Value) {
			isValidProduct = true
			continue
		}
		if claim.Name == api.ManagedClusterClaimVersionAnnotation && util.IsSupportedVersion(claim.Value) {
			isValidVersion = true
			continue
		}
	}
	return isValidProduct && isValidVersion
}

// in utils pkg
// moving here makes sense since comparing version is more generic task and might be reused
func IsSupportedVersion(productVersion string) bool {
	version, err := semver.Parse(productVersion)
	if err != nil {
		klog.Log("unable to parse %q version", productVersion)
		return false
	}
	return version.Compare(semver.MustParse("4.0.0")) == 1
}
```

This way we only need to test the `isClusterSupported()` as part of the unit test.

jhadvig (approved):
Changes look good! Thanks @zherman0 :)
/lgtm
/approve

jhadvig:
QE Approver:
 /assign @yapei
Docs Approver:
 /assign @opayne1
PX Approver:
 /assign @RickJWagner

zherman0:
@jhadvig @spadgett - After talking about qe testing yesterday, the point was brought that our unsupported list was not correct so I change the code to use a "Supported" list instead that covers the agreed upon values OpenShift, ROSA, ARO, ROKS, and  OpenShiftDedicated.  Please take another look and I will squash if it looks good.  Thanks,

yapei:
> OpenShift, ROSA, ARO, ROKS, and OpenShiftDedicated

@zherman0 so are these the final lists of supported products?

zherman0:
@yapei - That is the final list as of yesterday.  It is basically the same as before with the exception of the ROKS now being allowed. Also, the logic has changed to look for things in the supported list instead of excluded things in the unsupported list.

jhadvig (approved):
/lgtm

Thanks @zherman0

yapei:
approving the changes since currently CI is passing
/label qe-approved

PR #691, opened 2022-10-24
CONSOLE-3337: Dynamic plugin proxy reference issue

jhadvig:
We should be using an array entry reference rather then a memory address, while casting the object into a variable.

Also adding couple of unit tests to back case where there are multiple proxy services defined in a single plugin.

/assign @TheRealJon


TheRealJon (approved):
/lgtm

spadgett (approved):
/lgtm

PR #693, opened 2022-10-27
CONSOLE-3286: Set disabledCopiedCSVs in console

zherman0:
Console-Operator: Set disabledCopiedCSVs for console using console config
Also added RBAC commands so we can access the OLM config file

https://issues.redhat.com/browse/CONSOLE-3286

### DIFF
--- a/manifests/03-rbac-role-cluster.yaml
+++ b/manifests/03-rbac-role-cluster.yaml
@@ -9,13 +9,13 @@ metadata:
     capability.openshift.io/name: Console
 rules:
   - apiGroups:
-    - ""
+      - ""
     resources:
-    - nodes
+      - nodes
     verbs:
-    - get
-    - list
-    - watch
+      - get
+      - list
+      - watch
   - apiGroups:
       - oauth.openshift.io
     resources:
@@ -82,6 +82,14 @@ rules:
       - cluster.open-cluster-management.io
     resources:
       - managedclusters
+    verbs:
+      - get
+      - list
+      - watch
+  - apiGroups:
+      - operators.coreos.com
+    resources:
+      - olmconfigs
     verbs:
       - get
       - list
@@ -183,15 +191,15 @@ metadata:
     rbac.authorization.k8s.io/aggregate-to-edit: "true"
     rbac.authorization.k8s.io/aggregate-to-admin: "true"
 rules:
-- apiGroups:
-  - helm.openshift.io
-  resources:
-  - projecthelmchartrepositories
-  verbs:
-  - get
-  - list
-  - update
-  - create
-  - watch 
-  - patch
-  - delete
+  - apiGroups:
+      - helm.openshift.io
+    resources:
+      - projecthelmchartrepositories
+    verbs:
+      - get
+      - list
+      - update
+      - create
+      - watch
+      - patch
+      - delete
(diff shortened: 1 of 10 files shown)


pkg/console/subresource/consoleserver/types.go
```diff
@@ -68,6 +68,7 @@ type ClusterInfo struct {
 	ControlPlaneToplogy configv1.TopologyMode `yaml:"controlPlaneTopology,omitempty"`
 	ReleaseVersion      string                `yaml:"releaseVersion,omitempty"`
 	NodeArchitectures   []string              `yaml:"nodeArchitectures,omitempty"`
+	CopiedCSVsDisabled  bool                  `yaml:"copiedCSVsDisabled,omitempty"`
```
jhadvig, line 71: remove the omitempty here otherwise the yaml pkg whats Marshalling into yaml will drop the variable if its set to `false`

manifests/03-rbac-role-cluster.yaml
```diff
@@ -84,6 +84,14 @@ rules:
       - get
       - list
       - watch
+  - apiGroups:
```
jhadvig, line 87: Based on the error this role needs to be added to the console-operator in its namespace.

```
Err getting olmconfig olmconfigs.operators.coreos.com "cluster" is forbidden: User "system:serviceaccount:openshift-console-operator:console-operator" cannot get resource "olmconfigs" in API group "operators.coreos.com" at the cluster scope
```

jhadvig (changes requested):

jhadvig:
@zherman0 I've tracked down the issue, there have been several of them, + added several comments
Also I've sent you a fixing PR against your branch. PTAL

zherman0:
QE Approver
/assign @yapei

Docs Approver:
/assign @opayne1

PX Approver:
/assign @RickJWagner

pkg/console/subresource/consoleserver/types.go
zherman0: I would prefer it drops the variable if it is false.  I set it back to `omitempty` which fixes the testcases and this prevents the config from showing `disableCopiedCSVs: false`

jhadvig: Dont have any strong feelings either way. Lets keep it then.

jhadvig (changes requested):
Changes look good.
Lets remove the unnecessary  `Dockerfile.dev`and we are ready to tag 👍

jhadvig (approved):
/lgtm
/approve

zherman0:
@yapei or whomever test this, here is the procedure:
1) Deploy the PR
2) check console config:
```oc get cm console-config -n openshift-console -o yaml``` 
and verify disableCopiedCSVs is not present in the ClusterInfo section
3) run:
```oc apply -f - <<EOF                                                                            
apiVersion: operators.coreos.com/v1
kind: OLMConfig
metadata:
  name: cluster
spec:
  features:
    disableCopiedCSVs: true # Disabled Copied CSVs for AllNamespace operators.
EOF```
4) check console config:
```oc get cm console-config -n openshift-console -o yaml``` 
and verify disableCopiedCSVs is present in the ClusterInfo section and that it is set to true

XiyunZhao:
This PR has been tested on the private build by following the procedure shared by zheman0.
All below checkpoint on below is passed
1. User can find "copiedCSVsDisabled" is enabled in the Data section on ComfigMap of "console-config" after user update "copiedCSVsDisabled" to true on OLMConfig
2. The default value for the property of "DisabledCopiedCSVs" is False.
3. The status value of "DisabledCopiedCSVs" on the Conditions Section on the OLMConfig page is shown correctly
4. The newly added "olmconfigs" rules can be found on the ClusterRole and Role details page for console-operator
5. The rbac-role-cluster file format update isn't impacted the rule for "nodes" and helm

PR #701, opened 2022-12-06
CONSOLE-3355: Console operator should sync managed cluster 'copiedCSVsDisabled' flag to managed cluster config

jhadvig:
Story - https://issues.redhat.com/browse/CONSOLE-3355

Extending the `ManagedClusterController` functionality to set the  `ManagedClusterConfig` with the new `CopiedCSVsDisabled` bool that indicates if the cluster has copying of CSV enabled or disabled.

Originally I wanted to create a new controller that would either create a new CM with array of cluster names that have copying of CSV disabled, or even putting it to `console-config.yaml`, but after reading though the `ManagedClusterController` I've rather decided to extend `ManagedClusterConfig` with the new `CopiedCSVsDisabled`. The main motivation was that all the MCV and MCA need to be synced anyway, otherwise console wont work with the ACM setup, so haven't seen any added value in creating a new controller.

TC:
In multicluster setup, check if `managed-clusters` ConfigMap in the `openshift-console` NS, on the hub cluster, contains `copiedCSVsDisabled` entries for each managed cluster instance.

/assign @TheRealJon


TheRealJon (commented):
Just a couple of small tweaks. Also, this would be new, but should we have an informer for the ManagedClusterViews so that we resync on change?

pkg/console/controllers/advancedclustermanagement/managedclusters/controller.go
```diff
@@ -90,12 +90,15 @@ func NewManagedClusterController(
...
 	configV1Informers := configInformer.Config().V1()
+	managedClusterViewInformer := dynamicInformers.ForResource(api.ManagedClusterViewGroupVersionResource)
 
 	return factory.New().
 		WithFilteredEventsInformers( // configs
 			util.IncludeNamesFilter(api.ConfigResourceName),
 			configV1Informers.Consoles().Informer(),
 			operatorConfigInformer.Informer(),
+			managedClusterInformers.Informer(),
+			managedClusterViewInformer.Informer(),
```
TheRealJon, line 101: I don't think this will work as expected. The name filter on L93 wouldn't match any resources. We probably need a new `WithFilteredEventsInformers` call. I think I used a label filter in the past to make sure we are only watching the resources managed by console.

TheRealJon (approved):
/lgtm

jhadvig:
QE Approver: 
/assign @yapei
Docs Approver: 
/assign @opayne1
PX Approver:	
 /assign @RickJWagner

yapei:
verified that `copiedCSVsDisabled` flag is synced to cm/managed-clusters on hub cluster
/label qe-approved

TheRealJon (approved):
/lgtm

PR #705, opened 2022-12-12
OCPBUGS-4008: Add `app: console` label to managed-clusters configmap

TheRealJon:
Ensure main console sync loop is re-run when managed-cluster config changes


jhadvig (approved):
/lgtm

PR #719, opened 2023-01-16
CONSOLE-3331, OCPBUGS-2397: Migrate to ManifestWork API

TheRealJon:


pkg/console/controllers/advancedclustermanagement/managedclusteroauthclients/controller.go
```diff
@@ -0,0 +1,200 @@
...
+			localOAuthClient.RedirectURIs,
+		)
+		statusHandler.AddConditions(status.HandleProgressingOrDegraded(ConditionPrefix, reason, err))
+		if err != nil {
+			klog.V(4).Infof("Error syncing OAuthClient ManifestWork for managed cluster %s: %v./n", managedCluster.Name, err)
+			errs = append(errs, err)
+		}
+	}
+
+	err = utilerrors.NewAggregate(errs)
```
jhadvig, line 150: Lets rather move the `statusHandler.AddConditions` here and try to aggregate the `reason` as well, otherwise only the last failing sync will be surfaced by in the condition. We should surface all of them

pkg/console/subresource/manifestwork/oauth-client-work.go
```diff
@@ -0,0 +1,85 @@
...
+			},
+			Workload: workv1.ManifestsTemplate{
+				Manifests: []workv1.Manifest{},
+			},
+		},
+	}
+}
+
+func WithManifest(manifestWork *workv1.ManifestWork, object runtime.Object) {
+	if &manifestWork.Spec == nil {
```
jhadvig, line 49: dont think these checks are necessary since we are always using the stub from the `ManagedClusterOAuthClientManifestWorkStub` which should always contain all those fields

pkg/console/subresource/manifestwork/oauth-client-work.go
```diff
@@ -0,0 +1,85 @@
...
+
+	if &manifestWork.Spec.Workload.Manifests == nil {
+		manifestWork.Spec.Workload.Manifests = []workv1.Manifest{}
+	}
+
+	manifest := workv1.Manifest{RawExtension: runtime.RawExtension{Object: object}}
+	manifestWork.Spec.Workload.Manifests = append(manifestWork.Spec.Workload.Manifests, manifest)
+}
+
+func GetOAuthClientSecret(manifestWork *workv1.ManifestWork) (string, error) {
```
jhadvig, line 65: isnt this the same secret as the [local one](https://github.com/openshift/console-operator/blob/58ea3f2d1d678adb4c0424b345b91bf3fa24b051/pkg/console/controllers/advancedclustermanagement/managedclusteroauthclients/controller.go#L115)?

jhadvig (commented):
Thanks @TheRealJon the PR looks good overall, adding couple of comments

pkg/console/subresource/manifestwork/oauth-client-work.go
TheRealJon: I'll probably be removing this logic in preference to the workbuilder helper here: https://github.com/open-cluster-management-io/api/tree/main/utils/work/v1

pkg/console/subresource/manifestwork/oauth-client-work.go
TheRealJon: Yes, but in the future, we should generate unique secret strings for each managed cluster, and when we do, it would be nice not to have to update this logic.

jhadvig: makes sense 👍 
> we should generate unique secret strings for each managed cluster

Created a story for [tracking](https://issues.redhat.com/browse/CONSOLE-3429) the work on that.

TheRealJon:
@jhadvig I've addressed your comments. PTAL!

pkg/console/controllers/advancedclustermanagement/managedclusteroauthclients/controller.go
```diff
@@ -0,0 +1,199 @@
...
+	if err != nil || len(managedClusterList.Items) == 0 {
+		return c.remove(ctx)
+	}
+
+	// Create ManifestWorks which will create oauth clients on each managed cluster
+	errs := []error{}
+	for _, managedCluster := range managedClusterList.Items {
+		validCluster, validClusterVersion := util.IsSupportedMangedCluster(managedCluster)
+		if !validCluster || !validClusterVersion {
+			klog.V(4).Infof("[%s] Skipping OAuthClient ManifestWork, cluster is not supported.\n", managedCluster.Name)
```
jhadvig, line 132: Lets rather use the quotes
```suggestion
			klog.V(4).Infof("Skipping OAuthClient ManifestWork, %q cluster is not supported.\n", managedCluster.Name)
```

pkg/console/controllers/advancedclustermanagement/managedclusteroauthclients/controller.go
```diff
@@ -0,0 +1,199 @@
...
+	if err != nil {
+		return nil, "GetError", fmt.Errorf("Failed to get local OAuthClient: %v", err)
+	}
+	return oAuthClient, "", nil
+}
+
+func (c *Controller) SyncManagedClusters(ctx context.Context) (*clusterv1.ManagedClusterList, string, error) {
+	managedClusterList, err := c.managedClusterClient.ManagedClusters().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return managedClusterList, "ListError", err
```
jhadvig, line 164:
```suggestion
		return managedClusterList, "FailedListManagedClusters", err
```

pkg/console/subresource/manifestwork/helpers.go
```diff
@@ -0,0 +1,72 @@
...
+
+	if &manifestWork.Spec.Workload == nil {
+		manifestWork.Spec.Workload = workv1.ManifestsTemplate{}
+	}
+
+	if &manifestWork.Spec.Workload.Manifests == nil {
+		manifestWork.Spec.Workload.Manifests = []workv1.Manifest{}
+	}
+
+	manifestWork.Spec.Workload.Manifests = append(
```
jhadvig, line 27: Cant we just set the whole `manifestWork.Spec.Workload.Manifests` without those checks? since the stub is returning a object without any spec

jhadvig (commented):
Most of the changes look good. Adding couple of smaller comments.

pkg/console/subresource/manifestwork/helpers.go
TheRealJon: That's what the checks above are for. We're sequentially checking that all parent properties exist before we get to this line.

quickstarts/install-multicluster-engine.yaml
```diff
@@ -1,49 +0,0 @@
-apiVersion: console.openshift.io/v1
```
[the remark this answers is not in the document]
TheRealJon: I think this was a rebase error

pkg/console/subresource/oauthclient/apply.go
```diff
@@ -0,0 +1,51 @@
...
+	// at present, we only care about these two fields. this is NOT generic to all oauth clients
+	secretSame := equality.Semantic.DeepEqual(existing.Secret, required.Secret)
+	redirectsSame := equality.Semantic.DeepEqual(existing.RedirectURIs, required.RedirectURIs)
+	// nothing changed, so don't update
+	if secretSame && redirectsSame && !*modified {
+		// per ApplyService, etc, if nothing changed, return nil.
+		return nil, false, nil
+	}
+	existing.Secret = required.Secret
+	// existing.AdditionalSecrets = required.AdditionalSecrets
```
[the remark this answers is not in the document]
TheRealJon: This was just copied over from another file for housekeeping purposes. No changes were made other than the func name.

TheRealJon:
@jhadvig I've addressed your comments! PTAL!

pkg/console/controllers/advancedclustermanagement/managedclusteroauthclients/controller.go
```diff
@@ -0,0 +1,196 @@
...
+
+	configV1Informers := configInformer.Config().V1()
+
+	return factory.New().
+		WithFilteredEventsInformers( // configs
+			util.IncludeNamesFilter(api.ConfigResourceName),
+			configV1Informers.Consoles().Informer(),
+			operatorConfigInformer.Informer(),
+		).
+		ResyncEvery(1*time.Minute).
```
jhadvig, line 83: shouldnt we also add informer for the oauthClient ? Or we dont need to care if wheather it will get updated ?

pkg/console/controllers/util/util.go
```diff
@@ -52,11 +54,36 @@ func SliceContains(s []string, value string) bool {
...
+		}
+		if claim.Name == api.ManagedClusterClaimVersionAnnotation && IsSupportedManagedClusterVersion(claim.Value) {
+			isValidVersion = true
+			continue
+		}
+	}
+	return isValidProduct, isValidVersion
+}
+
+func IsConditionMet(conditions []metav1.Condition, conditionType string, conditionStatus metav1.ConditionStatus) bool {
```
jhadvig, line 82: WE already have `getClusterVersionCondition` if `/pkg/console/controllers/upgradenotification/controller.go`. Lets update it's params so we can use `IsConditionMet` instead

pkg/console/subresource/manifestwork/helpers.go
jhadvig: Right, but based on the API the both `Spec` and `Spec.Workload` are required fields, they for they will be present in the object no matter what.

jhadvig (commented):
Thanks @TheRealJon ! Adding a few more comments. PTAL

pkg/console/controllers/util/util.go
TheRealJon: I found an equivalent helper in the k8s apimachinery package: https://github.com/kubernetes/apimachinery/blob/6ecd1c8969025104b032f6a7f14ae93afc50e107/pkg/api/meta/conditions.go#L95

TheRealJon:
@jhadvig I've addressed the latest feedback! PTAL!

jhadvig (approved):
/lgtm
/approve
/retest

TheRealJon:
QE Approver
/assign @yapei

Docs Approver:
/assign @opayne1

PX Approver:
/assign @RickJWagner

TheRealJon:
@yapei Could you take a look at this when you get a chance? Let me know if you have questions!

pkg/api/api.go
```diff
@@ -11,84 +11,80 @@ const (
...
+	OpenShiftConsoleDeploymentName          = OpenShiftConsoleName
+	OpenShiftConsoleDownloadsDeploymentName = DownloadsResourceName
+	OpenShiftConsoleDownloadsPDBName        = DownloadsResourceName
+	OpenShiftConsoleDownloadsRouteName      = DownloadsResourceName
+	OpenShiftConsoleNamespace               = TargetNamespace
+	OpenShiftConsolePDBName                 = OpenShiftConsoleName
+	OpenShiftConsoleRouteName               = OpenShiftConsoleName
+	OpenShiftConsoleServiceName             = OpenShiftConsoleName
+	RedirectContainerTargetPort             = RedirectContainerPort
 )
```
TheRealJon, line 88: Alphabetized these constants 👀

yapei:
did some regression testing about multi cluster features, no issues found
/label qe-approved

PR #722, opened 2023-01-19
[WIP] Pause the MCO during proxy test to prevent nodes from applying bogus proxy config

jkyros:
MCO will now sync `controllerConfig` (which includes proxy config) even when a daemon rollout is failing. This results in nodes actually trying to reboot into the fake proxy config applied during the console-operator's e2e proxy test, which doesn't work (the kubelet can't start properly). 

This PR uses the MCO api to attempt to pause the MCO's pools during the proxy test, so it doesn't get rolled out before it can be reverted. 

Possible fix to:[OCPBUGS-5780]( https://issues.redhat.com/browse/OCPBUGS-5780)


jkyros:
looks like the MCO didn't break the cluster, that might be progress. Let's try again and see if it's flakes. 
/test e2e-aws-operator

jkyros:
That was promising, let's see if we can repeat our luck. 
/test e2e-aws-operator

jkyros:
Yeah, that last one the proxy config definitely still made it through. 

The MCO goes  `cluster/proxy` -> `renderConfig` -> `controllerConfig` -> `templates` -> `machineConfig` ->  `pool` -> `node` and the only place we can pause is at the pool level, we can't pause any of the previous stages, and depending on timing either the initial processing or the "revert" could be in-flight at one of the previous stages, so I think we're unpausing too early. 

We (the MCO) know we need more "knobs and dials" to address this, but we don't have them yet. 

Anyway, I've made it so we wait for the machineconfigs to actually render, get blocked by pause, and revert, so we should survive the race now. 

It seems like sometimes we do die to: 
```
proxy_test.go:34: operator has not reached settled state in 4m0s attempts due to [RouteHealthDegraded RouteHealthAvailable] - timed out waiting for the condition
```
but that looks unrelated to anything the MCO is doing.

jkyros:
Yep, got another:
```
proxy_test.go:34: operator has not reached settled state in 4m0s attempts due to [RouteHealthDegraded SyncLoopRefreshProgressing RouteHealthAvailable] - timed out waiting for the condition 
```
but I don't think I can fix that. 

There are other operators that react to that bogus proxy configuration too (and I'm totally not stopping them), so it wouldn't surprise me if something was going on.

PR #734, opened 2023-02-23
OCPBUGS-2783: Distinguish between route conditions and remove the old ones

jhadvig:
This issue was introduce by https://github.com/openshift/console-operator/pull/638/files where we only delivered the fix into the 4.9 release and when upgrading to 4.10 the new conditions havent been cleaned out by the `staleConditionsController`.
Due to that issue we will need to backport to release-4.10

/assign @TheRealJon @stlaz 

ccing @palonsoro


jhadvig:
-retest

TheRealJon (approved):
/lgtm

PR #738, opened 2023-03-14
OCPBUGS-10108: ART image update

jhadvig:
Superseds https://github.com/openshift/console-operator/pull/737


spadgett:
> /cherry-pick release-4.13

We'll want to make sure we're using the 4.13 base images (and the correct Go version) when we backport.

Dockerfile.rhel7
```diff
@@ -1,10 +1,10 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.18-openshift-4.12 AS builder
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.19-openshift-4.14 AS builder
```
spadgett, line 1: Do we need to update the regular Dockerfile as well now?

spadgett:
@jhadvig @TheRealJon I think we missed the important Dockerfile, which is the Dockerfile with no suffix in the master branch. It still references 4.13:

https://github.com/openshift/console/blob/master/Dockerfile

PR #741, opened 2023-03-30
OCPBUGS-11071: Operator should not add empty strings node arch list i…

TheRealJon:
…n console config

### DIFF
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -621,7 +621,9 @@ func getNodeArchitectures(nodes *corev1.NodeList) []string {
 	nodeArchitecturesSet := sets.NewString()
 	for _, node := range nodes.Items {
 		nodeArch := node.Labels[api.NodeArchitectureLabel]
-		nodeArchitecturesSet.Insert(nodeArch)
+		if nodeArch != "" {
+			nodeArchitecturesSet.Insert(nodeArch)
+		}
 	}
 	return nodeArchitecturesSet.List()
 }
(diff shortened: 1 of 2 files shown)


spadgett (commented):
Code LGTM, can we add a unit test?

pkg/console/operator/sync_v400.go
```diff
@@ -621,7 +621,9 @@ func getNodeArchitectures(nodes *corev1.NodeList) []string {
 	nodeArchitecturesSet := sets.NewString()
 	for _, node := range nodes.Items {
 		nodeArch := node.Labels[api.NodeArchitectureLabel]
-		nodeArchitecturesSet.Insert(nodeArch)
+		if nodeArch != "" {
+			nodeArchitecturesSet.Insert(nodeArch)
```
spadgett, line 625: Should we log a warning? I don't know if this is ever expected

TheRealJon:
@spadgett I've addressed your comments, PTAL!

pkg/console/operator/sync_v400_test.go
spadgett, line 1: Thanks, looks good. Can we add one more test where multiple nodes have the same arch label (to make sure it only appears in the list once)? This is the most common case.

This isn't new with your fix, but it's a gap in our tests.

spadgett (commented):
Thank you for adding tests 👍

TheRealJon:
/label tide/merge-method-squash

@spadgett I added the test case you requested, PTAL!

spadgett (approved):
/lgtm

jhadvig (approved):
/lgtm

PR #743, opened 2023-03-30
OCPBUGS-11164: Disable multicluster tech preview

TheRealJon:


TheRealJon:
/hold

Waiting on input as to whether we should also remove roles and rolebindings from the manifests.

spadgett (approved):
/lgtm

TheRealJon:
@spadgett I've updated to remove multicluster roles/bindings from the manifests and delete the code from starter.go rather than comment it out. PTAL when you get a chance!

spadgett (approved):
/lgtm

PR #757, opened 2023-04-24
OCPBUGS-12165: Proper cleanup of route sync conditions

jhadvig:
Accidentally I've set wrong conditional names for removal in https://issues.redhat.com/browse/OCPBUGS-2783
Conditiona name also needs to contain the type, which is in this case either `Degraded` or `Progressing`.

/assign @TheRealJon


spadgett (approved):
/lgtm

PR #761, opened 2023-05-02
OCPBUGS-12990: Update command line tools URL with custom downloads route

jhadvig:
When the custom downloads route is set we should be updating the ConsoleCLIDownloads link as well.

/assign @spadgett


TheRealJon (approved):
/lgtm

PR #765, opened 2023-05-23
update to consume featuregates via the API

deads2k:
This also brings client-go and openshift/api up to date.

I can't find the spot where https://github.com/openshift/console-operator/blob/cdc47ee0e8a18957cb10f9f9efeef1c5d9287b29/pkg/console/controllers/advancedclustermanagement/managedclusters/controller.go#L57 is initialized to update both the control loop and the caller at the same time.


jhadvig (commented):
@deads2k thanks for the PR. Is there any story/bug associated with it? so we have the background about this change

PR #763, opened 2023-05-16
OCPBUGS-11751: Remove multicluster conditions from console operator

TheRealJon:
Remove multicluster related conditions in 4.13 and 4.14 to prevent blocking updates.

PR #774, opened 2023-06-28
OCPBUGS-15499: Remove managed cluster logic from the operator's controller

jhadvig:
The issue with hotlooping operator was traced down to this [PR](https://github.com/openshift/console-operator/pull/763/files)
Found out that the sync_400 controller's `statusHandler` was still reporting the `ManagedClusterConfigSync` condition even though it got removed in the mentioned PR.
Due to that I've removed all the occurrences and also updated the tests.

### DIFF
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -75,10 +75,6 @@ func (co *consoleOperator) sync_v400(ctx context.Context, controllerContext fact
 		return statusHandler.FlushAndReturn(routeErr)
 	}
 
-	// managed-clusters ConfigMap is managed by another controller and is not required, we don't need to exit the sync loop if it's not present
-	managedClusterConfigMap, managedClusterConfigErrReason, managedClusterConfigErr := co.GetManagedClusterConfigMap(ctx)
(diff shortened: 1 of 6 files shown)


jerolimov (approved):
Lgtm! But since I'm not involved yet in the managed cluster logic, so I prefer that @spadgett or @TheRealJon takes also a look for an official lgtm.

But I can confirm that it fixes the issue on a cluster bot!

And for you both, I asked on slack why there are left overs in `api.go`:

```
	ManagedClusterConfigKey                   = "managed-clusters.yaml"
	ManagedClusterConfigMapName               = "managed-clusters"
```

and `SyncManagedClusterConfigMap` in `pkg/console/controllers/advancedclustermanagement/managedclusters/controller.go`, and others.

@jhadvig said this will be handled as part of [CONSOLE-3636](https://issues.redhat.com/browse/CONSOLE-3636)

jhadvig:
Yes this PR is removing just the bare minimum

spadgett (approved):
/lgtm

This change looks good to me, but we need to understand how this impacts #766 and what we need to backport to 4.13.z. It looks like we need to cherry-pick this change into #766.

PR #665, opened 2022-07-26
CONSOLE-3220: Add console capability to all manifests

raspbeep:
https://issues.redhat.com/browse/CONSOLE-3220

Signed-off-by: Pavel Kratochvil <pakratoc@redhat.com>

PR #784, opened 2023-08-03
chore: manifests/09-console-link-*: Drop obsolete deletion manifests

wking:
Since these manifests became deletion manifests in 6c6c5ce5f6 (#565, [4.9][1]), the only change has been 0f860bb07a (#665).  We can remove these deletion references now, because a 4.8 cluster that might have included these resources should have completed an update to 4.9 which would have removed them. And if that failed, they should have completed an update to one of the later 4.y and removed the resources. By removing the resource, we save the cluster-version operator some time checking to ensure deletion, and only expose ourselves to leaking the resources on clusters that updated from 4.8 through to 4.14 without ever having completed an update before reaching 4.14.

This change is very similar to openshift/machine-api-operator#1159

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1980531#c8


jhadvig (approved):
/lgtm
/approve

jhadvig:
Adding approvals manually, since the change is removing unused files.

PR #790, opened 2023-08-30
OCPBUGS-20129: Fix triggering `upgradenotification` in the cluster installation time

kevydotvinu:
- Check the `currentVersion` and `desiredVersion` variable before triggering the notification.

image


jhadvig (approved):
/lgtm
/approve

PR #795, opened 2023-09-23
OCPVE-719: feat: add support for olm capability

eggfoobar:
OLM is an optional component, adding conditional to remove informer on OLM resource

/hold

Need to find more context on the use of OLM resource in console and need to hold for API to be merged in https://github.com/openshift/api/pull/1589

### DIFF
--- a/pkg/console/operator/operator.go
+++ b/pkg/console/operator/operator.go
@@ -4,6 +4,7 @@ import (
 	// standard lib
 	"context"
 	"fmt"
+	"syscall"
 	"time"
 
 	// kube
@@ -11,6 +12,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/dynamic/dynamicinformer"
 	corev1 "k8s.io/client-go/informers/core/v1"
@@ -73,6 +75,9 @@ type consoleOperator struct {
 	consolePluginLister listerv1.ConsolePluginLister
 
 	resourceSyncer resourcesynccontroller.ResourceSyncer
+
+	// used to keep track of OLM capability
+	isOLMDisabled bool
 }
 
 func NewConsoleOperator(
@@ -142,18 +147,34 @@ func NewConsoleOperator(
 	configNameFilter := util.IncludeNamesFilter(api.ConfigResourceName)
 	targetNameFilter := util.IncludeNamesFilter(api.OpenShiftConsoleName)
 
-	olmConfigInformer := dynamicInformers.ForResource(schema.GroupVersionResource{Group: api.OLMConfigGroup, Version: api.OLMConfigVersion, Resource: api.OLMConfigResource})
+	informers := []factory.Informer{
+		configV1Informers.Consoles().Informer(),
+		operatorConfigInformer.Informer(),
+		configV1Informers.Infrastructures().Informer(),
+		configV1Informers.Ingresses().Informer(),
+		configV1Informers.Proxies().Informer(),
+		configV1Informers.OAuths().Informer(),
+	}
+
+	olmGroupVersionResource := schema.GroupVersionResource{
+		Group:    api.OLMConfigGroup,
+		Version:  api.OLMConfigVersion,
+		Resource: api.OLMConfigResource,
+	}
+
+	if found, _ := isResourceEnabled(dynamicClient, olmGroupVersionResource); found {
+		olmConfigInformer := dynamicInformers.ForResource(olmGroupVersionResource)
+		informers = append(informers, olmConfigInformer.Informer())
+	} else {
+		klog.Info("olmconfigs resource does not exist in cluster, launching poll and disabling olmconfigs informer")
+		c.isOLMDisabled = true
+		c.startPollAndRestartIfResourceEnabled(olmGroupVersionResource)
+	}
 
 	return factory.New().
 		WithFilteredEventsInformers( // configs
 			configNameFilter,
-			configV1Informers.Consoles().Informer(),
-			operatorConfigInformer.Informer(),
-			configV1Informers.Infrastructures().Informer(),
-			configV1Informers.Ingresses().Informer(),
-			configV1Informers.Proxies().Informer(),
-			configV1Informers.OAuths().Informer(),
-			olmConfigInformer.Informer(),
+			informers...,
 		).WithFilteredEventsInformers( // console resources
 		targetNameFilter,
 		deploymentInformer.Informer(),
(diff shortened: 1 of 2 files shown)


pkg/console/operator/operator.go
```diff
@@ -176,6 +190,20 @@ func NewConsoleOperator(
 		ToController("ConsoleOperator", recorder.WithComponentSuffix("console-operator"))
 }
 
+func isCapabilityEnabled(client configclientv1.ConfigV1Interface, expectedCap configv1.ClusterVersionCapability) bool {
+	cv, err := client.ClusterVersions().Get(context.TODO(), "version", metav1.GetOptions{})
```
wking, line 194: one-shot ClusterVersion fetch attempt here on operator initiation without a limiting `Context` has some risks.  A lower-stakes approach would be to not enable the OLM portions until we can confirm they're enabled, so you could have behavior like:

1. Console operator is coming up.
2. Console operator attempts to retrieve ClusterVersion with a 5m deadline, but this fails. 
3. Console operator assumes that OLM is not enabled, and goes on to operate its other aspects.
4. Follow-up attempt to check in on ClusterVersion succeeds, and it turns out that OLM is enabled.
5. Console operator starts operating on the OLM-gated stuff too.

You'll want something like that anyway, because admins may decide to enable capabilities post-install.  Only [post-install _disabling_ is not supported][1].

[1]: https://github.com/openshift/enhancements/blob/286e46ad8c1fc0baf9f940b21ea85381c3e54640/enhancements/installer/component-selection.md#capabilities-cannot-be-uninstalled

eggfoobar: I think @wking would just like this to be a bit more robust for future use. So in the event that a user re-enables OLM after the fact, so the pod will need to be restarted for the console to correctly see the resources.

I'm not entirely sure what to do about the informer since it's attached at startup, I updated the code a bit since we just needed to know if a resource exists for that informer. I also updated the error checking so that we don't always skip based off of initial state. If I understand this right, the informer won't be attached, but the lister will be queried after the fact, so if the user re-enables OLM, then the CSV query will go through.

pkg/console/operator/operator.go
```diff
@@ -142,18 +142,34 @@ func NewConsoleOperator(
...
+		Group:    api.OLMConfigGroup,
+		Version:  api.OLMConfigVersion,
+		Resource: api.OLMConfigResource,
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute*5)
+	defer cancel()
+
+	// We check if the resource exists in the cluster at all before attaching an informer
+	if _, err := dynamicClient.Resource(olmGroupVersionResource).List(ctx, metav1.ListOptions{}); !isNoMatchError(err) {
```
jhadvig, line 164: @eggfoobar what if the OLM capability is enabled post installation? Can the capability be enable at any time? or is there any restriction ?
In case the capability will be disabled at start and then enabled, I dont think the informer will be registered, since we are registering the informers in the `NewConsoleOperator` constructor.
Wondering if it wont be better to restart the operator's pod on this change?

pkg/console/operator/operator.go
jhadvig: Right but in case the capability will get enabled post-installation we are not registering the informer. Please check my comment below.

pkg/console/operator/operator.go
eggfoobar: Yeah, that might be a good idea, I added some WIP logic to get this running in my e2e tests that just checks if the resource exists but in the future it would be preferable to have an informer on the clusterversions and if we have the resource enabled, we trigger a reboot then. There might be a more elegant way of doing that, but that's what came to mind, what do you think?

jhadvig: So this is something I had in mind, to basically kill the pod end force new rollout so the informer is registered 👍

jhadvig (approved):
@eggfoobar thanks for addressing this 👍  I like how this got implemented 

/lgtm
/approve

jhadvig:
QE Approver:
/assign @yanpzhan 
Docs Approver:
/assign @opayne1 
PX Approver:
/assign @RickJWagner

eggfoobar:
/hold 

Apologies to add a hold on this so late, noticing some errors in the no-cap job that I would like to look over before we merge this in. Will unhold asap on.

eggfoobar:
/unhold

@jhadvig I was able to verify the error I was seeing wasn't happening again. I went a head and added some more comments and logs to make sure it's clear what's happening. Please take a look when you have a moment, thanks!

yanpzhan:
I built an image against the pr using cluster-bot, have done these tests:
1. Launched a cluster using the image successfully, and updated it to 4.15 nightly payload successfully. 
2. Launched another 4.14 nightly cluster, updated it to the image I created successfully.
3. Launched a cluster using the image with installation options: "baselineCapabilitySet: None" and "additionalEnabledCapabilities: ["Console"]", the cluster was installed successfully. Checked on web console, there was not operators in OperatorHub page.

@eggfoobar @jhadvig except above test If any other tests are needed for this pr?

eggfoobar:
@yanpzhan That looks good to me, just to make sure the current changes were in. Can you tell me which 4.15 nightly release you used, and if you still have the cluster up, can you check `oc get clusterversions version -oyaml` and make sure the `knownCapabilities` include `OperatorLifecycleManager`?

yanpzhan:
@eggfoobar The target 4.15 nightly payload is 4.15.0-0.nightly-2023-10-09-101435, on the cluster, I checked oc get clusterversions version -oyaml, it doesn't contain "OperatorLifecycleManager" in "knownCapabilities".  I wonder if this build doesn't contains changes about OLM Capability.

yanpzhan:
@eggfoobar The target 4.15 nightly payload is 4.15.0-0.nightly-2023-10-09-101435, on the cluster, I checked oc get clusterversions version -oyaml, it doesn't contain "OperatorLifecycleManager" in "knownCapabilities".  I wonder if this build doesn't contains changes about OLM Capability.

eggfoobar:
Ah yup, @yanpzhan would you be able to verify again with this nightly, [4.15.0-0.nightly-2023-10-17-065657](https://amd64.ocp.releases.ci.openshift.org/releasestream/4.15.0-0.nightly/release/4.15.0-0.nightly-2023-10-17-065657) ?That should have the updated capabilities

yanpzhan:
I launched a cluster with the pr and image 4.15.0-0.nightly-2023-10-09-101435 successfully, and checked clusterversions 'version', it contains "OperatorLifecycleManager" in "knownCapabilities" now.

eggfoobar:
Perfect, thanks @yanpzhan !

In terms of what we need to verify two things with this PR:

1. Console should function normally with out OLM being enabled ✅ (you've already verified this part if no errors are happening)
2. When you re-enable OLM via `oc edit clusterversions version` to enable `OperatorLifecycleManager`, after CVO installs OLM, we expect that console will restart in about 5 minutes and continue operating as normal.

yanpzhan:
@eggfoobar I've checked again on new cluster. The 2 points you mentioned passed, no errors. Console works after re-enabled OLM by set 'OperatorLifecycleManager' in Clusterversions 'version'.
**Before :**
# oc get clusterversions.config.openshift.io version -ojson | jq .status.capabilities
{
  "enabledCapabilities": [
    "Console"
  ],
  "knownCapabilities": [
    "Build",
    "CSISnapshot",
    "Console",
    "DeploymentConfig",
    "ImageRegistry",
    "Insights",
    "MachineAPI",
    "NodeTuning",
    "OperatorLifecycleManager",
    "Storage",
    "baremetal",
    "marketplace",
    "openshift-samples"
  ]
}
**After:**
# oc get clusterversions.config.openshift.io version -ojson | jq .status.capabilities
{
  "enabledCapabilities": [
    "Console",
    "OperatorLifecycleManager"
  ],
  "knownCapabilities": [
    "Build",
    "CSISnapshot",
    "Console",
    "DeploymentConfig",
    "ImageRegistry",
    "Insights",
    "MachineAPI",
    "NodeTuning",
    "OperatorLifecycleManager",
    "Storage",
    "baremetal",
    "marketplace",
    "openshift-samples"
  ]
}
Could you tell me what resources about OLM I could check to compare them before/after enable OLM?

eggfoobar:
Perfect, once OLM has been enabled, the specific resource that Console is looking for is `olmconfigs` so if you see that then things should be good, `oc get olmconfigs`, as long as that call doesn't fail then Console will work

yanpzhan:
Thanks for your guide. There is no issue for now. 
/label qe-approved

PR #796, opened 2023-09-25
OCPBUGS-19080: Reset console operator's conditions

jhadvig:
/assign @TheRealJon

### DIFF
--- a/pkg/console/status/status.go
+++ b/pkg/console/status/status.go
@@ -64,6 +64,36 @@ func HandleUpgradable(typePrefix string, reason string, err error) v1helpers.Upd
 	return v1helpers.UpdateConditionFn(condition)
 }
 
+func (c *StatusHandler) ResetConditions(conditions []operatorsv1.OperatorCondition) []v1helpers.UpdateStatusFunc {
+	updateStatusFuncs := []v1helpers.UpdateStatusFunc{}
+	for _, condition := range conditions {
+		klog.V(2).Info("\nresetting condition: ", condition.Type)
+		if strings.HasSuffix(condition.Type, operatorsv1.OperatorStatusTypeDegraded) {
+			conditionPrefix := strings.TrimSuffix(condition.Type, operatorsv1.OperatorStatusTypeDegraded)
+			updateStatusFuncs = append(updateStatusFuncs, HandleDegraded(conditionPrefix, "", nil))
+			continue
+		}
+		if strings.HasSuffix(condition.Type, operatorsv1.OperatorStatusTypeAvailable) {
+			conditionPrefix := strings.TrimSuffix(condition.Type, operatorsv1.OperatorStatusTypeAvailable)
+			updateStatusFuncs = append(updateStatusFuncs, HandleAvailable(conditionPrefix, "", nil))
+			continue
+		}
+		if strings.HasSuffix(condition.Type, operatorsv1.OperatorStatusTypeProgressing) {
+			conditionPrefix := strings.TrimSuffix(condition.Type, operatorsv1.OperatorStatusTypeProgressing)
+			updateStatusFuncs = append(updateStatusFuncs, HandleProgressing(conditionPrefix, "", nil))
+			continue
+		}
+		if strings.HasSuffix(condition.Type, operatorsv1.OperatorStatusTypeUpgradeable) {
+			conditionPrefix := strings.TrimSuffix(condition.Type, operatorsv1.OperatorStatusTypeUpgradeable)
+			updateStatusFuncs = append(updateStatusFuncs, HandleUpgradable(conditionPrefix, "", nil))
+			continue
+		}
+		klog.V(2).Info("unable to reset condition: ", condition.Type)
+	}
+
+	return updateStatusFuncs
+}
+
 // HandleProgressingOrDegraded exists until we remove type SyncError
 // If isSyncError
 // - Type suffix will be set to Progressing
--- a/go.mod
+++ b/go.mod
@@ -51,7 +51,6 @@ require (
 	github.com/golang/protobuf v1.5.3 // indirect
 	github.com/google/cel-go v0.16.0 // indirect
 	github.com/google/gnostic v0.6.9 // indirect
-	github.com/google/gnostic-models v0.6.8 // indirect
 	github.com/google/go-cmp v0.5.9 // indirect
 	github.com/google/gofuzz v1.2.0 // indirect
 	github.com/google/pprof v0.0.0-20230705174524-200ffdc848b8 // indirect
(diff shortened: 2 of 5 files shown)


pkg/console/status/status.go
```diff
@@ -64,6 +64,36 @@ func HandleUpgradable(typePrefix string, reason string, err error) v1helpers.Upd
 	return v1helpers.UpdateConditionFn(condition)
 }
 
+func (c *StatusHandler) ResetConditions(conditions []operatorsv1.OperatorCondition) []v1helpers.UpdateStatusFunc {
```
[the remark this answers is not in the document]
jhadvig: I see your point, but I would rather have a generic function like this, that would loop though all the conditions in the main controller and reset them.
The point is that is more reliable then if we forget to add a new prefix to the controller's list of responsible conditions or even if a typo appears. 
Also once the `Removed` state is set on the operator we are not setting/updating any conditions, just logging into the operator's pod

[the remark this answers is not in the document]
jhadvig: Thats why I would reset all the conditions in this main sync loop or create a something like a maintenance controller. But since we dont have any other use-cases for it I would just do the simple thing and reset it in the main controller.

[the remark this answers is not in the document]
TheRealJon: Sounds good to me!

pkg/console/status/status.go
```diff
@@ -64,6 +64,36 @@ func HandleUpgradable(typePrefix string, reason string, err error) v1helpers.Upd
...
+	updateStatusFuncs := []v1helpers.UpdateStatusFunc{}
+	for _, condition := range conditions {
+		klog.V(2).Info("\nresetting condition: ", condition.Type)
```
TheRealJon, line 70: Should we move this output to each "if" block so that we can also output what we are resetting the value to?

```suggestion
```

pkg/console/status/status.go
```diff
@@ -64,6 +64,36 @@ func HandleUpgradable(typePrefix string, reason string, err error) v1helpers.Upd
...
+		if strings.HasSuffix(condition.Type, operatorsv1.OperatorStatusTypeDegraded) {
```
TheRealJon, line 71:
```suggestion
		if strings.HasSuffix(condition.Type, operatorsv1.OperatorStatusTypeDegraded) {
		  klog.Infof("resetting condition %v to false", condition.type)

```

pkg/console/status/status.go
```diff
@@ -64,6 +64,36 @@ func HandleUpgradable(typePrefix string, reason string, err error) v1helpers.Upd
...
+			conditionPrefix := strings.TrimSuffix(condition.Type, operatorsv1.OperatorStatusTypeDegraded)
+			updateStatusFuncs = append(updateStatusFuncs, HandleDegraded(conditionPrefix, "", nil))
+			continue
+		}
+		if strings.HasSuffix(condition.Type, operatorsv1.OperatorStatusTypeAvailable) {
```
TheRealJon, line 76:
```suggestion
		if strings.HasSuffix(condition.Type, operatorsv1.OperatorStatusTypeAvailable) {
      klog.Infof("resetting condition %v to true", condition.type)
```

pkg/console/status/status.go
```diff
@@ -64,6 +64,36 @@ func HandleUpgradable(typePrefix string, reason string, err error) v1helpers.Upd
...
+			conditionPrefix := strings.TrimSuffix(condition.Type, operatorsv1.OperatorStatusTypeAvailable)
+			updateStatusFuncs = append(updateStatusFuncs, HandleAvailable(conditionPrefix, "", nil))
+			continue
+		}
+		if strings.HasSuffix(condition.Type, operatorsv1.OperatorStatusTypeProgressing) {
```
TheRealJon, line 81:
```suggestion
		if strings.HasSuffix(condition.Type, operatorsv1.OperatorStatusTypeProgressing) {
		  klog.Infof("resetting condition %v to false", condition.type)
```

pkg/console/status/status.go
```diff
@@ -64,6 +64,36 @@ func HandleUpgradable(typePrefix string, reason string, err error) v1helpers.Upd
...
+			conditionPrefix := strings.TrimSuffix(condition.Type, operatorsv1.OperatorStatusTypeProgressing)
+			updateStatusFuncs = append(updateStatusFuncs, HandleProgressing(conditionPrefix, "", nil))
+			continue
+		}
+		if strings.HasSuffix(condition.Type, operatorsv1.OperatorStatusTypeUpgradeable) {
```
TheRealJon, line 86:
```suggestion
		if strings.HasSuffix(condition.Type, operatorsv1.OperatorStatusTypeUpgradeable) {
		  klog.Infof("resetting condition %v to true", condition.type)
```

pkg/console/status/status.go
```diff
@@ -64,6 +64,36 @@ func HandleUpgradable(typePrefix string, reason string, err error) v1helpers.Upd
...
+			conditionPrefix := strings.TrimSuffix(condition.Type, operatorsv1.OperatorStatusTypeUpgradeable)
+			updateStatusFuncs = append(updateStatusFuncs, HandleUpgradable(conditionPrefix, "", nil))
+			continue
+		}
+		klog.V(2).Info("unable to reset condition: ", condition.Type)
```
TheRealJon, line 91: Should we keep the logs relating to individual conditions at a higher log level so they won't be logged by default?
```suggestion
		klog.Info("unable to reset condition: ", condition.Type)
```

pkg/console/status/status.go
```diff
@@ -64,6 +64,36 @@ func HandleUpgradable(typePrefix string, reason string, err error) v1helpers.Upd
...
+	}
```
TheRealJon, line 92:
```suggestion
	}
	
	klog.V(2).Info("Done resetting console operator conditions to default values.")
```

pkg/console/status/status.go
```diff
@@ -64,6 +64,36 @@ func HandleUpgradable(typePrefix string, reason string, err error) v1helpers.Upd
 	return v1helpers.UpdateConditionFn(condition)
 }
 
+func (c *StatusHandler) ResetConditions(conditions []operatorsv1.OperatorCondition) []v1helpers.UpdateStatusFunc {
+	updateStatusFuncs := []v1helpers.UpdateStatusFunc{}
+	for _, condition := range conditions {
```
TheRealJon, line 69:
```suggestion
  klog.V(2).Info("Resetting console operator conditions to default values.")
	for _, condition := range conditions {
```

TheRealJon (approved):
Just a few suggestions about what we log. Nothing that should hold up the PR.

/lgtm

PR #798, opened 2023-10-12
CONSOLE-3767: Update deps

jhadvig:


jhadvig:
Closing since deps have been update part of https://github.com/openshift/console-operator/pull/782

PR #799, opened 2023-10-13
CONSOLE-3613: Remove multicluster code from console operator

TheRealJon:
Remove all multicluster related code from console operator repo


jhadvig (approved):
/lgtm

jhadvig:
Adding tags manually since this is just removing dead code.

PR #800, opened 2023-10-16
PoC: allow configuring console in an external OIDC identity provider environments

stlaz:
/cc @sjenning @deads2k @liouk 

/hold
This is currently a PoC, final version should use actual API instead of grabbing config from a secret.

TODOs:
- [x] don't look for oauth-server-cert if authn=None | OIDC, make it possible to configure the CA via the secret

Allows client ID/secret and OIDC CA config via secret:
```
openshift-config/console-client-config

.data:
  issuer
  client-id
  client-secret
  ca.crt
  extra-scopes
```

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -1,6 +1,8 @@
 package api
 
 const (
+	AuthServerCAMountDir                = "/var/auth-server-ca"
+	AuthServerCAFileName                = "ca.crt"
 	ClusterOperatorName                 = "console"
 	ConfigResourceName                  = "cluster"
 	ConsoleContainerPort                = 443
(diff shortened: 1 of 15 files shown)


stlaz:
TODO: don't look for oauth-server-cert if authn=None, make it possible to configure the CA via the secret

pkg/api/api.go
```diff
@@ -1,6 +1,8 @@
 package api
 
 const (
+	AuthServerCAMountDir                = "/var/auth-server-ca"
+	AuthServerCAFileName                = "ca.crt"
```
stlaz, line 5: with the API, this would be "ca-bundle.crt"

stlaz:
/close
in favour of https://github.com/openshift/console-operator/pull/811

PR #801, opened 2023-10-17
HOSTEDCP-1250: remove redundant cert check

stlaz:
The oauth-serving-cert is now present in all supported OpenShift versions.

The SA CA bundle would've never been a good choice to get the certs from :|

Removes additional unused function that I randomly found.

The original configurable CA logic is kept as it's going to get reused in #800 

/assign @jhadvig


jhadvig (approved):
/lgtm
/approve

jhadvig:
adding all the necessary labels manually since its removing dead code

PR #808, opened 2023-10-31
CONSOLE-3804: split oauthclient controller from the common logic

stlaz:
First PR to prevent oauth-client creation when the oauth.openshift.io API is unavailable.

This splits the controller handling the oauthclients away from the main sync loop.

/assign @jhadvig

### DIFF
--- a/pkg/console/controllers/oauthclients/oauthclients.go
+++ b/pkg/console/controllers/oauthclients/oauthclients.go
@@ -1,0 +1,207 @@
+package oauthclients
+
+import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1informers "k8s.io/client-go/informers/core/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/klog/v2"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	configv1informers "github.com/openshift/client-go/config/informers/externalversions/config/v1"
+	configv1lister "github.com/openshift/client-go/config/listers/config/v1"
+	oauthv1client "github.com/openshift/client-go/oauth/clientset/versioned/typed/oauth/v1"
+	oauthv1informers "github.com/openshift/client-go/oauth/informers/externalversions/oauth/v1"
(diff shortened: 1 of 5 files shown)


pkg/console/controllers/oauthclients/oauthclients.go
```diff
@@ -0,0 +1,203 @@
...
+	case operatorv1.Unmanaged:
+		klog.V(4).Infoln("console is in an unmanaged state.")
+		return false, nil
+	case operatorv1.Removed:
+		klog.V(4).Infoln("console has been removed.")
+		return false, c.deregisterClient(ctx)
+	default:
+		return false, fmt.Errorf("console is in an unknown state: %v", managementState)
+	}
+}
```
TheRealJon, line 155: Could we add a TODO comment on this func to make it reusable for all of our controllers? We repeat this same logic in all of them.

pkg/console/operator/operator.go
```diff
@@ -123,37 +135,31 @@ func NewConsoleOperator(
...
-	targetNameFilter := util.IncludeNamesFilter(api.OpenShiftConsoleName)
-
 	informers := []factory.Informer{
 		configV1Informers.Consoles().Informer(),
 		operatorConfigInformer.Informer(),
 		configV1Informers.Infrastructures().Informer(),
 		configV1Informers.Ingresses().Informer(),
 		configV1Informers.Proxies().Informer(),
 		configV1Informers.OAuths().Informer(),
+		oauthClientInformer.Informer(),
```
TheRealJon, line 162: This means we're only watching an OAuthClient named 'cluster'. Is that correct?

TheRealJon (commented):
I think this looks good for the most part. Just one small question. I definitely like to see more controllers getting broken out of `operator.go`.

pkg/console/operator/operator.go
stlaz: good catch, I did not notice these informers were actually filtered by name

pkg/console/controllers/oauthclients/oauthclients.go
stlaz: Absolutely. A long time ago I was thinking about how to add similar functionality to the controller factory itself but then there was a lot of other work and I never got back to it.

TheRealJon (approved):
/lgtm

TheRealJon:
QE Approver
/assign @yanpzhan 

Docs Approver:
/assign @opayne1

PX Approver:
/assign @RickJWagner

stlaz:
/retest-required
unrelated installation failure

jhadvig (approved):
/lgtm

yanpzhan:
Regression test about idp related cases passed on the cluster launched against the pr.
/label qe-approved

PR #813, opened 2023-11-17
TRT-1361: Revert #808 "CONSOLE-3804: split oauthclient controller from the common logic"

dgoodwin:
Reverts #808 ; tracked by [TRT-1361](https://issues.redhat.com//browse/TRT-1361)

Per [OpenShift policy](https://github.com/openshift/enhancements/blob/master/enhancements/release/improving-ci-signal.md#quick-revert), we are reverting this breaking change to get CI and/or nightly payloads flowing again.

We suspect this PR broke hypershift upgrades though one does occasionally still slip through. This would have been hard to detect pre-merge for that fact, as well as the possibility the problem only surfaces once the change is in an accepted payload. (i.e. it only shows when in the version being upgraded FROM)

To unrevert this, revert this PR, and layer an additional separate commit on top that addresses the problem. Before merging the unrevert, please run these jobs on the PR and check the result of these jobs to confirm the fix has corrected the problem:

```
Please coordinate with hypershift team to see if there's a way to get reasonable pre-merge confidence before bringing it back in, and monitor job status carefully once it re-merges.
```

CC: @stlaz

<div align="right">
PR created by <a href="https://github.com/stbenjam/revertomatic">Revertomatic<sup>:tm:</sup></a>
</div>


dgoodwin:
If we got this wrong, our apologies and we will help with the unrevert. We're in a difficult spot as this appears impossible to verify pre-merge, it must be in an accepted ci payload.

PR #814, opened 2023-11-22
CONSOLE-3804: Reapply separation of the oauthclients controller

stlaz:
Includes two fixes, one to status handling, another is an oversight of a cache write.

/assign @rhamilto 
/cc @csrwng @dgoodwin

### DIFF
--- a/pkg/console/controllers/downloadsdeployment/controller.go
+++ b/pkg/console/controllers/downloadsdeployment/controller.go
@@ -14,10 +14,10 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
-	configclientv1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	configinformer "github.com/openshift/client-go/config/informers/externalversions"
-	operatorclientv1 "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1"
+	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
 	operatorinformerv1 "github.com/openshift/client-go/operator/informers/externalversions/operator/v1"
+	operatorlistersv1 "github.com/openshift/client-go/operator/listers/operator/v1"
 	"github.com/openshift/console-operator/pkg/api"
 	"github.com/openshift/console-operator/pkg/console/status"
 	"github.com/openshift/library-go/pkg/controller/factory"
@@ -33,18 +33,15 @@ import (
 type DownloadsDeploymentSyncController struct {
 	operatorClient v1helpers.OperatorClient
 	// configs
-	operatorConfigClient       operatorclientv1.ConsoleInterface
-	infrastructureConfigClient configclientv1.InfrastructureInterface
+	consoleOperatorLister operatorlistersv1.ConsoleLister
+	infrastructureLister  configlistersv1.InfrastructureLister
 	// core kube
 	deploymentClient appsclientv1.DeploymentsGetter
 }
 
 func NewDownloadsDeploymentSyncController(
-	// top level config
-	configClient configclientv1.ConfigV1Interface,
 	// clients
 	operatorClient v1helpers.OperatorClient,
-	operatorConfigClient operatorclientv1.ConsoleInterface,
 	// informer
 	configInformer configinformer.SharedInformerFactory,
 	operatorConfigInformer operatorinformerv1.ConsoleInformer,
@@ -58,9 +55,9 @@ func NewDownloadsDeploymentSyncController(
 
 	ctrl := &DownloadsDeploymentSyncController{
 		// configs
-		operatorClient:             operatorClient,
-		operatorConfigClient:       operatorConfigClient,
-		infrastructureConfigClient: configClient.Infrastructures(),
+		operatorClient:        operatorClient,
+		consoleOperatorLister: operatorConfigInformer.Lister(),
+		infrastructureLister:  configInformer.Config().V1().Infrastructures().Lister(),
 		// client
 		deploymentClient: deploymentClient,
 	}
@@ -81,13 +78,13 @@ func NewDownloadsDeploymentSyncController(
 }
 
 func (c *DownloadsDeploymentSyncController) Sync(ctx context.Context, controllerContext factory.SyncContext) error {
-	operatorConfig, err := c.operatorConfigClient.Get(ctx, api.ConfigResourceName, metav1.GetOptions{})
+	operatorConfig, err := c.consoleOperatorLister.Get(api.ConfigResourceName)
 	if err != nil {
 		return err
 	}
-	updatedOperatorConfig := operatorConfig.DeepCopy()
+	operatorConfigCopy := operatorConfig.DeepCopy()
 
-	switch updatedOperatorConfig.Spec.ManagementState {
+	switch operatorConfigCopy.Spec.ManagementState {
 	case operatorv1.Managed:
 		klog.V(4).Infoln("console is in a managed state: syncing downloads deployment")
 	case operatorv1.Unmanaged:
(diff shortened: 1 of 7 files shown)


dgoodwin:
/payload-job periodic-ci-openshift-hypershift-release-4.15-periodics-e2e-aws-ovn

stlaz:
/retest-required
/payload-job periodic-ci-openshift-hypershift-release-4.15-periodics-e2e-aws-ovn
- azure-own-upgrade - the failure is unrelated to console
- gcp-ovn - the API server appears to have been unavailable for a bit an console was the slowest to pick it up. It became available but failed to update its status. Retest to see if this is reproducible or the behavior was caused by unstable infra
- payload test: the upgrade failure disappeared. The other tests are not green but it does not seem like the console-operator is the culprit there.

stlaz:
/hold
ready for review, putting hold here so that @dgoodwin might remove it if he's fine with the PR

pkg/console/controllers/downloadsdeployment/controller.go
```diff
@@ -81,13 +78,13 @@ func NewDownloadsDeploymentSyncController(
 }
 
 func (c *DownloadsDeploymentSyncController) Sync(ctx context.Context, controllerContext factory.SyncContext) error {
-	operatorConfig, err := c.operatorConfigClient.Get(ctx, api.ConfigResourceName, metav1.GetOptions{})
+	operatorConfig, err := c.consoleOperatorLister.Get(api.ConfigResourceName)
 	if err != nil {
 		return err
 	}
-	updatedOperatorConfig := operatorConfig.DeepCopy()
+	operatorConfigCopy := operatorConfig.DeepCopy()
```
TheRealJon, line 85: Not new, but why do we even need a copy? Do we make changes to this resource somewhere?

pkg/console/controllers/oauthclients/oauthclients.go
```diff
@@ -0,0 +1,205 @@
...
+	} else if !shouldSync {
+		return nil
+	}
+
+	operatorConfig, err := c.consoleOperatorLister.Get("cluster")
+	if err != nil {
+		return err
+	}
+
+	ingressConfig, err := c.ingressConfigLister.Get("cluster")
```
TheRealJon, line 106: nit, magic string
```suggestion
	ingressConfig, err := c.ingressConfigLister.Get(api.ConfigResourceName)
```

pkg/console/controllers/oauthclients/oauthclients.go
```diff
@@ -0,0 +1,205 @@
...
+	if err != nil {
+		return err
+	}
+
+	routeName := api.OpenShiftConsoleRouteName
+	routeConfig := routesub.NewRouteConfig(operatorConfig, ingressConfig, routeName)
+	if routeConfig.IsCustomHostnameSet() {
+		routeName = api.OpenshiftConsoleCustomRouteName
+	}
```
TheRealJon, line 115: not new, but we are creating an otherwise unused intermediary `routeConfig` instance here just to set a string value. Should this just be a helper func like:

```golang
routeName := routesub.GetRouteName(operatorConfig, ingressConfig)
```

pkg/console/controllers/oauthclients/oauthclients.go
```diff
@@ -0,0 +1,205 @@
...
+
+	_, consoleURL, _, routeErr := operator.GetActiveRouteInfo(ctx, c.routesLister, routeName)
+	if routeErr != nil {
+		return routeErr
+	}
+
+	clientSecret, _, secErr := c.syncSecret(ctx, operatorConfig, controllerContext.Recorder())
```
TheRealJon, line 122: If we don't need the second value, could we just update the `c.syncSecret` to only return what we need?
```suggestion
	clientSecret, secErr := c.syncSecret(ctx, operatorConfig, controllerContext.Recorder())
```

pkg/console/controllers/oauthclients/oauthclients.go
```diff
@@ -0,0 +1,205 @@
...
+	secret, err := c.targetNSSecretsLister.Secrets(api.TargetNamespace).Get(secretsub.Stub().Name)
+	if apierrors.IsNotFound(err) || secretsub.GetSecretString(secret) == "" {
+		return resourceapply.ApplySecret(ctx, c.secretsClient, recorder, secretsub.DefaultSecret(operatorConfig, crypto.Random256BitsString()))
+	}
+	// any error should be returned & kill the sync loop
+	if err != nil {
+		return nil, false, err
+	}
+	return secret, false, nil
+}
```
TheRealJon, line 168:
```suggestion
func (c *oauthClientsController) syncSecret(ctx context.Context, operatorConfig *operatorv1.Console, recorder events.Recorder) (*corev1.Secret, error) {
	secret, err := c.targetNSSecretsLister.Secrets(api.TargetNamespace).Get(secretsub.Stub().Name)
	if apierrors.IsNotFound(err) || secretsub.GetSecretString(secret) == "" {
		secret, _, err = resourceapply.ApplySecret(ctx, c.secretsClient, recorder, secretsub.DefaultSecret(operatorConfig, crypto.Random256BitsString()))
        return secret, err
	}
	// any error should be returned & kill the sync loop
	if err != nil {
		return nil, err
	}
	return secret, nil
}
```

pkg/console/operator/operator.go
```diff
@@ -248,55 +259,57 @@ type configSet struct {
...
 
 	configs := configSet{
-		Console:        consoleConfig,
-		Operator:       operatorConfig,
-		Infrastructure: infrastructureConfig,
-		Proxy:          proxyConfig,
-		OAuth:          oauthConfig,
-		Ingress:        ingressConfig,
+		Console:        consoleConfig.DeepCopy(),
+		Operator:       operatorConfig.DeepCopy(),
```
TheRealJon, line 308: This is also copied again on L323

pkg/console/operator/sync_v400.go
```diff
@@ -119,28 +118,20 @@ func (co *consoleOperator) sync_v400(ctx context.Context, controllerContext fact
...
-	oauthClient, oauthChanged, oauthErrReason, oauthErr := co.SyncOAuthClient(ctx, set.Operator, sec, consoleURL.String())
-	toUpdate = toUpdate || oauthChanged
-	statusHandler.AddConditions(status.HandleProgressingOrDegraded("OAuthClientSync", oauthErrReason, oauthErr))
-	if oauthErr != nil {
-		return statusHandler.FlushAndReturn(oauthErr)
-	}
-
 	actualDeployment, depChanged, depErrReason, depErr := co.SyncDeployment(
 		ctx,
 		set.Operator,
```
TheRealJon, line 129: not new, `set.Operator` is already deep copied because of the [this change](https://github.com/openshift/console-operator/pull/814/files#diff-d4d44dccf244addcf7cfdaf11b2e27171132f629a5c14a6382bdb712a82bedf5R307-R312), but we copy it again in `syncDeployment`. As things are, this resource gets deep copied 3 times:

- https://github.com/openshift/console-operator/pull/814/files#diff-d4d44dccf244addcf7cfdaf11b2e27171132f629a5c14a6382bdb712a82bedf5R308
- https://github.com/openshift/console-operator/pull/814/files#diff-d4d44dccf244addcf7cfdaf11b2e27171132f629a5c14a6382bdb712a82bedf5R323
- https://github.com/openshift/console-operator/pull/814/files#diff-b2ca349b96cbce4c3caf1c27d93f89b881f4e11ab0091498ec48e1bedcc25be2R255

Also, if all resources in the `set` arg to `synv_v400` are already deep copied, I think we can get rid of the `updatedOperatorConfig` arg, right?

pkg/console/operator/sync_v400.go
```diff
@@ -224,8 +209,8 @@ func (co *consoleOperator) sync_v400(ctx context.Context, controllerContext fact
 	return statusHandler.FlushAndReturn(nil)
 }
 
-func (co *consoleOperator) GetActiveRouteInfo(ctx context.Context, activeRouteName string) (route *routev1.Route, routeURL *url.URL, reason string, err error) {
-	route, routeErr := co.routeClient.Routes(api.TargetNamespace).Get(ctx, activeRouteName, metav1.GetOptions{})
+func GetActiveRouteInfo(ctx context.Context, routeClient routev1listers.RouteLister, activeRouteName string) (route *routev1.Route, routeURL *url.URL, reason string, err error) {
```
TheRealJon, line 212: This should probably be moved out of this file. Maybe into the `route` subresource package?

pkg/console/controllers/downloadsdeployment/controller.go
stlaz: You actually wouldn't need a copy in this function, but you might need it further on I believe.

When you're using listers, a `Get()` method works with an informer cache. It actually returns a pointer *into* the cache. If you did not create a copy, you would be mutating your local cache for the object, which may have unwanted consequences and might prevent you from being able to update the object in the cluster (until the cache is refreshed again).

pkg/console/controllers/oauthclients/oauthclients.go
stlaz: I guess that would work, I am a bit unfamiliar with these helpers :)

pkg/console/operator/operator.go
```diff
@@ -248,55 +259,57 @@ type configSet struct {
...
+		Infrastructure: infrastructureConfig.DeepCopy(),
+		Proxy:          proxyConfig.DeepCopy(),
+		OAuth:          oauthConfig.DeepCopy(),
+		Ingress:        ingressConfig.DeepCopy(),
```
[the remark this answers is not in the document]
stlaz: See https://github.com/openshift/console-operator/pull/814#discussion_r1411768006.

I suppose some of them could be just the original object as they would only ever be read but since I don't know the code too well, it's easier for me to just create copies right away.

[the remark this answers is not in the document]
stlaz: Another point I wanted to raise - this actually adheres to the original workflow of the operator but if anyone's addressing the tech debt later, I would strongly discourage this kind of pre-sync state store in favour of just-in-time state retrievals.

This way you would get rid of any possible side-effects which is a typical problem in these situations.

pkg/console/operator/operator.go
stlaz: removed the latter deep copy

pkg/console/controllers/oauthclients/oauthclients.go
```diff
@@ -0,0 +1,205 @@
...
+
+func (c *oauthClientsController) sync(ctx context.Context, controllerContext factory.SyncContext) error {
+	if shouldSync, err := c.handleStatus(ctx); err != nil {
+		return err
+	} else if !shouldSync {
+		return nil
+	}
+
+	operatorConfig, err := c.consoleOperatorLister.Get("cluster")
```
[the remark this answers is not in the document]
stlaz: The `console.operator/cluster` is actually technically not a config resource but an operator resource. I find it confusing so I avoided those constants completely.

pkg/console/controllers/downloadsdeployment/controller.go
TheRealJon: Yeah. From what I can tell we never actually muatate this resource, so it seems like we could get rid of the deep copy operation altogether. But, we don't need to address it here, it just caught my eye.

yanpzhan:
IDP related regression test on cluster launched against the pr passed.
/label qe-approved

PR #819, opened 2023-11-29
OCPBUGS-24203: ConsolePlugin metrics must no longer be grouped by the vendor

jerolimov:
The monitoring team is fine that we save all installed, known RH plugin names in our metrics/promtheus, instead of grouping them by "redhat", "demo" and "others".

https://github.com/openshift/console/pull/13383 removed this mapping to a vendor and replaced it with a mapping to simplified plugin names.

This PR is required so that the metrics from the console are exported correctly to the cluster monitoring.


jerolimov:
Thanks @jhadvig for merging this faster then expected :+1: 

Double-checked this now on a nightly build:

image

image

PR #822, opened 2023-12-08
OCPBUGS-15827: Update console conversion webhook server to use sig.k8s.io certwatcher

TheRealJon:
Console conversion webhook server will automatically update TLS config when cert files change.


pkg/cmd/crdconversionwebhook/cmd.go
```diff
@@ -31,22 +35,63 @@ func NewConverter() *cobra.Command {
...
-}
-
 func startServer() {
-	config := Config{CertFile: certFile, KeyFile: keyFile}
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// Initialize klog
+	klog.InitFlags(nil)
+	log.SetLogger(klog.NewKlogr()) // controller-runtime/log will complain if we don't set this
```
jhadvig, line 44: We should do the initialisation for the operator pod as well.

jhadvig (approved):
One comment, otherwise LGTM
/lgtm
/approve

PR #825, opened 2023-12-18
OCPBUGS-20467: openshift-enterprise-console-operator-container: openshift: no rate limiting on console or downloads routes [openshift-4]

cyril-ui-developer:


bindata/assets/routes/console-route.yaml
```diff
@@ -7,6 +7,10 @@ metadata:
   namespace: openshift-console
   annotations:
     haproxy.router.openshift.io/timeout: 5m
+    haproxy.router.openshift.io/rate-limit-connections: "true"
+    haproxy.router.openshift.io/rate-limit-connections.concurrent-tcp: 3
```
spadgett, line 11: Will this cause problems for console itself? Console frequently holds more watches than this open, and we make many concurrent requests on load in the background for API discovery. This will limit it to 3 only from a browser, right? Does this setting impact websockets?

spadgett (commented):
/hold

bindata/assets/routes/console-route.yaml
cyril-ui-developer: > Will this cause problems for console itself? Console frequently holds more watches than this open, and we make many concurrent requests on load in the background for API discovery. This will limit it to 3 only from a browser, right? Does this setting impact websockets?

@spadgett Right, 3 requests per second.
cc: @jhadvig

jhadvig:
```
level=error msg=Cluster operator console Degraded is True with OAuthClientsController_SyncError::RouteHealth_RouteNotAdmitted::SyncLoopRefresh_FailedGet: OAuthClientsControllerDegraded: route.route.openshift.io "console" not found
level=error msg=RouteHealthDegraded: console route is not admitted
level=error msg=SyncLoopRefreshDegraded: route.route.openshift.io "console" not found
level=error msg=Cluster operator console Available is False with RouteHealth_RouteNotAdmitted: RouteHealthAvailable: console route is not admitted
```

/test e2e-aws-console
/test e2e-aws-operator

PR #830, opened 2024-01-02
NO-JIRA:(chore) remove unused loglevel

lance5890:
As the operator loglevel is controlled by the cr operatorLogLevel, and the --v in the deployment has no effect anymore, we should remove the redundant flag


jhadvig:
@lance5890 could you please rebase ? Thank you :)

lance5890:
> @lance5890 could you please rebase ? Thank you :)

rebased, please review

PR #834, opened 2024-01-08
OCPBUGS-24041: Prevent healthcheck controller from Available=False blipping 

jhadvig:
Wrapping functionality of the `CheckRouteHealth` function into a retry so the operator wont be degraded when a route health check starts failing for short period of time.

/assign @TheRealJon 

ccing @wking

### DIFF
--- a/pkg/console/controllers/healthcheck/controller.go
+++ b/pkg/console/controllers/healthcheck/controller.go
@@ -13,6 +13,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	coreinformersv1 "k8s.io/client-go/informers/core/v1"
 	coreclientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 
 	// openshift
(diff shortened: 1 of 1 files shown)


pkg/console/controllers/healthcheck/controller.go
```diff
@@ -150,32 +151,41 @@ func (c *HealthCheckController) Sync(ctx context.Context, controllerContext fact
...
-		return "RouteNotAdmitted", fmt.Errorf("console route is not admitted")
-	}
-
-	caPool, err := c.getCA(ctx, route.Spec.TLS)
-	if err != nil {
-		return "FailedLoadCA", fmt.Errorf("failed to read CA to check route health: %v", err)
-	}
-	client := clientWithCA(caPool)
+	var reason string
+	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
wking, line 155: Trying to understand how this works, [Godocs][1]:

> On a non-"Conflict" error, or if it retries too many times and gives up, RetryOnConflict will return an error to the caller.

I'm not sure if [`Client.Timeout exceeded while awaiting headers`][2] would satisfy [`IsConflict`][6] or not.  And then drilling into [the backing Go][3]:

> backoff defines the maximum retries...

And from [`ExponentialBackoff`][5], the limit seems to be `Steps`.  Which [`DefaultRetry` sets to 5][4].  So other than "will these match [`IsConflict`][6]?", this all makes sense to me.

[1]: https://pkg.go.dev/k8s.io/client-go/util/retry#RetryOnConflict
[2]: https://issues.redhat.com/browse/OCPBUGS-24041
[3]: https://github.com/kubernetes/client-go/blob/v0.29.0/util/retry/util.go#L46
[4]: https://github.com/kubernetes/client-go/blob/v0.29.0/util/retry/util.go#L26-L33
[5]: https://github.com/kubernetes/apimachinery/blob/v0.29.0/pkg/util/wait/backoff.go#L452-L464
[6]: https://github.com/kubernetes/apimachinery/blob/v0.29.0/pkg/api/errors/errors.go#L544-L555

jhadvig: @wking good point. Originally I wanted to use `retry.DefaultRetry` but I've ended up using `retry.DefaultBackoff`. 
Reading though the docs I think `retry.DefaultRetry` would be more suitable.
@TheRealJon thoughts ?

TheRealJon: I agree with @wking that `retry.RetryOnConflict` may not behave as expected. We could probably use `retry.OnError`, which allows us to provide a predicate function to determine if the target function should be retried. I think it would look something like this:

```go
err := retry.OnError(
  retry.DefaultRetry, 
  func (err error) bool { return err != nil }
  func() error { /* health check logic */ }
)

jhadvig:
@TheRealJon thanks for comment! Addressed :)
PTAL

PR #838, opened 2024-01-18
OCPBUGS-27335: The console-deployment should set the number of replicas based on the ControlPlaneTopology

aleskandro:
The node selector for the console deployment requires deploying it on the master nodes, The node selector for the console deployment requires deploying it on the master nodes, while the replica count is determined by the infrastructureTopology, which primarily tracks the workers' setup.

When an OpenShift cluster is installed with a single master node and multiple workers, this leads the console deployment to request 2 replicas as infrastructureTopology is set to HighlyAvailable. Instead, ControlPlaneTopology is set to SingleReplica as expected.

This commit makes the operator drive the number of replicas for the console deployment based on the ControlPlaneTopology rather than the InfrastructureTopology.

### DIFF
--- a/pkg/console/subresource/deployment/deployment.go
+++ b/pkg/console/subresource/deployment/deployment.go
@@ -114,7 +114,7 @@ func DefaultDownloadsDeployment(
 
 func withReplicas(deployment *appsv1.Deployment, infrastructureConfig *configv1.Infrastructure) {
 	replicas := int32(SingleNodeConsoleReplicas)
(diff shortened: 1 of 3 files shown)


jhadvig (approved):
/lgtm
/approve

Thanks @aleskandro 👍

aleskandro:
You're welcome, @jhadvig. I'm giving some more shots of testing, although the failures don't seem related to this change.

PR #840, opened 2024-01-22
CONSOLE-3902: Add cli login commands to console configmap

TheRealJon:
Add external OIDC CLI login command to console config

### DIFF
--- a/pkg/console/controllers/clioidcclientstatus/controller.go
+++ b/pkg/console/controllers/clioidcclientstatus/controller.go
@@ -1,0 +1,151 @@
+package clioidcclientstatus
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog/v2"
+
+	configv1 "github.com/openshift/api/config/v1"
+	configv1informers "github.com/openshift/client-go/config/informers/externalversions/config/v1"
+	configv1listers "github.com/openshift/client-go/config/listers/config/v1"
+	operatorv1informers "github.com/openshift/client-go/operator/informers/externalversions/operator/v1"
+	"github.com/openshift/console-operator/pkg/api"
+	"github.com/openshift/console-operator/pkg/console/controllers/util"
+	"github.com/openshift/console-operator/pkg/console/status"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+
+	authnsub "github.com/openshift/console-operator/pkg/console/subresource/authentication"
+)
+
+// cliOIDCClientStatusController:
+//
+//	writes:
+//	- authentication.config.openshift.io/cluster .status.oidcClients:
+//		- componentName=cli
+//		- componentNamespace=openshift-console
+//		- currentOIDCClients
+//		- conditions:
+//			- Available
+//			- Progressing
+//			- Degraded
+//	- consoles.operator.openshift.io/cluster .status.conditions:
+//		- type=CLIOIDCClientStatusProgressing
+//		- type=CLIOIDCClientStatusDegraded
+//		- type=CLIAuthStatusHandlerProgressing
+//		- type=CLIAuthStatusHandlerDegraded
+type cliOIDCClientStatusController struct {
+	authnLister                configv1listers.AuthenticationLister
+	authStatusHandler          *status.AuthStatusHandler
+	operatorClient             v1helpers.OperatorClient
+	statusHandler              status.StatusHandler
+	externalOIDCFeatureEnabled bool
+}
+
+func NewCLIOIDCClientStatusController(
+	operatorClient v1helpers.OperatorClient,
+	authnInformer configv1informers.AuthenticationInformer,
+	authenticationClient configv1client.AuthenticationInterface,
+	consoleOperatorInformer operatorv1informers.ConsoleInformer,
+	externalOIDCFeatureEnabled bool,
+	recorder events.Recorder,
+) factory.Controller {
+	c := &cliOIDCClientStatusController{
+		authnLister:                authnInformer.Lister(),
+		authStatusHandler:          status.NewAuthStatusHandler(authenticationClient, api.CLIOIDCClientComponentName, api.TargetNamespace, "CLIOIDCClientStatusController"),
+		externalOIDCFeatureEnabled: externalOIDCFeatureEnabled,
+		operatorClient:             operatorClient,
+	}
+	return factory.New().
+		WithSync(c.sync).
+		ResyncEvery(wait.Jitter(time.Minute, 1.0)).
+		WithInformers(
+			authnInformer.Informer(),
+			consoleOperatorInformer.Informer(),
+		).
+		ToController("CLIOIDCClientStatusController", recorder.WithComponentSuffix("CLIOIDCClientStatusController"))
+}
+
+func (c *cliOIDCClientStatusController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	c.statusHandler = status.NewStatusHandler(c.operatorClient)
+	return util.HandleManagementState(ctx, c, c.operatorClient)
+}
+
+func (c *cliOIDCClientStatusController) HandleUnmanaged(ctx context.Context) error {
+	klog.V(4).Info("Console is in unmanaged state, skipping CLI OIDC client status sync.")
+	return nil
+}
+
+func (c *cliOIDCClientStatusController) HandleRemoved(ctx context.Context) error {
+	klog.V(4).Info("Console is in removed state, skipping CLI OIDC client status sync.")
+	return nil
+}
+
+func (c *cliOIDCClientStatusController) HandleManaged(ctx context.Context) error {
+	klog.V(4).Info("Console is in managed state, syncing CLI OIDC client status.")
+
+	if !c.externalOIDCFeatureEnabled {
+		c.statusHandler.AddConditions(status.HandleProgressingOrDegraded("CLIOIDCClientStatus", "", nil))
+		c.statusHandler.AddConditions(status.HandleProgressingOrDegraded("CLIAuthStatusHandler", "", nil))
+		return c.statusHandler.FlushAndReturn(nil)
+	}
+
+	authnConfig, err := c.authnLister.Get(api.ConfigResourceName)
+	if err != nil {
+		return err
+	}
+
+	if authnConfig.Spec.Type != configv1.AuthenticationTypeOIDC {
+		applyErr := c.authStatusHandler.Apply(ctx, authnConfig)
+		c.statusHandler.AddConditions(status.HandleProgressingOrDegraded("CLIAuthStatusHandler", "FailedApply", applyErr))
+
+		// reset the other condition set by this controller
+		c.statusHandler.AddConditions(status.HandleProgressingOrDegraded("CLIOIDCClientStatus", "", nil))
+		return c.statusHandler.FlushAndReturn(applyErr)
+	}
+
+	// we need to keep track of errors during the sync so that we can requeue
+	// if any occur
+	var errs []error
+	syncErr := c.syncOIDCCLient(authnConfig)
+	c.statusHandler.AddConditions(
+		status.HandleProgressingOrDegraded(
+			"CLIOIDCClientStatus", "CLIOIDCClientStatusSyncFailed",
+			syncErr,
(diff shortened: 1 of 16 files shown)


pkg/console/subresource/authentication/cluster.go
```diff
@@ -0,0 +1,39 @@
...
+	config "github.com/openshift/api/config/v1"
+)
+
+const (
+	CLIOIDCClientComponentName = "openshift-console/cli"
+	OCLoginCommand             = "oc login"
+	ExecPluginFlag             = "--exec-plugin"
+	OCExecPlugin               = "oc-oidc"
+	ClientIDFlag               = "--client-id"
+	IssuerURLFlag              = "--issuer-url"
```
ardaguclu, line 15: is this API server or external issuer url?. Because `oc login` extracts issuer url from API server's well-known endpoint and there is no `--issuer-url` flag. Additionally, where is setting API server endpoint of the cluster?.

TheRealJon: This was a misunderstanding on my part. I need to switch this to `--server` flag and use the API server endpoint.

pkg/console/subresource/authentication/cluster.go
```diff
@@ -0,0 +1,43 @@
...
+func GetOIDCOCLoginCommand(authConfig *config.Authentication, apiServerURL string) string {
+	clientConfig := GetOIDCClientConfig(authConfig, api.TargetNamespace, api.OCOIDCClientComponentName)
+	if clientConfig == nil {
+		return ""
+	}
+
+	extraArgs := ""
+	if len(clientConfig.ExtraScopes) > 0 {
+		for _, scope := range clientConfig.ExtraScopes {
+			extraArgs = fmt.Sprintf("%s --extra-args scope=%s", extraArgs, strings.TrimSpace(scope))
```
ardaguclu, line 21: This is `--extra-scopes` (e.g. `--extra-scopes=email,profile`) https://github.com/openshift/oc/blob/master/pkg/cli/login/login.go#L99

TheRealJon: That was my original solution. I'll change it back. Is [this doc](https://docs.google.com/document/d/1LJz7xyD5InFSNfSxiG8X2C3uBk0JQKSAMWdWZXwXEtI/edit) outdated?

ardaguclu: In terms of the flag names, this document would include same incorrect names. I'd recommend to check the flags from here https://github.com/openshift/oc/blob/79dc671bdaeafa74b92f14ad9f6d84e344608034/pkg/cli/login/login.go#L94-L99

pkg/console/subresource/authentication/cluster_test.go
```diff
@@ -0,0 +1,100 @@
...
+			name: "OIDC client with CLI client",
+			args: args{
+				&config.Authentication{
+					Spec: config.AuthenticationSpec{
+						OIDCProviders: []config.OIDCProvider{
+							{
+								OIDCClients: []config.OIDCClientConfig{
+									{
+										ComponentNamespace: api.TargetNamespace,
+										ComponentName:      api.OCOIDCClientComponentName,
```
jhadvig, line 57:
```suggestion
										ComponentName:      api.CLIOIDCClientComponentName,
```

pkg/console/subresource/authentication/cluster_test.go
```diff
@@ -0,0 +1,100 @@
...
+			name: "OIDC client with extra scopes",
+			args: args{
+				&config.Authentication{
+					Spec: config.AuthenticationSpec{
+						OIDCProviders: []config.OIDCProvider{
+							{
+								OIDCClients: []config.OIDCClientConfig{
+									{
+										ComponentNamespace: api.TargetNamespace,
+										ComponentName:      api.OCOIDCClientComponentName,
```
jhadvig, line 78:
```suggestion
										ComponentName:      api.CLIOIDCClientComponentName,
```

pkg/console/controllers/clioidcclient/controller.go
```diff
@@ -0,0 +1,174 @@
...
+	if len(clientConfig.ClientID) == 0 {
+		return fmt.Errorf("no ID set on CLI OIDC client")
+	}
+	c.authStatusHandler.WithCurrentOIDCClient(clientConfig.ClientID)
+	c.authStatusHandler.Available("CLIOIDCConfigAvailable", "")
+	return nil
+}
+
+func (c *cliOIDCClientController) resetConditions() error {
+	// reset all conditions set by this controller
```
jhadvig, line 170: lets reset `CLIOIDCConfigAvailable` as well

pkg/console/controllers/util/util.go
```diff
@@ -57,3 +59,28 @@ func LabelFilter(labels map[string]string) factory.EventFilterFunc {
 		return true
 	}
 }
+
+type ConsoleOperatorController interface {
```
jhadvig, line 63: 👍

jhadvig (commented):
Couple of comments, otherwise the PR looks good 👍

pkg/console/controllers/clioidcclient/controller.go
```diff
@@ -0,0 +1,174 @@
...
+
+	authnsub "github.com/openshift/console-operator/pkg/console/subresource/authentication"
+	crdsub "github.com/openshift/console-operator/pkg/console/subresource/crd"
+)
+
+// cliOIDCClientController:
+//
+//	writes:
+//	- authentication.config.openshift.io/cluster .status.oidcClients:
+//		- componentName=cli
```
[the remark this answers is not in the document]
TheRealJon: This is the main reason for this controller. The client can't be added to the spec unless it's reported in the status.

TheRealJon:
QE Approver
/assign @yapei

Docs Approver:
/assign @opayne1

PX Approver:
/assign @RickJWagner

TheRealJon:
@jhadvig @stlaz This PR is ready for review!

TheRealJon:
Looks like flakes
/retest

pkg/console/controllers/util/util.go
```diff
@@ -57,3 +59,28 @@ func LabelFilter(labels map[string]string) factory.EventFilterFunc {
 		return true
 	}
 }
+
+type consoleOperatorController interface {
+	HandleManaged(context.Context) error
+	HandleUnmanaged(context.Context) error
+	HandleRemoved(context.Context) error
+	GetOperatorSpec() (*operatorv1.OperatorSpec, error)
```
stlaz, line 67: in which case is this different from what you'd get from `GetOperatorState()` from `operatorClient v1helpers.OperatorClient`?

pkg/console/controllers/util/util.go
```diff
@@ -57,3 +59,28 @@ func LabelFilter(labels map[string]string) factory.EventFilterFunc {
...
+}
+
+func HandleManagementState(ctx context.Context, c consoleOperatorController) error {
+	operatorSpec, err := c.GetOperatorSpec()
+	if err != nil {
+		return fmt.Errorf("failed to retrieve operator config: %w", err)
+	}
+
+	switch managementState := operatorSpec.ManagementState; managementState {
```
stlaz, line 76: just a note, no action necessary:
I've always thought that if I'm going to do it this, I'd keep the handled conditions in a map, so that I can do this:
```go
handler, ok := c.stateHandlers[managementState]
if !ok {
    return fmt.Errorf("console is in an unknown state: %v", managementState)
}

// noop handling a state
if handler == nil {
    klog.V(4).Infof("handling %q as noop", managementState)
    return nil
}

return handler(ctx)
```

I find it a bit more elegant than the interfaces but it's just a personal preference.

pkg/console/controllers/clioidcclientstatus/controller.go
```diff
@@ -0,0 +1,178 @@
...
+
+func (c *cliOIDCClientStatusController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	c.statusHandler = status.NewStatusHandler(c.operatorClient)
+	return util.HandleManagementState(ctx, c)
+}
+
+func (c *cliOIDCClientStatusController) GetOperatorSpec() (*operatorv1.OperatorSpec, error) {
+	spec, _, _, err := c.operatorClient.GetOperatorState()
+	return spec, err
+}
```
stlaz, line 90: I'd suggest dropping the `GetOperatorSpec()` from the interface and always use the `operatorClient` instead to remove the need for unnecessary code duplication.

pkg/console/status/auth_status.go
```diff
@@ -94,69 +94,133 @@ func (c *AuthStatusHandler) Apply(ctx context.Context, authnConfig *configv1.Aut
 	if err != nil {
 		return err
 	}
+	existingClientStatusIndex, existingClientStatus := c.getExistingClientStatus(applyConfig)
```
stlaz, line 97: This looks incorrect, you shouldn't need to worry about any indices at all.

Use server-side apply so that you don't have to care about any indices. Pass a controller name to `AuthStatusHandler`. Each controller is taking care of their own client.

You can see that the slice is ready to be handled that way if you look at the API: https://github.com/openshift/api/blob/master/config/v1/types_authentication.go#L115-L117

Read more about SSA at https://kubernetes.io/docs/reference/using-api/server-side-apply/ and/or ask me.

The diff looks much bigger than I expected given we only want to parametrize componentName/componentNamespace. Is it because the above principle was not well understood prior to writing the code?

stlaz (changes requested):

pkg/console/subresource/crd/authentications.go
```diff
@@ -0,0 +1,33 @@
+package crd
+
+import (
+	"fmt"
+
+	apiexensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apiexensionsv1listers "k8s.io/apiextensions-apiserver/pkg/client/listers/apiextensions/v1"
+)
+
+func AuthnConfigHasOIDCFields(crdLister apiexensionsv1listers.CustomResourceDefinitionLister) (bool, error) {
```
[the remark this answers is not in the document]
stlaz: ignore this comment, I just realized we actually need to backport this

pkg/console/status/auth_status.go
TheRealJon: Yes, I had a limited understanding of SSA. My first pass at this was a much smaller change, but caused an infinite loop. I didn't realize the cause, which you pointed out, is the duplicate field manager.

pkg/console/controllers/util/util.go
TheRealJon: It's not really different, we could just pass the operatorClient as a third arg to HandleMangementState and remove this func from the interface.

stlaz (commented):
I don't see any serious issue but I think we may want to consider a bit better API around the login command.

pkg/console/subresource/authentication/cluster.go
```diff
@@ -0,0 +1,45 @@
...
+	if provider == nil || clientConfig == nil || provider.Issuer.URL == "" {
+		return ""
+	}
+
+	extraScopes := ""
+	if len(clientConfig.ExtraScopes) > 0 {
+		extraScopes = fmt.Sprintf(" --extra-scopes %s", strings.Join(clientConfig.ExtraScopes, ","))
+	}
+
+	return fmt.Sprintf("oc login %s --issuer-url %s --exec-plugin oc-oidc --client-id %s%s", apiServerURL, provider.Issuer.URL, clientConfig.ClientID, extraScopes)
```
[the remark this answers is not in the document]
TheRealJon: This was discussed at first, but decided against as a more complex solution that wasn't really necessary for the current need. At this point, it's probably too late in the game to change as it would require pretty significant changes here and in console.

pkg/console/subresource/authentication/cluster.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"fmt"
+	"strings"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/console-operator/pkg/api"
+	"golang.org/x/exp/slices"
+)
+
+func GetOIDCOCLoginCommand(authConfig *configv1.Authentication, apiServerURL string) string {
```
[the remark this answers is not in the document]
TheRealJon: My preference is to keep resource-specific utils decoupled from the controllers as much as possible, even if they are only used once. It doesn't cost anything now, but could save work or repetitive code in the future.

pkg/console/subresource/authentication/cluster.go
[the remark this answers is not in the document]
jhadvig: I agree, lets add it as a followUp

jhadvig:
@TheRealJon we need to do a rebase

TheRealJon:
/retest

install failed

PR #846, opened 2024-01-24
OCPBUGS-25484: OAuthClientsController: Fix warmlooping on status updates

stlaz:
The OAuthClientsController uses a status handler that's not updated on every sync but it's rather re-used. The status handler did not flush the status updating changes in `FlushAndReturn()` and so on every sync there would be updates from all previous loops, too.

/assign @jhadvig 
/cc @deads2k


jhadvig (approved):
/lgtm
/approve

stlaz:
/hold
/close
in favour of https://github.com/openshift/console-operator/pull/848

PR #847, opened 2024-01-24
OAuthClients: only add `oidcClients` to authentication status if the field exists

stlaz:
/assign @jhadvig 
/cc @deads2k 
/cc @xingxingxia


yanpzhan:
Console operator doesn't degrade on the cluster launched against the pr now.
/label qe-approved

pkg/console/controllers/oauthclients/oauthclients.go
```diff
@@ -193,9 +195,12 @@ func (c *oauthClientsController) sync(ctx context.Context, controllerContext fac
...
-	// AuthStatusHandler manages fields that are behind the CustomNoUpgrade and TechPreviewNoUpgrade featuregate sets
-	// call Apply() only if they are enabled, otherwise server-side apply will complain
-	if featureGates.Spec.FeatureSet == configv1.TechPreviewNoUpgrade || featureGates.Spec.FeatureSet == configv1.CustomNoUpgrade {
+	oidcClientsSchema, err := authnConfigHasOIDCFields(c.crdLister)
+	if err != nil {
+		return err
+	}
+
+	if oidcClientsSchema {
 		if err := c.authStatusHandler.Apply(ctx, authnConfig); err != nil {
```
deads2k, line 204: why can't this just move into the block with `configv1.AuthenticationTypeOIDC` a few lines up?

stlaz:
/hold
won't work with both HCP and OCP

deads2k:
> won't work with both HCP and OCP

Agreed.  if the spec includes eitehr an OIDC config or a clietn config, then it's active.

stlaz:
/remove-label qe-approved
The PR no longer fixes the issue observed by the QE where the operator gets degraded after the switch to TechPreview.

We still need the PR so that the feature works in HCP properly, as the feature is not featuregated there.

deads2k:
will now work on hcp and ocp

/hold cancel

jhadvig:
Re-adding the qe-approved label manually.
Adding other label as well.

stlaz:
/retest-required
build log:
"Contained sensitive information -- content removed."
eh

PR #848, opened 2024-01-24
OCPBUGS-25484: make it impossible double set conditions in a single loop

deads2k:
This likely fixes the pathological lastTransitionTime update problem, but does NOT fix the hotlooping that we're seeing from the operator.


deads2k:
/override ci/prow/e2e-azure-ovn-upgrade 

failed on the baremetal bug

PR #855, opened 2024-02-02
OCPBUGS-28982: oauthclients: fix oauthclients degraded condition that never gets removed

stlaz:
/assign @jhadvig


jhadvig (approved):
/lgtm

PR #857, opened 2024-02-05
CONSOLE-3912: always sync oidcClients in authn status if the field is present

stlaz:
Makes it bit easier on GitOpts and Hypershift people to sync the resource.

/assign @deads2k


pkg/console/controllers/oauthclients/oauthclients.go
```diff
@@ -163,47 +173,42 @@ func (c *oauthClientsController) sync(ctx context.Context, controllerContext fac
...
-			syncErr = oauthErr
-			break
+			return statusHandler.FlushAndReturn(oauthErr)
 		}
 
 	case configv1.AuthenticationTypeOIDC:
 		syncErr = c.syncAuthTypeOIDC(ctx, controllerContext, statusHandler, operatorConfig, authnConfig)
 		if syncErr != nil {
-			break
+			return statusHandler.FlushAndReturn(syncErr)
```
deads2k, line 194: Because this combines multiple fields, this early return breaks the AuthStatusHandler. do you want a separate controller?

deads2k:
it's an improvement, but I think it still has potential issues.

/lgtm
/approve
/hold

@stlaz you can decide whether we take this intermediate and fix better later or fix it all now.

stlaz:
/hold cancel
I choose going with the current approach.

This contributes to the tech debt in the operator by reusing its current pattern. I believe this issue should be fixed on a global scale in this repo instead.

cc @jhadvig

jhadvig (approved):
/approve
/retest

jhadvig:
Adding rest of the approve labels manually since this PR is mainly fixing issues in the eternal OIDC feature

PR #861, opened 2024-02-08
ouathclients: further split oidc and integrated oauth handling

stlaz:
HCP installs directly with OIDC support and as such removes the oauth-apiserver. It then wants to sync its config to the hosted authentication/cluster but it can't because the console-operator needs to set the clients in the status. However, the console-operator short-circuits on an error while querying the oauth APIs.

Split the controller to separate the fail modes in integrated oauth and OIDC config.

/assign @jhadvig 
/cc @deads2k


pkg/console/controllers/oauthclients/oauthclients.go
```diff
@@ -116,42 +96,45 @@ func NewOAuthClientsController(
...
+		// if we're not using integrated oauth, reset all degraded conditions
+		statusHandler.AddConditions(status.HandleProgressingOrDegraded("OAuthClientSecretSync", "", nil))
+		statusHandler.AddConditions(status.HandleProgressingOrDegraded("OAuthClientSync", "", nil))
+		return statusHandler.FlushAndReturn(nil)
+	}
+
+	operatorConfig, err := c.consoleOperatorLister.Get(api.ConfigResourceName)
 	if err != nil {
 		return err
 	}
```
deads2k, line 136: Add a check to be sure the `authnConfig.Spec.Type` check is AuthenticationTypeOIDC or reset all the conditions you're setting here and return.

pkg/console/controllers/oauthclients/oauthclients.go
```diff
@@ -58,32 +51,23 @@ type oauthClientsController struct {
...
 	targetNSSecretsLister       corev1listers.SecretLister
-	configNSSecretsLister       corev1listers.SecretLister
-	targetNSDeploymentsLister   appsv1listers.DeploymentLister
-	targetNSConfigLister        corev1listers.ConfigMapLister
 
-	authStatusHandler status.AuthStatusHandler
+	authStatusHandler *status.AuthStatusHandler
 }
 
 func NewOAuthClientsController(
```
deads2k, line 61: godoc telling me which things this writes?  

```go
// writes .status.condition[thing]
// writes .status.oidcclient[thing]
// writes secret -n <namespace> name
```

pkg/console/controllers/oauthclients/oauthclients.go
deads2k: > Add a check to be sure the `authnConfig.Spec.Type` check is AuthenticationTypeOIDC or reset all the conditions you're setting here and return.

I think I actually assumed this backwards.  https://github.com/openshift/console-operator/pull/861/files?diff=split&w=0#r1483192431 will help

pkg/console/oidcsetup/oidcsetup.go
```diff
@@ -0,0 +1,283 @@
...
+	if err != nil {
+		return err
+	}
+
+	authnConfig, err := c.authnLister.Get(api.ConfigResourceName)
+	if err != nil {
+		return err
+	}
+
+	if authnConfig.Spec.Type == configv1.AuthenticationTypeOIDC {
```
deads2k, line 119: if not equal, shouldn't you clear all conditions set by this control loop

TheRealJon (approved):
/lgtm

pkg/console/controllers/oauthclientsecret/oauthclientsecret.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+func (c *oauthClientSecretController) syncSecret(ctx context.Context, clientSecret string, recorder events.Recorder) error {
+	operatorConfig, err := c.consoleOperatorLister.Get(api.ConfigResourceName)
+	if err != nil {
+		return err
+	}
+
+	secret, err := c.targetNSSecretsLister.Secrets(api.TargetNamespace).Get("console-oauth-config")
+	if apierrors.IsNotFound(err) || secretsub.GetSecretString(secret) == clientSecret {
```
deads2k, line 151:
```suggestion
	if apierrors.IsNotFound(err) || secretsub.GetSecretString(secret) != clientSecret {
```

?

pkg/console/controllers/oauthclients/oauthclients.go
```diff
@@ -116,42 +83,45 @@ func NewOAuthClientsController(
...
 	if err != nil {
 		return err
 	}
 
-	ingressConfig, err := c.ingressConfigLister.Get(api.ConfigResourceName)
+	switch authnConfig.Spec.Type {
+	case "", configv1.AuthenticationTypeIntegratedOAuth:
+	default:
+		// if we're not using integrated oauth, reset all degraded conditions
+		statusHandler.AddConditions(status.HandleProgressingOrDegraded("OAuthClientSecretSync", "", nil))
```
deads2k, line 114: This appears to be controlled in NewOAuthClientSecretController

pkg/console/controllers/oauthclientsecret/oauthclientsecret.go
```diff
@@ -0,0 +1,179 @@
...
+			statusHandler.AddConditions(status.HandleProgressingOrDegraded("OAuthClientSecretSync", "FailedClientSecretGet", err))
+			return statusHandler.FlushAndReturn(err)
+		}
+
+		secretString = secretsub.GetSecretString(conficClientSecret)
+		if len(secretString) == 0 {
+			statusHandler.AddConditions(status.HandleProgressingOrDegraded("OAuthClientSecretSync", "ClientSecretKeyMissing", fmt.Errorf("missing the 'clientSecret' key in the client secret secret %q", clientConfig.ClientSecret.Name)))
+			return statusHandler.FlushAndReturn(nil)
+		}
+	default:
```
deads2k, line 134: reset the OAuthClientSecretSync conditions in here.

pkg/console/oidcsetup/oidcsetup.go
```diff
@@ -0,0 +1,283 @@
...
+		err = c.syncAuthTypeOIDC(ctx, syncCtx, statusHandler, operatorConfig, authnConfig)
+		if err != nil {
+			return statusHandler.FlushAndReturn(err)
```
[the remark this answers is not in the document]
deads2k: > if I hit this path, doesn't returning here prevent this loop from setting the .status.oidcclient stanzas required to understand which oidcprovider the console is currently using?

This looks to still be an issue with this code.

pkg/console/oidcsetup/oidcsetup.go
deads2k: > if not equal, shouldn't you clear all conditions set by this control loop

This looks to still be an issue

pkg/console/oidcsetup/oidcsetup.go
```diff
@@ -0,0 +1,262 @@
...
+	if err != nil {
+		return false, "", err
+	}
+
+	deplAvailableUpdated := deploymentsub.IsAvailableAndUpdated(depl)
+	if !deplAvailableUpdated {
+		return false, "deployment unavailable or outdated", nil
+	}
+
+	if clientSecret.GetResourceVersion() != depl.ObjectMeta.Annotations["console.openshift.io/oauth-secret-version"] {
```
deads2k, line 192: this looks questionable. ResourceVersion rarely matches since anything can do a read/write cycle or annotate for instance.

deads2k: Generation or hash is more suitable.  This might work well enough if that's how the annotation is currently set, but it's vulnerable and ought to be fixed in a z

pkg/console/oidcsetup/oidcsetup.go
deads2k: maybe not all of them, but OIDCClientConfig

pkg/console/oidcsetup/oidcsetup.go
```diff
@@ -0,0 +1,262 @@
...
+		return err
+	}
+
+	authnConfig, err := c.authnLister.Get(api.ConfigResourceName)
+	if err != nil {
+		return err
+	}
+
+	if authnConfig.Spec.Type == configv1.AuthenticationTypeOIDC {
+		err = c.syncAuthTypeOIDC(ctx, syncCtx, statusHandler, operatorConfig, authnConfig)
```
deads2k, line 109: sticky.  I think I finally get this and I recommend some factorization updates.  The conditional part of this logic is "get the clientID if available in spec to be provided in status" and "if the clientID is available, determine the status of using this clientID".

I'm wondering if most of this function control loop is actually

```go
func sync(){
  return if not managed
  if !oidcClientsSchema{
    clear all conditions
    return
  }

  if .spec.type != OIDC{
    apply an empty authentication.config.openshift.io .status.oidcClients[componentNamespace=openshift-console,componentName=console]
    return
  } 

  read current authentication .spec.oidc.oidcClient[console]
  check status of rolling out the clientid 
  apply complete authentication.config.openshift.io .status.oidcClients[componentNamespace=openshift-console,componentName=console]

  which would eliminate the if-blocks-with-continue that made this difficult to read.
}
```

pkg/console/controllers/oauthclients/oauthclients.go
```diff
@@ -116,42 +83,45 @@ func NewOAuthClientsController(
...
+
+	authnConfig, err := c.authnLister.Get(api.ConfigResourceName)
 	if err != nil {
 		return err
 	}
 
-	ingressConfig, err := c.ingressConfigLister.Get(api.ConfigResourceName)
+	switch authnConfig.Spec.Type {
+	case "", configv1.AuthenticationTypeIntegratedOAuth:
+	default:
```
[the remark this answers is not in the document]
stlaz: The API might be gone already at that point

pkg/console/controllers/oauthclients/oauthclients.go
stlaz: good catch

pkg/console/controllers/oauthclientsecret/oauthclientsecret.go
stlaz: :facepalm:

pkg/console/oidcsetup/oidcsetup.go
stlaz: I rewrote pieces of it so that this error does not make the loop finish early.

pkg/console/oidcsetup/oidcsetup.go
stlaz: I'd keep it for now. We set the annotation this way and the secret is controlled by us.

pkg/console/controllers/oauthclients/oauthclients.go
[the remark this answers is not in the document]
deads2k: > The API might be gone already at that point

Accepted, but make a bug to come back to by 4.17 to cleanup if available.

pkg/console/oidcsetup/oidcsetup.go
deads2k: > I'd keep it for now. We set the annotation this way and the secret is controlled by us.

Needs bug to correct by 4.17.  annotations can be set by nearly anyone and we share them, so the RV isn't guaranteed to match.

pkg/console/controllers/oidcsetup/oidcsetup.go
```diff
@@ -0,0 +1,280 @@
...
+	operatorConfig, err := c.consoleOperatorLister.Get(api.ConfigResourceName)
+	if err != nil {
+		return err
+	}
+
+	authnConfig, err := c.authnLister.Get(api.ConfigResourceName)
+	if err != nil {
+		return err
+	}
+
```
deads2k, line 122: Reduce nesting by 

```go
  if .spec.type != OIDC{
    apply an empty authentication.config.openshift.io .status.oidcClients[componentNamespace=openshift-console,componentName=console]
    return
  } 
```

pkg/console/controllers/oidcsetup/oidcsetup.go
```diff
@@ -0,0 +1,280 @@
...
+	// we need to keep track of errors during the sync so that we can requeue
+	// if any occur
+	var errs []error
+	if authnConfig.Spec.Type == configv1.AuthenticationTypeOIDC {
+		syncErr := c.syncAuthTypeOIDC(ctx, syncCtx, statusHandler, operatorConfig, authnConfig)
```
deads2k, line 127: without the nesting, this entire function can be inlined to avoid the authSTatus side effect.

And once that is done, authStatus becomes a per-sync function variable, not an instance variable which better represents the state and prevents accidents.

pkg/console/controllers/oidcsetup/oidcsetup.go
```diff
@@ -0,0 +1,280 @@
...
+
+	clientSecret, err := c.targetNSSecretsLister.Secrets(api.TargetNamespace).Get("console-oauth-config")
+	if err != nil {
+		c.authStatusHandler.Degraded("OIDCClientSecretGet", err.Error())
+		return err
+	}
+
+	if valid, msg, err := c.checkClientConfigStatus(authnConfig, clientSecret); err != nil {
+		c.authStatusHandler.Degraded("DeploymentOIDCConfig", err.Error())
+		return err
```
deads2k, line 184: I haven't yet looked, but I bet that this will cause the progressing and available conditions on the authentication.status.oidcclients[console] to disappear.

stlaz: https://github.com/openshift/console-operator/blob/68e37a85d0d3a083c396f7654707b49b72553c49/pkg/console/status/auth_status.go#L119-L127 should be handling that I beleive

pkg/console/controllers/oauthclientsecret/oauthclientsecret.go
```diff
@@ -0,0 +1,188 @@
...
+		secretsClient:  secretsClient,
+
+		authConfigLister:      authnInformer.Lister(),
+		consoleOperatorLister: consoleOperatorInformer.Lister(),
+		configSecretsLister:   configSecretsInformer.Lister(),
+		targetNSSecretsLister: targetNSsecretsInformer.Lister(),
+	}
+
+	return factory.New().
+		WithSync(c.sync).
```
deads2k, line 77: missing `ResyncEvery(wait.Jitter(time.Minute, 1.0)).`

pkg/console/controllers/oauthclientsecret/oauthclientsecret.go
```diff
@@ -0,0 +1,188 @@
...
+			secretString = secretsub.GetSecretString(clientSecret)
+		}
+		if len(secretString) == 0 {
+			secretString = crypto.Random256BitsString()
+		}
+	case configv1.AuthenticationTypeOIDC:
+		clientConfig := utilsub.GetOIDCClientConfig(authConfig)
+		if clientConfig == nil {
+			// no config, flush the condition and return
+			statusHandler.AddConditions(status.HandleProgressingOrDegraded("OAuthClientSecretSync", "", nil))
```
deads2k, line 122: in 4.16, this case needs to make the desired configuration real.  if the desired configuration is "nothing" break it.

pkg/console/controllers/oidcsetup/oidcsetup.go
deads2k: If we take this early return, what sets the Progressing and Available condition since the authStatusHandler self-clears the conditions here: on every apply.

I typed all the way to here before realizing that the combination of stateful, plus early return on error protected us.

this really needs to be transitioned to per-invocation scoped, not instance scoped for ease of reading+maintenance.  I'm pretty good at this and I can't read it.

pkg/console/controllers/oidcsetup/oidcsetup.go
```diff
@@ -0,0 +1,303 @@
...
+		status.HandleProgressingOrDegraded(
+			"OIDCClientConfig", "OIDCConfigSyncFailed",
+			syncErr,
+		),
+	)
+	if syncErr != nil {
+		errs = append(errs, syncErr)
+	}
+
+	applyErr := c.authStatusHandler.Apply(ctx, authnConfig)
```
deads2k, line 162: if the syncError != nil, when we get here we won't have all the conditions specified.

deads2k:

that job history is really bad, I'll chance this one since it's not likely to be testing it anyway, but in 4.15 will need a better eye on it.

PR #862, opened 2024-02-09
OCPBUGS-28541: Remove Janus IDP and update RHDH quickstarts

jerolimov:
1. Removes the complete Janus IDP Quick starts and the references to Janus IDP from the RHDH quick starts.
2. Added a note that the RHDH Quick start might not be available in the Operator Hub and that users might use the Helm chart instead.

Screenshot of that note (esp. because a `<small>` tag was needed:

image

/cc @vikram-raj @jhadvig @opayne1

PR #863, opened 2024-02-09
[release-4.15] OCPBUGS-29332: Remove Janus IDP and update RHDH quickstarts

jerolimov:
This is a manual backport (to run the tests in parallel) of #862 to update these Quick Starts in 4.15 as well.

1. Removes the complete Janus IDP Quick starts and the references to Janus IDP from the RHDH quick starts.
2. Added a note that the RHDH Quick start might not be available in the Operator Hub and that users might use the Helm chart instead.

Screenshot of that note (esp. because a `<small>` tag was needed:

image

/cc @vikram-raj @jhadvig @opayne1


spadgett (approved):
/lgtm

PR #869, opened 2024-02-15
OCPBUGS-29479: Add startupProbe to console container.

TheRealJon:
Use startup probe on console container to account for possible long start times from auth initialization. This will help to avoid unnecessary restarts of the console pod.

### DIFF
--- a/bindata/assets/deployments/console-deployment.yaml
+++ b/bindata/assets/deployments/console-deployment.yaml
@@ -38,15 +38,6 @@ spec:
             requests:
               cpu: 10m
               memory: 100Mi
-          readinessProbe:
-            httpGet:
-              path: /health
-              port: 8443
-              scheme: HTTPS
-            timeoutSeconds: 1
-            periodSeconds: 10
-            successThreshold: 1
-            failureThreshold: 3
           lifecycle:
             preStop:
               exec:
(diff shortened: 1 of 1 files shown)


TheRealJon:
cc @sdodson

bindata/assets/deployments/console-deployment.yaml
```diff
@@ -70,11 +70,15 @@ spec:
               path: /health
               port: 8443
               scheme: HTTPS
-            initialDelaySeconds: 150
-            timeoutSeconds: 1
+            failureThreshold: 1
             periodSeconds: 10
-            successThreshold: 1
-            failureThreshold: 3
+          startupProbe:
```
deads2k, line 75: I like the idea of this change.  This makes the startup wait 5 minutes instead of 2.5 minutes. That seems ok to me if you find it reasonable, but I want to be sure the difference is obvious.

bindata/assets/deployments/console-deployment.yaml
```diff
@@ -70,11 +70,15 @@ spec:
               path: /health
               port: 8443
               scheme: HTTPS
-            initialDelaySeconds: 150
```
deads2k, line 73: Glad to see this changed since the startup probe means that as soon as the server is healthy it will properly handled.

bindata/assets/deployments/console-deployment.yaml
deads2k, line 75: I recommend explicitly setting all the various values

```
      initialDelaySeconds: 0
      periodSeconds: 5
      timeoutSeconds: 10
      successThreshold: 1
      failureThreshold: 30
```

I'm less critical of the content, but future owners will likely thank you for explaining their various options even if they are the defaults.

jhadvig (approved):
/lgtm

bindata/assets/deployments/console-deployment.yaml
TheRealJon: Yes, this was intentional. In the worst case, the console will take 5 minutes to start. See https://github.com/openshift/console/pull/13610

TheRealJon:
@deads2k @jhadvig I added default values to all the probes and re-ordered the fields to be a bit more consistent and logical.

PR #871, opened 2024-02-26
AUTH-482: set required-scc for openshift workloads

liouk:


liouk:
/hold
need to verify SCC

jhadvig (changes requested):
@liouk could you please update the unit tests accordingly ?

jhadvig (approved):
/lgtm
/approve

jhadvig:
QE Approver:
/assign @yapei 
Docs Approver:
/assign @opayne1 
PX Approver:
/assign @RickJWagner

bindata/assets/deployments/console-deployment.yaml
```diff
@@ -21,6 +21,7 @@ spec:
         component: ui
       annotations:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
+        openshift.io/required-scc: restricted-v2
```
yapei, line 24: sorry @liouk  May I know why it is pinned to `restricted-v2` instead of other default SCCs?

liouk: The console pods are anyway getting admitted with `restricted-v2`, as they don't need more privileged access. Here we are just pinning the default SCC that the pods are already getting admitted with.

opayne1:
Verified with @bergerhoffer 

/label docs-approved

PR #872, opened 2024-02-26
OCPBUGS-6208: Updating openshift-enterprise-console-operator images to be consistent with ART

jhadvig:
Clone of https://github.com/openshift/console-operator/pull/702 casue its missing bindata update.


jhadvig:
Adding all the added labels from https://github.com/openshift/console-operator/pull/702 manually.

PR #883, opened 2024-04-02
Bump k8s.io/component-base to latest

TheRealJon:
Update "k8s.io/component-base" so that indirect dependency, "google.golang.org/protobuf", also gets updated.


TheRealJon:
Opened against the wrong bug. See #884

PR #886, opened 2024-04-09
OCPBUGS-33787: Tolerate the absence of ingress capability on HyperShift clusters

alebedev87:
Part of the implementation of [Make Ingress optional for HyperShift EP](https://github.com/openshift/enhancements/pull/1415). Note that this change targets only HyperShift managed deployments and should not impact standalone OpenShift installations.

This PR:
- Implements [alternative ingress fields](https://github.com/openshift/api/pull/1887) from the console operator config API
- Skips component route customizations if ingress capability is disabled
- Uses NodePort type for console and downloads services if ingress capability is disabled
- Adds document to describe how to implement alternative ingress on ROSA

### DIFF
--- a/docs/alb-ingress-rosa-hcp.md
+++ b/docs/alb-ingress-rosa-hcp.md
@@ -1,0 +1,115 @@
+# Use AWS ALB as alternative ingress on ROSA HCP
+
+This doc aims at showing the effort needed to expose the OpenShift console via AWS ALB on a ROSA HCP cluster.
+The use case in mind is [HyperShift hosted clusters where the Ingress capability is disabled](https://github.com/openshift/enhancements/pull/1415).
+
+## Requirements
+
+- ROSA HCP OpenShift cluster.
+- [AWS Load Balancer Operator installed and its controller created](https://docs.openshift.com/rosa/networking/aws-load-balancer-operator.html).
+- User logged as a cluster admin.
+
+## Procedure
+
+### Create certificate in AWS Certificate Manager
+
+In order to configure an HTTPS listener on AWS ALB you need to have a certificate created in AWS Certificate Manager.
+You can import an existing certificate or request a new one. Make sure the certificate is created in the same region as your cluster.
+Note the certificate ARN, you will need it later.
+
+### Create Ingress resources for the NodePort services
+
+To provision ALBs create the following resources:
+```bash
+cat <<EOF | oc apply -f -
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  annotations:
+    alb.ingress.kubernetes.io/scheme: internet-facing
+    alb.ingress.kubernetes.io/target-type: instance
+    alb.ingress.kubernetes.io/backend-protocol: HTTPS
+    alb.ingress.kubernetes.io/certificate-arn: ${CERTIFICATE_ARN}
+  name: console
+  namespace: openshift-console
+spec:
+  ingressClassName: alb
+  rules:
+    - http:
+        paths:
+          - path: /
+            pathType: Prefix
+            backend:
+              service:
+                name: console
+                port:
+                  number: 443
+---
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  annotations:
+    alb.ingress.kubernetes.io/scheme: internet-facing
+    alb.ingress.kubernetes.io/target-type: instance
+    alb.ingress.kubernetes.io/backend-protocol: HTTP
+    alb.ingress.kubernetes.io/certificate-arn: ${CERTIFICATE_ARN}
+  name: downloads
+  namespace: openshift-console
+spec:
+  ingressClassName: alb
+  rules:
+    - http:
+        paths:
+          - path: /
+            pathType: Prefix
+            backend:
+              service:
+                name: downloads
+                port:
+                  number: 80
+EOF
+```
+
+### Update console config
+
+Once the console ALBs are ready you need to let the console operator know which urls to use.
+Update the console operator config providing the custom urls:
+```bash
+$ CONSOLE_ALB_HOST=$(oc -n openshift-console get ing console -o yaml | yq .status.loadBalancer.ingress[0].hostname)
+$ DOWNLOADS_ALB_HOST=$(oc -n openshift-console get ing downloads -o yaml | yq .status.loadBalancer.ingress[0].hostname)
+$ oc patch console.operator.openshift.io cluster --type=merge -p "{\"spec\":{\"ingress\":{\"consoleURL\":\"https://${CONSOLE_ALB_HOST}\",\"clientDownloadsURL\":\"https://${DOWNLOADS_ALB_HOST}\"}}}"
+```
+
+## Notes
+
+1. ROSA HCP does not have the authentication operator, the authentication server is managed centrally by the HyperShift layer:
+```bash
+$ oc -n openshift-authentication-operator get deploy,route
+No resources found
+
+$ oc -n openshift-authentication get pods,routes
+No resources found
+
+$ oc get oauthclient | grep -v console
+NAME                           SECRET                                        WWW-CHALLENGE   TOKEN-MAX-AGE   REDIRECT URIS
+openshift-browser-client                                                     false           default         https://oauth.mytestcluster.5199.s3.devshift.org:443/oauth/token/display
+openshift-challenging-client                                                 true            default         https://oauth.mytestcluster.5199.s3.devshift.org:443/oauth/token/implicit
+
+$ oc -n openshift-console rsh deploy/console curl -k https://openshift.default.svc/.well-known/oauth-authorization-server
+{
+"issuer": "https://oauth.mytestcluster.5199.s3.devshift.org:443",
+"authorization_endpoint": "https://oauth.mytestcluster.5199.s3.devshift.org:443/oauth/authorize",
+"token_endpoint": "https://oauth.mytestcluster.5199.s3.devshift.org:443/oauth/token",
+```
+
+2. When the ingress capability is disabled, the console operator relies on the end user to provide the console and download URLs (using the operator API) for health checks and oauthclient.
+
+3. When the ingress capability is disabled, the console operator skips the implementation of the component route customization.
+
+4. To simulate the absence of ingress connectivity when the ingress capability is disabled, set the desired replicas to zero in the default ingress controller:
+```bash
+$ oc -n openshift-ingress-operator patch ingresscontroller default --type='json' -p='[{"op": "replace", "path": "/spec/replicas", "value":0}]'
+```
+
+## Links
+- [Demo of ALB ingress for the console on ROSA HCP](https://drive.google.com/file/d/1uWZgFbSeZTlDzlFyPW7QcH-625JsbSbw/view)
--- a/pkg/console/controllers/clidownloads/controller.go
+++ b/pkg/console/controllers/clidownloads/controller.go
@@ -4,6 +4,7 @@ import (
 	// standard lib
 	"context"
 	"fmt"
+	"net/url"
 	"time"
 
 	// kube
@@ -114,25 +115,37 @@ func (c *CLIDownloadsSyncController) Sync(ctx context.Context, controllerContext
 	}
 
 	statusHandler := status.NewStatusHandler(c.operatorClient)
-	ingressConfig, err := c.ingressConfigLister.Get(api.ConfigResourceName)
-	if err != nil {
-		return statusHandler.FlushAndReturn(err)
-	}
 
-	activeRouteName := api.OpenShiftConsoleDownloadsRouteName
-	routeConfig := routesub.NewRouteConfig(updatedOperatorConfig, ingressConfig, activeRouteName)
-	if routeConfig.IsCustomHostnameSet() {
-		activeRouteName = api.OpenshiftDownloadsCustomRouteName
-	}
+	var (
+		downloadsURI *url.URL
+		downloadsErr error
+	)
+	if len(operatorConfig.Spec.Ingress.ClientDownloadsURL) == 0 {
+		ingressConfig, err := c.ingressConfigLister.Get(api.ConfigResourceName)
+		if err != nil {
+			return statusHandler.FlushAndReturn(err)
+		}
 
-	downloadsRoute, downloadsRouteErr := c.routeLister.Routes(api.TargetNamespace).Get(activeRouteName)
-	if downloadsRouteErr != nil {
-		return downloadsRouteErr
-	}
+		activeRouteName := api.OpenShiftConsoleDownloadsRouteName
+		routeConfig := routesub.NewRouteConfig(updatedOperatorConfig, ingressConfig, activeRouteName)
+		if routeConfig.IsCustomHostnameSet() {
+			activeRouteName = api.OpenshiftDownloadsCustomRouteName
+		}
 
-	downloadsURI, _, downloadsRouteErr := routeapihelpers.IngressURI(downloadsRoute, downloadsRoute.Spec.Host)
-	if downloadsRouteErr != nil {
-		return downloadsRouteErr
+		downloadsRoute, downloadsRouteErr := c.routeLister.Routes(api.TargetNamespace).Get(activeRouteName)
+		if downloadsRouteErr != nil {
+			return downloadsRouteErr
+		}
+
+		downloadsURI, _, downloadsErr = routeapihelpers.IngressURI(downloadsRoute, downloadsRoute.Spec.Host)
+		if downloadsErr != nil {
+			return downloadsErr
+		}
+	} else {
+		downloadsURI, downloadsErr = url.Parse(operatorConfig.Spec.Ingress.ClientDownloadsURL)
+		if downloadsErr != nil {
+			return fmt.Errorf("failed to parse downloads url: %w", downloadsErr)
+		}
 	}
 
 	ocConsoleCLIDownloads := PlatformBasedOCConsoleCLIDownloads(downloadsURI.String(), api.OCCLIDownloadsCustomResourceName)
--- a/pkg/console/controllers/oauthclients/oauthclients.go
+++ b/pkg/console/controllers/oauthclients/oauthclients.go
@@ -3,6 +3,7 @@ package oauthclients
 import (
 	"context"
 	"fmt"
+	"net/url"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
@@ -131,15 +132,26 @@ func (c *oauthClientsController) sync(ctx context.Context, controllerContext fac
 		return err
 	}
 
-	routeName := api.OpenShiftConsoleRouteName
-	routeConfig := routesub.NewRouteConfig(operatorConfig, ingressConfig, routeName)
-	if routeConfig.IsCustomHostnameSet() {
-		routeName = api.OpenshiftConsoleCustomRouteName
-	}
+	var consoleURL *url.URL
 
-	_, consoleURL, _, routeErr := routesub.GetActiveRouteInfo(c.routesLister, routeName)
-	if routeErr != nil {
-		return routeErr
+	if len(operatorConfig.Spec.Ingress.ConsoleURL) == 0 {
+		routeName := api.OpenShiftConsoleRouteName
+		routeConfig := routesub.NewRouteConfig(operatorConfig, ingressConfig, routeName)
+		if routeConfig.IsCustomHostnameSet() {
+			routeName = api.OpenshiftConsoleCustomRouteName
+		}
+
+		_, url, _, routeErr := routesub.GetActiveRouteInfo(c.routesLister, routeName)
+		if routeErr != nil {
+			return routeErr
+		}
+		consoleURL = url
+	} else {
+		url, err := url.Parse(operatorConfig.Spec.Ingress.ConsoleURL)
+		if err != nil {
+			return fmt.Errorf("failed to parse console url: %w", err)
+		}
+		consoleURL = url
 	}
 
 	waitCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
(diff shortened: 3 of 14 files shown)


pkg/console/controllers/clidownloads/controller.go
```diff
@@ -114,8 +121,27 @@ func (c *CLIDownloadsSyncController) Sync(ctx context.Context, controllerContext
 	}
 
 	statusHandler := status.NewStatusHandler(c.operatorClient)
+
+	infrastructureConfig, err := c.infrastructureClient.Get(ctx, api.ConfigResourceName, metav1.GetOptions{})
```
jhadvig, line 125: If the ingress can be enabled/disabled during the clusters lifecycle then we need to keep the fetching logic per controller, but we need to use listers rather then client.
If the ingress can be re-enabled then please move the fetching logic `starter.go`  and pass the config to each of the controllers.

pkg/console/controllers/route/controller.go
```diff
@@ -116,8 +124,26 @@ func (c *RouteSyncController) Sync(ctx context.Context, controllerContext factor
...
+	}
+
+	clusterVersionConfig, err := c.clusterVersionLister.Get("version")
+	if err != nil {
+		klog.Errorf("cluster version config error: %v", err)
+		return statusHandler.FlushAndReturn(err)
+	}
+
+	// Disable the route check for external control plane topology (hypershift) if the ingress capability is disabled.
+	if util.IsExternalControlPlaneWithIngressDisabled(infrastructureConfig, clusterVersionConfig) {
```
jhadvig, line 140: by doing this you are basically disabling the operator from doing its work by syncing the console routes.

pkg/console/controllers/oauthclients/oauthclients.go
```diff
@@ -108,6 +116,24 @@ func (c *oauthClientsController) sync(ctx context.Context, controllerContext fac
...
+
+	clusterVersionConfig, err := c.clusterVersionLister.Get("version")
+	if err != nil {
+		klog.Errorf("cluster version config error: %v", err)
+		return statusHandler.FlushAndReturn(err)
+	}
+
+	// Disable the oauth client update for external control plane topology (hypershift) if the ingress capability is disabled.
+	// HyperShift handles the updating of the oauth client for the console.
+	if util.IsExternalControlPlaneWithIngressDisabled(infrastructureConfig, clusterVersionConfig) {
```
jhadvig, line 133: by doing this you are basically disabling the operator from doing its work by syncing the console oauthclients.

pkg/console/controllers/clidownloads/controller.go
```diff
@@ -114,8 +121,27 @@ func (c *CLIDownloadsSyncController) Sync(ctx context.Context, controllerContext
...
+	}
+
+	clusterVersionConfig, err := c.clusterVersionLister.Get("version")
+	if err != nil {
+		klog.Errorf("cluster version config error: %v", err)
+		return statusHandler.FlushAndReturn(err)
+	}
+
+	// Disable the client download check for external control plane topology (hypershift) if the ingress capability is disabled.
+	if controllersutil.IsExternalControlPlaneWithIngressDisabled(infrastructureConfig, clusterVersionConfig) {
```
jhadvig, line 138: by doing this you are basically disabling the operator from doing its work by syncing the console CLI downloads CRDs.

jhadvig (changes requested):
Thanks @alebedev87 
This change does not make the operator to go downgraded when the ingress is disabled but whats basically because you are prematurely exiting all of the updated controllers. I doubt that the console will be working in this state.
Have you tested it ?

alebedev87:
>  I doubt that the console will be working in this state. Have you tested it ?

Yes. You are right, the console is not not working out of the box because the routes are left unserved. I was playing with [AWS Load Balancer Operator](https://github.com/openshift/enhancements/blob/master/enhancements/ingress/optional-ingress-hypershift.md#setup-aws-alb-as-the-new-ingress) (ALBO) to find the right `Ingress` configuration to make the console work without routes. I manged to find it but it sill has some gaps:
- the console service should be of type `NodePort` to work with the AWS ALB. This can be treated as a gap in the console operator which we can address later (on demand).
- the redirect to the oauth server (at the login time) still uses the oauth route despite the fact that I changed all `OAuth` clients to use the AWS ALBs provisioned by the aws-lb-controller. This part may be configurable though, I just me miss the console knowledge.

pkg/console/controllers/clidownloads/controller.go
alebedev87: On the standalone OpenShift the Ingress capability [will always be enabled](https://github.com/openshift/cluster-version-operator/pull/946). On HyperShift: once enabled it [cannot be disabled](https://github.com/openshift/enhancements/blob/master/enhancements/installer/component-selection.md#capabilities-cannot-be-uninstalled).

pkg/console/controllers/oauthclients/oauthclients.go
alebedev87: By "syncing the console oauthclients" you mean adding `redirectURI`? From what I understand this part will be handled by HyperShift as it takes care of some of the `OAuthClient`s already. I suppose it's because [the authentication server is not exposed via a route](https://github.com/openshift/enhancements/blob/master/enhancements/ingress/optional-ingress-hypershift.md#authentication-operator).

@csrwng:  What do you think? Should be disable the updates of console's `OAuthClient` on HyperShift or we can keep it for the time being?

pkg/console/controllers/route/controller.go
alebedev87: @jhadvig: From what I see both of the functions used by the route controller (`SyncCustomRoute`,`SyncDefaultRoute`) are related to the component route implementation (custom hostname or tls). The component routes are not supported on HyperShift, so I think we are safe to skip them.

@csrwng: can you please confirm that the component routes are not supported on HyperShift?

pkg/console/controllers/oauthclients/oauthclients.go
alebedev87: Also, in the absence of the ingress operator the required RBAC won't be created for the components (including the console operator): [Component routes chapter of the EP](https://github.com/openshift/enhancements/blob/master/enhancements/ingress/optional-ingress-hypershift.md#component-routes).

pkg/console/controllers/clidownloads/controller.go
alebedev87: Right, `oc-cli-downloads` contains the links to download the `oc` client from the download route:

```bash
$ oc get consoleclidownloads oc-cli-downloads -o yaml | yq .spec.links[].href
https://downloads-openshift-console.apps.alebedev-0424.devcluster.openshift.com/amd64/linux/oc.tar
https://downloads-openshift-console.apps.alebedev-0424.devcluster.openshift.com/amd64/mac/oc.zip
https://downloads-openshift-console.apps.alebedev-0424.devcluster.openshift.com/amd64/windows/oc.zip
https://downloads-openshift-console.apps.alebedev-0424.devcluster.openshift.com/arm64/linux/oc.tar
https://downloads-openshift-console.apps.alebedev-0424.devcluster.openshift.com/arm64/mac/oc.zip
https://downloads-openshift-console.apps.alebedev-0424.devcluster.openshift.com/ppc64le/linux/oc.tar
https://downloads-openshift-console.apps.alebedev-0424.devcluster.openshift.com/s390x/linux/oc.tar
https://downloads-openshift-console.apps.alebedev-0424.devcluster.openshift.com/oc-license
```

So, the risk is to have the broken links in `/command-line-tools` path of the console UI.

alebedev87: @csrwng : can we accept the risk I described above?

pkg/console/controllers/route/controller.go
csrwng: That is correct. If you've disabled ingress you are also indirectly disabling component routes, no?

alebedev87: Right, I put my followup comment into a wrong conversation. Here it is:

Also, in the absence of the ingress operator the required RBAC won't be created for the components (including the console operator): [Component routes chapter of the EP](https://github.com/openshift/enhancements/blob/master/enhancements/ingress/optional-ingress-hypershift.md#component-routes).

pkg/console/controllers/oauthclients/oauthclients.go
csrwng: Makes sense to disable updates of the OAuthClient

pkg/console/controllers/clidownloads/controller.go
csrwng: Yes

alebedev87:
Rebased from `master` to fix the merge conflict.

pkg/console/controllers/clidownloads/controller.go
jhadvig: Then we should be also prematurely exiting the DownloadsDeployment controller and [remove the deployment itself](https://github.com/openshift/console-operator/blob/master/pkg/console/controllers/downloadsdeployment/controller.go#L129).

jhadvig:
>Yes. You are right, the console is not not working out of the box because the routes are left unserved. I was playing with [AWS Load Balancer Operator](https://github.com/openshift/enhancements/blob/master/enhancements/ingress/optional-ingress-hypershift.md#setup-aws-alb-as-the-new-ingress) (ALBO) to find the right Ingress configuration to make the console work without routes. I managed to find it but it sill has some gaps:
> - the console service should be of type NodePort to work with the AWS ALB. This can be treated as a gap in the console operator which we can address later (on demand).

So this case will need to be handle in the service controller, which will then update the service spec accordingly. If I understand it, console itself wont be available though the service if the NodePort is not set? If thats the case I think it needs to be address in this change.

> - the redirect to the oauth server (at the login time) still uses the oauth route despite the fact that I changed all OAuth clients to use the AWS ALBs provisioned by the aws-lb-controller. This part may be configurable though, I may just miss some console knowledge.

Are these steps documented somewhere?

I feel like there is too many gaps - in order to merge this change. What would be the benefit for Hypershift user if he disables the Ingress capability, but the console wont be working? I feel that at minimum we should have a documented way how to make the console work.
Also this will need to be tested with QE before merging and they will need to know the steps how to:
1. setup the env (hypershift with disabled ingress capability)
after which they will test the non-functional console. In which case they will assume that its a bug. Therefor they will need a reproduction steps how to make console work.

spadgett:
> > I doubt that the console will be working in this state. Have you tested it ?
> 
> Yes. You are right, the console is not not working out of the box because the routes are left unserved. I was playing with [AWS Load Balancer Operator](https://github.com/openshift/enhancements/blob/master/enhancements/ingress/optional-ingress-hypershift.md#setup-aws-alb-as-the-new-ingress) (ALBO) to find the right `Ingress` configuration to make the console work without routes. I managed to find it but it sill has some gaps:
> 
> * the console service should be of type `NodePort` to work with the AWS ALB. This can be treated as a gap in the console operator which we can address later (on demand).
> * the redirect to the oauth server (at the login time) still uses the oauth route despite the fact that I changed all `OAuth` clients to use the AWS ALBs provisioned by the aws-lb-controller. This part may be configurable though, I may just miss some console knowledge.

Should we simply disable console for now if we know it won't work, even with manual load balancer configuration?

The console reads the OAuth URLs from the well-known endpoint. If the OAuth metadata can be updated in that document, console should discover the right URL.

https://docs.openshift.com/container-platform/4.15/authentication/configuring-internal-oauth.html#oauth-server-metadata_configuring-internal-oauth

alebedev87:
> Should we simply disable console for now if we know it won't work, even with manual load balancer configuration?

The console is still an important part of HyperShift based offerings like ROSA. The HyperShift engineers prefer the console to be running even without the default ingress (router). The way to expose it is a subject to a discussion, one of the ways to expose the console is described in [the doc I added today](https://github.com/openshift/console-operator/blob/ca595bff46855c31fc4c6204f13c726d9eaa5fbd/docs/alb-ingress-rosa-hcp.md).
However the console operator depends deeply on the router being present on the cluster. And we need to think pragmatically about what we can do in 4.16 release and what needs a follow up.

> The console reads the OAuth URLs from the well-known endpoint. If the OAuth metadata can be updated in that document, console should discover the right URL.

ROSA HCP clusters already have [the auth URL which is not an OpenShift route](https://github.com/openshift/console-operator/blob/ca595bff46855c31fc4c6204f13c726d9eaa5fbd/docs/alb-ingress-rosa-hcp.md#notes). This makes the authentication from console work smoothly even without the router.

spadgett:
> The console is still an important part of HyperShift based offerings like ROSA. The HyperShift engineers prefer the console to be running even without the default ingress (router). The way to expose it is a subject to a discussion, one of the ways to expose the console is described in [the doc I added today](https://github.com/openshift/console-operator/blob/ca595bff46855c31fc4c6204f13c726d9eaa5fbd/docs/alb-ingress-rosa-hcp.md). 

But I don't think this will work since the console will stomp those changes, unless you put it in unmanaged state.

> > The console reads the OAuth URLs from the well-known endpoint. If the OAuth metadata can be updated in that document, console should discover the right URL.
> 
> ROSA HCP clusters already have [the auth URL which is not an OpenShift route](https://github.com/openshift/console-operator/blob/ca595bff46855c31fc4c6204f13c726d9eaa5fbd/docs/alb-ingress-rosa-hcp.md#notes). This makes the authentication from console work smoothly even without the router.

That console doesn't really care whether it's a route or not. It just needs to be exposed in the well-known endpoint so it discovers the right URL. (We might need to double check how console gets the CA cert, however.)

alebedev87:
Commit added to let the oauthclient controller retrieve the ingress hostname from console config if the ingress capability is disabled.

alebedev87:
> > The console is still an important part of HyperShift based offerings like ROSA. The HyperShift engineers prefer the console to be running even without the default ingress (router). The way to expose it is a subject to a discussion, one of the ways to expose the console is described in [the doc I added today](https://github.com/openshift/console-operator/blob/ca595bff46855c31fc4c6204f13c726d9eaa5fbd/docs/alb-ingress-rosa-hcp.md).
> 
> But I don't think this will work since the console will stomp those changes, unless you put it in unmanaged state.

@spadgett: In this PR I modified the console operator to _not_ stomp on the console config and oauthclient if the ingress capability is disabled on external control plane topology (HyperShift). This allows the end user can provide it's own console base address (ALB is an example). I designed the way to provide the console base address too - via the same console config. WDYT?

spadgett:
> In this PR I modified the console operator to not stomp on the console config and oauthclient if the ingress capability is disabled on external control plane topology (HyperShift).

Are you referring to the internal console config ConfigMap managed by the operator or the public console operator config resource? Should we introduce the property to the console operator config itself? Then we could have proper API for specifying the public URLs when ingress is disabled.

Similarly, for auth, can we update the well-known endpoints with the right URLs if possible? Then console will discover the right values, along with anything else that reads the well-known document. With your current approach, I assume other components on the cluster won't be able to use auth? (For instance, anything that uses the OAuth proxy?)

How does this interact with external OIDC?

In general, it would be good to have public APIs for this configuration rather than asking customers to change internal config resources. I worry that if we change something in the internal console config in a future release, it won't get updated since console isn't stomping the changes, causing things to break.

Adding @stlaz for input.

spadgett:
Note that it is possible to manually edit these resources today without code changes if you put the console in `Unmanaged` state (but that will block cluster upgrades).

alebedev87:
> Are you referring to the internal console config ConfigMap managed by the operator or the public console operator config resource?

Internal config, `console-config` configmap.

> Should we introduce the property to the console operator config itself? Then we could have proper API for specifying the public URLs when ingress is disabled.

I chose the internal config because it already had the console base address field. Yes, the proper API may be a way to go. My only concern is time, this feature was promised in 4.14 and all the pieces are merged except for the console operator. 

> Similarly, for auth, can we update the well-known endpoints with the right URLs if possible?

It's not needed on HyperShift managed clusters. The auth server is managed centrally by HyperShift and it's not exposed via a route on the user cluster.

> Then console will discover the right values, along with anything else that reads the well-known document.

On HyperShift managed cluster the console already discovers the right value for the auth server.

```
$ oc -n openshift-console rsh deploy/console curl -k https://openshift.default.svc/.well-known/oauth-authorization-server
{
"issuer": "https://oauth.mytestcluster.5199.s3.devshift.org:443",
"authorization_endpoint": "https://oauth.mytestcluster.5199.s3.devshift.org:443/oauth/authorize",
"token_endpoint": "https://oauth.mytestcluster.5199.s3.devshift.org:443/oauth/token",
```

> How does this interact with external OIDC? Will this and external OIDC ever be configured together?

I don't know. I miss the knowledge here

> I worry that if we change something in the internal console config in a future release, it won't get updated since console isn't stomping the changes, causing things to break.

The way I implemented it only the `consoleBaseAddress` field is not stomped.

alebedev87:
@spadgett: talking about the public API, I can think about something like this: https://github.com/openshift/api/pull/1887.

spadgett:
> On HyperShift managed cluster the console already discovers the right value for the auth server.

OK, is the change that's needed then to set the right OAuth callback URL in the OAuthClient?

> The way I implemented it only the `consoleBaseAddress` field is not stomped.

This makes me feel better, but I still think a proper API would be better long term.

alebedev87:
>> On HyperShift managed cluster the console already discovers the right value for the auth server.

> OK, is the change that's needed then to set the right OAuth callback URL in the OAuthClient

In this PR I set the callback URL in console OAuthClient yes. But it's set with the console base address, not the auth server. All auth URL related resources are managed elsewhere (supposedly on HyperShift management side).

alebedev87:
The PR updated to use [the new field from the operator config](https://github.com/openshift/api/pull/1887).

alebedev87:
/retest

Seems like there are more test failures after the listers came back to `master`. Retrying to see if it's not transient.

pkg/console/controllers/clidownloads/controller.go
alebedev87: [The new API](https://github.com/openshift/api/pull/1887) has been implemented to setup alternative ingress for the downloads.

pkg/console/controllers/oauthclients/oauthclients.go
alebedev87: [The new API](https://github.com/openshift/api/pull/1887) has been implemented to support an alternative ingress.

pkg/console/subresource/configmap/configmap.go
```diff
@@ -47,12 +47,13 @@ func DefaultConfigMap(
 	nodeOperatingSystems []string,
 	copiedCSVsDisabled bool,
 	telemeterConfig map[string]string,
+	consoleHost string,
```
jhadvig, line 50: I see the point of adding this additional param, but I would rather avoid it, since for the non ALB setup, `consoleHost` will be the same as the host int the `activeConsoleRoute.spec`.
Lets rather create a new struct, which would include both of these types.

jhadvig (commented):
Thanks a lot @alebedev87 for the PR. Looks good.
Adding couple of comments. PTAL

docs/alb-ingress-rosa-hcp.md
```diff
@@ -0,0 +1,155 @@
...
+"issuer": "https://oauth.mytestcluster.5199.s3.devshift.org:443",
+"authorization_endpoint": "https://oauth.mytestcluster.5199.s3.devshift.org:443/oauth/authorize",
+"token_endpoint": "https://oauth.mytestcluster.5199.s3.devshift.org:443/oauth/token",
+```
+
+2. When the ingress capability is disabled, the console operator relies on the end user to provide the console and download URLs (using the operator API) for health checks and oauthclient.
+
+3. When the ingress capability is disabled, the console operator skips the implementation of the component route customization.
+
+4. To simulate a disabled ingress set the desired replicas to zero in the default ingress controller:
```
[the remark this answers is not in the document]
alebedev87: No, the routes will still be admitted. I rephrased it to:

```
To simulate the absence of ingress connectivity when the ingress capability is disabled, set the desired replicas to zero in the default ingress controller
```

pkg/console/subresource/configmap/configmap.go
alebedev87: > since for the non ALB setup, consoleHost will be the same as the host int the activeConsoleRoute.spec

Well, most of the time it's true. The `route.spec.host` is assigned automatically by the apiserver from the default ingress domain without the knowledge about the inrgesscontrollers which is only present in the route's status. That's why in `operator/sync_v400.go` I pass down the url returned by `routesub.GetActiveRouteInfo()` which returns only the admitted host.

docs/alb-ingress-rosa-hcp.md
```diff
@@ -0,0 +1,155 @@
...
+  selector:
+    app: console
+    component: downloads
+  type: NodePort
+EOF
+```
+
+### Create Ingress resources for the services
+
+To provision ALBs create the following resources:
```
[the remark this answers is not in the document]
alebedev87: > Wondering if this could be also done automatically by the console-operator if ALB setup is detected?

So far, we didn't see much automation around the certificate creation from the ALBO users. Probably because this requires quite some some effort to manage the cloud credentials. We can think about it as a followup enhancement. .

alebedev87:
Always apply nodeport services, to avoid the chicken and egg problem with alternative ingress: `Ingress` needs a nodeport service before it can provision a LB whose host we need to configure `console.operator.spec.ingress`.

alebedev87:
`ConsoleURL` removed from configmap config as it's not a customization but API.

alebedev87:
As discussed with @jhadvig: create `NodePort` service only for clusters with disabled ingress capability.

alebedev87:
/retest

CI outage.

alebedev87:
`e2e-aws-console` test is unstable and needs to be overridden [like we did for #907](https://github.com/openshift/console-operator/pull/907#issuecomment-2144718594).

jhadvig:
Thanks @alebedev87 for the PR. LGTM
/lgtm
/retest

jhadvig:
Overriding the TC due to flaking e2e test

PR #888, opened 2024-04-15
[WIP]CONSOLE-3340: Rewriting the downloads server into golang

Mylanos:
Adding the sidecar container to the console pod, that shares oc executables from cli-artifacts image to console container. The console will serve this executables in a goroutine on port 8080 instead of the standalone Downloads server. Rerouting the downloads service to point itself to console container.

### DIFF
--- a/bindata/assets/deployments/console-deployment.yaml
+++ b/bindata/assets/deployments/console-deployment.yaml
@@ -22,6 +22,9 @@ spec:
       annotations:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
         openshift.io/required-scc: restricted-v2
+    volumes:
+      - name: shared-data
+        emptyDir: {}
     spec:
       nodeSelector:
         node-role.kubernetes.io/master: ""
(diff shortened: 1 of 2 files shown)


bindata/assets/deployments/console-deployment.yaml
```diff
@@ -56,6 +75,7 @@ spec:
             - "--public-dir=/opt/bridge/static"
             - "--config=/var/console-config/console-config.yaml"
             - "--service-ca-file=/var/service-ca/service-ca.crt"
+            - "--downloads-server"
```
jhadvig, line 78: This flags is to determine if goroute, which is responsible for running the handler, should run ?

bindata/assets/deployments/console-deployment.yaml
```diff
@@ -21,6 +21,9 @@ spec:
         component: ui
       annotations:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
+    volumes:
+      - name: shared-data
+        emptyDir: {}
```
jhadvig, line 26: why we need the `emptyDir` ?

PR #900, opened 2024-05-10
OCPBUGS-33505: Add missing return statement to fix crash in healthcheck controller

TheRealJon:


jhadvig (approved):
Thanks @TheRealJon 👍 
/lgtm

PR #902, opened 2024-05-15
OCPBUGS-31777: Updates message verbs to use %q where appropriate.

theobarberbany:
This change updates the message verbs to use %q instead of %s, where appropriate, to properly escape output.

### DIFF
--- a/pkg/cmd/crdconversionwebhook/converter/framework.go
+++ b/pkg/cmd/crdconversionwebhook/converter/framework.go
@@ -78,13 +78,13 @@ func serve(w http.ResponseWriter, r *http.Request, convert convertFunc) {
 	contentType := r.Header.Get("Content-Type")
 	serializer := getInputSerializer(contentType)
 	if serializer == nil {
-		msg := fmt.Sprintf("invalid Content-Type header `%s`", contentType)
+		msg := fmt.Sprintf("invalid Content-Type header %q", contentType)
 		klog.Errorf(msg)
 		http.Error(w, msg, http.StatusBadRequest)
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/cmd/crdconversionwebhook/converter/framework.go
```diff
@@ -78,13 +78,13 @@ func serve(w http.ResponseWriter, r *http.Request, convert convertFunc) {
 	contentType := r.Header.Get("Content-Type")
 	serializer := getInputSerializer(contentType)
 	if serializer == nil {
-		msg := fmt.Sprintf("invalid Content-Type header `%s`", contentType)
+		msg := fmt.Sprintf("invalid Content-Type header `%q`", contentType)
```
elmiko, line 81: i thought `%q` added quotes around the string, do we also want the  single quotes  as well ?

eg something like this
```suggestion
		msg := fmt.Sprintf("invalid Content-Type header %q", contentType)
```

pkg/cmd/crdconversionwebhook/converter/framework.go
```diff
@@ -117,7 +117,7 @@ func serve(w http.ResponseWriter, r *http.Request, convert convertFunc) {
 	accept := r.Header.Get("Accept")
 	outSerializer := getOutputSerializer(accept)
 	if outSerializer == nil {
-		msg := fmt.Sprintf("invalid accept header `%s`", accept)
+		msg := fmt.Sprintf("invalid accept header `%q`", accept)
```
elmiko, line 120: same question here

elmiko (commented):
just a really nitty question ;)

elmiko (commented):
/lgtm

PR #904, opened 2024-05-20
OCPBUGS-33715: Cache organization ID

jhadvig:
Fetching ORGANIZATION_ID is too noisy on the OCM side. Lets rather cache it on main operator controller instance and use when building the console-config CM. 
The correct order for setting the ID should be:
1. custom ORGANIZATION_ID is awailable as telemetry annotation on console-operator config or in telemetry-config configmap
2. cached ORGANIZATION_ID is available on the operator controller instance

**else** fetch the ORGANIZATION_ID from OCM

/assign @TheRealJon

### DIFF
--- a/pkg/console/operator/operator.go
+++ b/pkg/console/operator/operator.go
@@ -91,10 +91,16 @@ type consoleOperator struct {
 
 	resourceSyncer resourcesynccontroller.ResourceSyncer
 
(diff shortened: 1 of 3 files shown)


TheRealJon:
/retest

Cluster install failed

TheRealJon (approved):
/lgtm

PR #916, opened 2024-07-03
 [release-4.16] OCPBUGS-34012: Cache organization ID

jhadvig:
Manual backport of https://github.com/openshift/console-operator/pull/905 since [additional commit](https://github.com/openshift/console-operator/pull/913) needed to be cherry-picked.

/assign @TheRealJon


jhadvig:
re-applying the label cause added commit which adds logging

PR #921, opened 2024-08-01
CONSOLE-4163: Add RBAC for console to GET PackageManifests

jhadvig:
/assign @TheRealJon


jhadvig:
overriding the ovn test since its flaking  and adding approval tags manually.

PR #925, opened 2024-08-19
OCPBUGS-38549: Get externalOrganizationID instead of the organizationID

jhadvig:
Tested the change locally and can confirm we are getting the external ID now.
/assign @TheRealJon


TheRealJon (approved):
/lgtm

PR #933, opened 2024-09-27
OCPBUGS-42534: Console-operator's route healthcheck should have longer retry

jhadvig:
Bumping the retry duration for the healthcheck to 5 seconds, but its up for a discussion whats the proper time.
Also adding additional logging around the healthcheck.

/assign @dgoodwin

### DIFF
--- a/pkg/console/controllers/healthcheck/controller.go
+++ b/pkg/console/controllers/healthcheck/controller.go
@@ -11,6 +11,7 @@ import (
 
 	// k8s
 
+	"k8s.io/apimachinery/pkg/util/wait"
 	coreinformersv1 "k8s.io/client-go/informers/core/v1"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/util/retry"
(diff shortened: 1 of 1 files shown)


pkg/console/controllers/healthcheck/controller.go
```diff
@@ -149,8 +154,14 @@ func (c *HealthCheckController) Sync(ctx context.Context, controllerContext fact
 
 func (c *HealthCheckController) CheckRouteHealth(ctx context.Context, operatorConfig *operatorsv1.Console, route *routev1.Route) (string, error) {
 	var reason string
+	healthCheckBackoff := wait.Backoff{
+		Steps:    5,
+		Duration: 5 * time.Second,
+		Factor:   1.0,
+		Jitter:   0.1,
+	}
```
dgoodwin, line 162: Not certain but I think this is approximately 25 seconds as written, 5 steps delayed by 5s each time plus some jitter/backoff.

Maybe we should start with 10 seconds to see if we can get this gone, so steps 10, duration 1 second?

pkg/console/controllers/healthcheck/controller.go
```diff
@@ -161,7 +172,7 @@ func (c *HealthCheckController) CheckRouteHealth(ctx context.Context, operatorCo
 				url, _, err = routeapihelpers.IngressURI(route, route.Spec.Host)
 				if err != nil {
 					reason = "RouteNotAdmitted"
-					return fmt.Errorf("console route is not admitted")
+					return fmt.Errorf("%s route is not admitted", route.Name)
```
dgoodwin, line 175: Could you add the logging in here before each return? I believe doing it above will only log once after the function returns, and not during the retry loop. Doing so here will show each error we hit and the timestamp, and give us something to search for to see if the problem is still happening if this retry change fixes it and the test start passsing. (which there's a very good chance of)

dgoodwin:
/lgtm

Thanks for this, hopefully this moves the console to green in 4.17 and 4.18 in a couple days. We'll keep monitoring.

jhadvig:
-retest

PR #907, opened 2024-05-27
OCPBUGS-33787: deps: bump openshift/api

alebedev87:
- Remove all code associated with discontinued v1alpha1 API (webhook)
- Remove webhook manifests
- Update client-go to eliminate v1alpha1 dependencies
- Update dependencies: k8s.io/component-base, k8s.io/apiserver
- Use `ConditionApplyConfiguration` for the auth status
- Vendor generated CRDs

Related PR: https://github.com/openshift/console-operator/pull/886.

PR #940, opened 2024-11-14
OCPBUGS-44556: console/status: set initial value of Message field

liouk:
Due to [this change](https://github.com/openshift/console-operator/pull/907/files#diff-8c5a69d560f0051604d71443b7331e22477075face20b6ea723b0fc71e1437b0L153-R163), the condition's `Message` field has now a zero value of `nil` (due to using the `ConditionApplyConfiguration`), while before it was simply `""`, which is a valid value for `Message`.

This PR simply initializes the `Message` field to an empty string.

### DIFF
--- a/pkg/console/status/auth_status.go
+++ b/pkg/console/status/auth_status.go
@@ -155,12 +155,14 @@ func existingOrNewCondition(applyConfig *configv1ac.AuthenticationApplyConfigura
 			tt     = metav1.Now()
 			reason = "Unknown"
 			status = metav1.ConditionUnknown
+			msg    = ""
 		)
 		condition = &applymetav1.ConditionApplyConfiguration{
 			Type:               &conditionType,
 			Status:             &status,
 			LastTransitionTime: &tt,
 			Reason:             &reason,
+			Message:            &msg,
 		}
 	}
 


spadgett (approved):
/lgtm

xingxingxia:
/jira backport release-4.18
For 4.17 and 4.16, I guess HCP has same issue, will confirm and backport later.

xingxingxia:
> For 4.17 and 4.16, I guess HCP has same issue, will confirm and backport later.

Tested 4.16.0-0.nightly-2024-12-02-183536 HCP BYO, it doesn't have the issue related to this PR. But it has the `"OIDC provider CA version not up to date in current deployment" and "status: Unknown"` issue. Create a 4.18/4.19 bug https://issues.redhat.com/browse/OCPBUGS-45300 and added 4.16 under its `Version-Release number of selected component` field.

Tested 4.17.0-0.nightly-2024-12-02-175016 HCP BYO, it has the issue of this PR:
```
 oc get co console
NAME      VERSION                              AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
console   4.17.0-0.nightly-2024-12-02-175016   True        False         True       170m    AuthStatusHandlerDegraded: Authentication.config.openshift.io "cluster" is invalid: [oidcClients[1].conditions[2].message: Required value, <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation]
```
So cherry-picking to 4.17 too (but I'll modify its bug description for HCP BYO):
/cherry-pick release-4.17

PR #954, opened 2024-12-18
ODC-7721: Disable developer perspective by default

Lucifergene:
Jira: https://issues.redhat.com/browse/ODC-7721

This pull request introduces changes to manage the visibility of different perspectives in the console server configuration. The changes include adding a new type for perspective IDs, updating the configuration builder to handle perspective visibility, and modifying the operator configuration manifest.

Changes to perspective management:

* [`pkg/console/subresource/consoleserver/types.go`](diffhunk://#diff-800af91abd4a802bcb0ff987850f14f9c4c8b7dd2ffdbb9ca601080deab21471R205-R214): Added `PerspectiveID` type and constants for admin and developer perspectives.

Configuration updates:

* [`pkg/console/subresource/consoleserver/config_builder.go`](diffhunk://#diff-4c176497477ce0b7c52110bc62b4d7cbe480530f729b644af85f38c053eb753eR503-R512): Modified the `customization` function to include perspective visibility settings and disable the developer perspective by default.
* [`manifests/01-operator-config.yaml`](diffhunk://#diff-2c1e82f8842fcdb54a3ccd5f686154e741b1c146316fb814b449277536c0ca98R22-R25): Added configuration for perspectives, setting the developer perspective to be disabled by default.

### DIFF
--- a/manifests/01-operator-config.yaml
+++ b/manifests/01-operator-config.yaml
@@ -13,13 +13,15 @@ spec:
   managementState: Managed
   customization:
     capabilities:
-    - name: LightspeedButton
-      visibility:
-        state: Enabled
-    - name: GettingStartedBanner
-      visibility:
-        state: Enabled
+      - name: LightspeedButton
+        visibility:
+          state: Enabled
+      - name: GettingStartedBanner
+        visibility:
+          state: Enabled
+    # The admin and the dev perspectives are merged into a single perspective and the dev perspective is disabled.
+    # Jira Epic: https://issues.redhat.com/browse/ODC-7716
     perspectives:
-    - id: dev
-      visibility:
-        state: Disabled
+      - id: dev
+        visibility:
+          state: Disabled
--- a/manifests/01-operator-config.yaml
+++ b/manifests/01-operator-config.yaml
@@ -19,3 +19,7 @@ spec:
     - name: GettingStartedBanner
       visibility:
         state: Enabled
+    perspectives:
+    - id: dev
+      visibility:
+        state: Disabled
--- a/pkg/console/subresource/consoleserver/config_builder.go
+++ b/pkg/console/subresource/consoleserver/config_builder.go
@@ -501,6 +501,16 @@ func (b *ConsoleServerCLIConfigBuilder) customization() Customization {
 		}
 
 		conf.Perspectives = perspectives
+	} else {
+		// Disable the developer perspective by default
+		perspectives := make([]Perspective, 1)
+		perspectives[0] = Perspective{
+			ID: string(PerspectiveIDDev),
+			Visibility: PerspectiveVisibility{
+				State: PerspectiveDisabled,
+			},
+		}
+		conf.Perspectives = perspectives
 	}
 
 	conf.Capabilities = b.capabilities
--- a/pkg/console/subresource/consoleserver/types.go
+++ b/pkg/console/subresource/consoleserver/types.go
@@ -202,6 +202,16 @@ const (
 	PerspectiveAccessReview PerspectiveState = "AccessReview"
 )
 
+// PerspectiveID defines the id of the perspective.
+// "admin" is the id of the admin perspective.
+// "dev" is the id of the developer perspective.
+type PerspectiveID string
+
+const (
+	PerspectiveIDAdmin PerspectiveID = "admin"
+	PerspectiveIDDev   PerspectiveID = "dev"
+)
+
 // ResourceAttributesAccessReview defines the visibility of the perspective depending on the access review checks.
 // `required` and  `missing` can work together esp. in the case where the cluster admin
 // wants to show another perspective to users without specific permissions. Out of `required` and `missing` atleast one property should be non-empty.
--- a/pkg/console/subresource/configmap/configmap_test.go
+++ b/pkg/console/subresource/configmap/configmap_test.go
@@ -127,6 +127,10 @@ session: {}
 customization:
   branding: ` + DEFAULT_BRAND + `
   documentationBaseURL: ` + DEFAULT_DOC_URL + `
+  perspectives:
+    - id: dev
+      visibility:
+        state: Disabled
 servingInfo:
   bindAddress: https://[::]:8443
   certFile: /var/serving-cert/tls.crt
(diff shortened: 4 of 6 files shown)


manifests/01-operator-config.yaml
```diff
@@ -19,3 +19,7 @@ spec:
     - name: GettingStartedBanner
       visibility:
         state: Enabled
+    perspectives:
+    - id: dev
+      visibility:
+        state: Disabled
```
spadgett, line 25: Will this only apply to new installs? Since this is a create-only resource, clusters upgraded from previous versions won't have the change, correct?

spadgett:
I suspect we're need to make a lot of Cypress test updates before we can merge this since many tests switch to the dev perspective.

manifests/01-operator-config.yaml
spadgett: I think we need some logic in the operator to default the value to false for `dev` if there is no existing perspective config.

We'll need to be mindful of the edge case where the `admin` perspective is disabled and there is no config for the `dev` perspective. In that case, we shouldn't also disable `dev` by default. We don't want to disable both.

vikram-raj: Do we need to do anything for existing users who upgrade the cluster to 4.19 if no config is present? As we are not sure if they are using Developer perspective.

vikram-raj:
> I suspect we're need to make a lot of Cypress test updates before we can merge this since many tests switch to the dev perspective.

Yes, we need to update ODC cypress tests otherwise it will break the CI. Here is the story for it [ODC-7718](https://issues.redhat.com/browse/ODC-7718) we will get this in first.

manifests/01-operator-config.yaml
spadgett: @vikram-raj Yes. Clusters upgraded to 4.19 should have the dev perspective disabled by default unless there is specific perspective config. Admins who want the perspective can always reenable it through config.

pkg/console/subresource/consoleserver/config_builder.go
```diff
@@ -500,6 +500,16 @@ func (b *ConsoleServerCLIConfigBuilder) customization() Customization {
...
+		conf.Perspectives = perspectives
+	} else {
+		// Disable the developer perspective by default
+		perspectives := make([]Perspective, 1)
+		perspectives[0] = Perspective{
+			ID: string(PerspectiveIDDev),
+			Visibility: PerspectiveVisibility{
+				State: PerspectiveDisabled,
+			},
+		}
```
Lucifergene, line 512: >I think we need some logic in the operator to default the value to false for dev if there is no existing perspective config.

Thanks @spadgett 
This should handle the defaulting.

jhadvig:
Thanks @Lucifergene 
It looks like you need to also update the e2e tests. PTAL

Lucifergene:
**Note: No need to run retests on the failing ones.**
The ongoing `ci/prow/e2e-aws-console` failure is caused due to the merging of perspective.

pkg/console/subresource/consoleserver/config_builder.go
```diff
@@ -500,6 +500,16 @@ func (b *ConsoleServerCLIConfigBuilder) customization() Customization {
 			}
 		}
 
+		conf.Perspectives = perspectives
```
spadgett, line 503: Let's add a unit test that confirms that the dev perspective is not disabled when there is any perspective config.

spadgett: I wonder if this is a little odd since if I add a config to customize perspective `foo` it could also enable the `dev` perspective when it was disabled before because there was no config. We might want to change the logic here to only enable the `dev` perspective if the `admin` perspective is disabled (or disabled for at least some users due to an access review check).

What do you guys think?

spadgett (commented):
@Lucifergene Any thoughts on how to approach updating the e2e tests in the console repo? Maybe a good first step is to update the tests to use the admin perspective for any nav items that are now exposed there.

vikram-raj:
Hi @spadgett, regarding the e2e tests, @sanketpathak is working on updating the e2e tests. Here is the Jira ticket [ODC-7718](https://issues.redhat.com/browse/ODC-7718) and PR https://github.com/openshift/console/pull/14684

manifests/01-operator-config.yaml
```diff
@@ -19,3 +19,7 @@ spec:
     - name: GettingStartedBanner
       visibility:
         state: Enabled
+    perspectives:
```
jhadvig, line 22: Lets add comment why this perspective is disabled. Maybe even with link to the epic.

jhadvig (approved):
/lgtm
/approve

jhadvig:
QE Approver:
/assign @yapei

pkg/console/subresource/consoleserver/types.go
```diff
@@ -202,6 +202,16 @@ const (
...
 
+// PerspectiveID defines the id of the perspective.
+// "admin" is the id of the admin perspective.
+// "dev" is the id of the developer perspective.
+type PerspectiveID string
+
+const (
+	PerspectiveIDAdmin PerspectiveID = "admin"
+	PerspectiveIDDev   PerspectiveID = "dev"
+)
```
yapei, line 213: Shall we limit the valid values for `.spec.customization.perspectives`? 
```
$ oc patch console.operator cluster -p '{"spec":{"customization":{"perspectives":[{"id":"dev","visibility":{"state":"Disabled"}},{"id":"admin","visibility":{"state":"Disabled"}},{"id":"test","visibility":{"state":"Disabled"}}]}}}' --type merge
console.operator.openshift.io/cluster patched
[yapei@yapei-mac frontend (support)]$ oc get console.operator cluster -o json | jq .spec.customization.perspectives
[
  {
    "id": "dev",
    "visibility": {
      "state": "Disabled"
    }
  },
  {
    "id": "admin",
    "visibility": {
      "state": "Disabled"
    }
  },
  {
    "id": "test",
    "visibility": {
      "state": "Disabled"
    }
  }
]
```

spadgett: @yapei Unfortunately we can't since we don't know the IDs of perspectives that can be contributed by plugins

yapei: thank you @spadgett

yapei:
tested the changes and user can toggle `Disabld/Enabled` state of dev perspective through `.spec.customization.perspectives` 
/label qe-approved

PR #960, opened 2025-02-14
OCPBUGS-48740: Vendor the missing CSP directives

jhadvig:
Vendoring:
- https://github.com/openshift/api/pull/2163
- https://github.com/openshift/api/pull/2176

/assign @TheRealJon


TheRealJon (approved):
Had trouble testing this locally, but it's passing CI tests, so I think it's ok. I probably just have an issue with local config.

/lgtm

PR #976, opened 2025-04-09
CONSOLE-4523: Add rhel8 and rhel9 oc binaries for Linux OS in CLI downloads

jhadvig:
Updating the downloads server in order for it to serve the rhel8 and rhel9 binaries for the Linux OS.
Also needed to update `CLIDownloadsSyncController` controller that is responsible for creating the ConsoleCLIDownloads CR of `oc` binaries.
Screen:
Screenshot 2025-04-09 at 13 20 54

Originally I wanted to put all the oc + oc.rhel8 + oc.rhel9 into a single archive, but that would be a huge one, more then 500MB 🤯  so I rather added an archive for each binary type.

/assign @Mylanos 

@opayne1 Im open for discussion about the name for the link, cause basically what I did is I've only appended ` - RHEL 8` or ` - RHEL 9` for RHEL binaries (`oc.rhel8` and `oc.rhel9`)

### DIFF
--- a/bindata/assets/deployments/downloads-deployment.yaml
+++ b/bindata/assets/deployments/downloads-deployment.yaml
@@ -125,19 +125,26 @@ spec:
 
               for arch, operating_system, path in [
                   ('amd64', 'linux', '/usr/share/openshift/linux_amd64/oc'),
+                  ('amd64', 'linux', '/usr/share/openshift/linux_amd64/oc.rhel8'),
+                  ('amd64', 'linux', '/usr/share/openshift/linux_amd64/oc.rhel9'),
                   ('amd64', 'mac', '/usr/share/openshift/mac/oc'),
                   ('amd64', 'windows', '/usr/share/openshift/windows/oc.exe'),
                   ('arm64', 'linux', '/usr/share/openshift/linux_arm64/oc'),
+                  ('arm64', 'linux', '/usr/share/openshift/linux_arm64/oc.rhel8'),
+                  ('arm64', 'linux', '/usr/share/openshift/linux_arm64/oc.rhel9'),
                   ('arm64', 'mac', '/usr/share/openshift/mac_arm64/oc'),
                   ('ppc64le', 'linux', '/usr/share/openshift/linux_ppc64le/oc'),
+                  ('ppc64le', 'linux', '/usr/share/openshift/linux_ppc64le/oc.rhel8'),
+                  ('ppc64le', 'linux', '/usr/share/openshift/linux_ppc64le/oc.rhel9'),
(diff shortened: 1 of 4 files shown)


pkg/console/controllers/clidownloads/controller_test.go
```diff
@@ -124,6 +124,14 @@ The oc binary offers the same capabilities as the kubectl binary, but it is furt
 							Href: "https://www.example.com/amd64/linux/oc.tar",
 							Text: "Download oc for Linux for x86_64",
 						},
+						{
+							Href: "https://www.example.com/amd64/linux/oc.rhel8.tar",
```
opayne1, line 128: Hey @jhadvig 😄 Since they all say "Download oc for ..." can we move that to the sentence that comes before the bullets to say something like 
"The oc binary offers the same capabilities as the kubectl binary, but it is further extended to natively support OpenShift Container Platform features. You can download oc using the following links."
We can also specify that last sentence further if you prefer.

Mylanos (commented):
Seems like we are missing oc x86_64 binaries for Windows and MacOS after these changes. I'm getting 404's for both of those. 
This is list of binaries provided prior to PR
Screenshot 2025-04-09 at 19 52 32

List of binaries after PR:
Screenshot 2025-04-09 at 19 52 29

bindata/assets/deployments/downloads-deployment.yaml
```diff
@@ -123,30 +123,52 @@ spec:
...
+
+              BASE_PATH = '/usr/share/openshift'
+
+              for arch, operating_system, binaries in ARCHIVE_MAPPINGS:
+                  dst_dir = os.path.join(arch, operating_system)
+                  os.makedirs(dst_dir, exist_ok=True)
+
+                  for binary in binaries:
+                      # Construct source path from base + os_arch + binary
+                      os_suffix = f"{operating_system}_{arch}"
```
Mylanos, line 144: nit: we could move this `os_suffix` assignment into the scope of the outer cycle.

```suggestion
                  os_suffix = f"{operating_system}_{arch}"
                  for binary in binaries:
                      # Construct source path from base + os_arch + binary
```

pkg/console/controllers/clidownloads/controller_test.go
opayne1: Yeah that looks good to me to keep consistency! Thanks for looping me in!!

jhadvig:
QE Approver:
/assign @yapei

PX Approver:
/assign @sferich888

yapei:
tested the new oc binaries, no issues found
/label qe-approved

PR #406, opened 2020-04-02
Move to go mods, use openshift/build-machinery-go for building

stlaz:

PR #982, opened 2025-04-28
OCPBUGS-55425: README and Makefile improvements

Mylanos:


README.md
```diff
@@ -169,18 +169,16 @@ OS_DEBUG=true OS_BUILD_PLATFORMS=linux/amd64 make
 # note that you can build for mulitiple platforms with:
 make build-cross
 ```
-But the `make` step is included in the `Dockerfile`, so this does not need to be done manually.
```
Mylanos, line 172: Ever since https://github.com/openshift/console-operator/pull/406 we do not have `Dockerfile` in our repository.

README.md
```diff
@@ -169,18 +169,16 @@ OS_DEBUG=true OS_BUILD_PLATFORMS=linux/amd64 make
 # note that you can build for mulitiple platforms with:
 make build-cross
```
Mylanos, line 170: Also since this PR merged we are missing the `build` and `build-cross` rules in our Makefile, wondering if we want to add one or we want to remove this part of README.md and only reference builds via docker.

jhadvig (approved):
/lgtm
/approve

PR #992, opened 2025-06-11
OCPBUGS-57129: Add missing Console capability annotation

jhadvig:
/assign @wking 

@linoyaslan FYI

### DIFF
--- a/manifests/0000_90_console-operator_03_console_plugin_storage_migration.yaml
+++ b/manifests/0000_90_console-operator_03_console_plugin_storage_migration.yaml
@@ -7,6 +7,7 @@ metadata:
     include.release.openshift.io/ibm-cloud-managed: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
+    capability.openshift.io/name: Console
 spec:
   resource:
     group:  console.openshift.io
--- a/manifests/07-operator-ibm-cloud-managed.yaml
+++ b/manifests/07-operator-ibm-cloud-managed.yaml
@@ -3,6 +3,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   annotations:
+    capability.openshift.io/name: Console
     config.openshift.io/inject-proxy: console-operator
     include.release.openshift.io/hypershift: "true"
     include.release.openshift.io/ibm-cloud-managed: "true"
--- a/profile-patches/ibm-cloud-managed/07-operator.yaml-patch
+++ b/profile-patches/ibm-cloud-managed/07-operator.yaml-patch
@@ -1,6 +1,7 @@
 - op: replace
   path: /metadata/annotations
   value:
+    capability.openshift.io/name: Console
     config.openshift.io/inject-proxy: console-operator
     include.release.openshift.io/hypershift: "true"
     include.release.openshift.io/ibm-cloud-managed: "true"


manifests/0000_90_console-operator_03_console_plugin_storage_migration.yaml
```diff
@@ -7,6 +7,7 @@ metadata:
     include.release.openshift.io/ibm-cloud-managed: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
+    capability.openshift.io/name: Console
```
wking, line 10: You can add this annotation to `include.release.openshift.io/hypershift` cluster-profile manifests, because openshift/hypershift#6183 is still in flight trying to make "HyperShift clusters with `Console` disabled" an option.

However, this particular manifest also belongs to other cluster profiles, like the default `include.release.openshift.io/self-managed-high-availability`.  You cannot add the `Console` capability annotation to _those_, because there are already `self-managed-high-availability` clusters with the `Console` disabled and the  cluster-version operator actively managing this resource.  Upon update, the CVO will [implicitly enable][1] the `Console` capability, to avoid orphaning that resource.  There's [OTA-1039][2] thinking that some day, we might grow some kind of knob to work around that, but there hasn't been enough interest to get it prioritized.  So for today's CVO, you'd need to shard this manifest into two:

* One for cluster profiles like `include.release.openshift.io/hypershift` where you knew there couldn't be any `Console`-less clusters yet, where you could add this `Console` annotation.
* Another for cluster profiles like `include.release.openshift.io/self-managed-high-availability` where there might be existing `Console`-less clusters, where you need to either live with the presence of the resource on clusters without `Console` enabled, or find some other workaround.

Also in this space, the `include.release.openshift.io/single-node-developer` profile was abandoned, so you can drop that (e.g. see openshift/cluster-version-operator#685)

[1]: https://github.com/openshift/enhancements/blob/db1b45fbf81c1552dd0527ffa5b4c65d759c63f3/enhancements/installer/component-selection.md#updates
[2]: https://issues.redhat.com/browse/OTA-1039

wking, line 10: You can add this annotation to `include.release.openshift.io/hypershift` cluster-profile manifests, because openshift/hypershift#6183 is still in flight trying to make "HyperShift clusters with `Console` disabled" an option.

However, this particular manifest also belongs to other cluster profiles, like the default `include.release.openshift.io/self-managed-high-availability`.  You cannot add the `Console` capability annotation to _those_, because there are already `self-managed-high-availability` clusters with the `Console` disabled and the  cluster-version operator actively managing this resource.  Upon update, the CVO will [implicitly enable][1] the `Console` capability, to avoid orphaning that resource.  There's [OTA-1039][2] thinking that some day, we might grow some kind of knob to work around that, but there hasn't been enough interest to get it prioritized.  So for today's CVO, you'd need to shard this manifest into two:

* One for cluster profiles like `include.release.openshift.io/hypershift` where you knew there couldn't be any `Console`-less clusters yet, where you could add this `Console` annotation.
* Another for cluster profiles like `include.release.openshift.io/self-managed-high-availability` where there might be existing `Console`-less clusters, where you need to either live with the presence of the resource on clusters without `Console` enabled, or find some other workaround.

Also in this space, the `include.release.openshift.io/single-node-developer` profile was abandoned, so you can drop that (e.g. see openshift/cluster-version-operator#685)

[1]: https://github.com/openshift/enhancements/blob/db1b45fbf81c1552dd0527ffa5b4c65d759c63f3/enhancements/installer/component-selection.md#updates
[2]: https://issues.redhat.com/browse/OTA-1039

PR #1012, opened 2025-07-07
OCPBUGS-58434: Console-operator should report plugin services as relatedObjects

jhadvig:
Console-operator should be reporting plugin's services and its namespace as relatedObjects, which means that the whole namespace should be scrapped when generating must-gather for debug purposes.
Currently we are only gathering the plugins and not their services.

/assign @TheRealJon @Leo6Leo

### DIFF
--- a/pkg/console/starter/starter.go
+++ b/pkg/console/starter/starter.go
@@ -67,6 +67,7 @@ import (
 
 	telemetry "github.com/openshift/console-operator/pkg/console/telemetry"
 
+	"github.com/go-test/deep"
 	consoleoperator "github.com/openshift/console-operator/pkg/console/operator"
 	"github.com/openshift/library-go/pkg/operator/loglevel"
 )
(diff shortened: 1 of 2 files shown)


pkg/console/starter/starter.go
```diff
@@ -456,20 +456,29 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
...
 			return false, nil
 		}
 		for _, plugin := range consolePlugins.Items {
-			relatadPlugin := configv1.ObjectReference{
+			relatedObjects = append(relatedObjects, configv1.ObjectReference{
 				Group:    "console.openshift.io",
 				Resource: "consoleplugins",
 				Name:     plugin.GetName(),
+			})
+			if plugin.Spec.Backend.Service != nil {
```
spadgett, line 472: We might include any plugin proxy namespaces as well.

Should we de-dup the list if multiple plugins are served from the same namespace? (Or if the plugin namespace is the same as one of the hard-coded ones, openshift-console or openshift-console-operator?)

spadgett:
/hold
for review comments

pkg/console/starter/starter.go
jhadvig: Added
- namespaces from the ConsolePluginProxy 
- deduplication of the relatedObjects
- unitTest

Thank you @spadgett PTAL

PR #1023, opened 2025-07-23
OCPBUGS-36520: Rename Dockerfile to .ocp

Leo6Leo:
# Description
Rename Dockerfile from Dockerfile.rhel7 to Dockerfile.ocp


jhadvig (changes requested):
We need to duplicate the, in order to avoid CI failures, until we do followup changes in the release repo. After those are in, we can remove the original `Dockerfile.rhel7`

README.md
```diff
@@ -178,7 +178,7 @@ docker build -t <registry>/<your-username>/console-operator:<version> .
 # following: docker.io/openshift/origin-console-operator:latest
 # for development, you are going to push to an alternate registry.
 # specifically it can look something like this:
-docker build -f Dockerfile.rhel7 -t quay.io/benjaminapetersen/console-operator:latest .
+docker build -f Dockerfile.ocp -t quay.io/benjaminapetersen/console-operator:latest .
```
cajieh, line 181: @Leo6Leo Would it be clearer to replace the `benjaminapetersen` with a placeholder like `<your-username>`. Also, this appears in multiple places in the file.

Leo6Leo:
Update: Based on my research and understanding, this is my planning for the next move.

The Problem Chain:
  1. This console-operator PR: Changed Makefile to reference Dockerfile.ocp
  2. CI configuration: Still expects Dockerfile.rhel7
  3. Build failure: CI can't find expected Dockerfile which is Dockerfile.rhel7
  4. No tests run: Build fails before e2e tests start

So what I will be doing is:
- [ ] 1. Update this PR to revert the changes in the Makefile.
   - At this point, the CI should already start passing since the CI is looking for `Dockerfile.rhel7` instead of `Dockerfile.ocp`.
- [ ] 2. Merge this PR
- [ ] 3. Merge the PR in `openshift/release` to update the CI https://github.com/openshift/release/pull/67416
   - At this point after the PR above get merged, the CI on the new PRs should start failing since the CI will now be looking for `Dockerfile.ocp` instead of `Dockerfile.rhel7`
- [ ] 4. Open a new PR to modify the Makefile, and at that point the CI on the new PR should be passing.

PR #1041, opened 2025-09-05
fix(oidc): fix  OIDCClientSecretGet condition

devguyio:
Since the `Authentication`  resource references CAs and Secrets in the `openshift-config` namespace, its validation should focus on the `openshift-config` namespace.

However the `OIDCClientSecretGet` condition logic checks for the synced client secret, namely `console-oauth-config ` under `openshift-console`, instead of the one referenced by the `Authentication` resource under  `openshift-config`.

This PR fixes this behavior.


jhadvig (approved):
/lgtm
/approve

PR #1053, opened 2025-10-06
OCPBUGS-62307: Remove orphaned console-conversion-webhook deployment …

jhadvig:
…


pkg/console/starter/starter.go
```diff
@@ -555,6 +526,11 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		controllerContext.EventRecorder,
 	)
 
+	migrationCleanupController := migration.NewMigrationCleanupController(
```
TheRealJon: Do we need the full overhead of a controller here? If it’s a one-and-done action at startup, a simple init function might be cleaner than a permanent sync loop.

PR #1067, opened 2025-11-13
OCPBUGS-61432: fix(oidc): fix OIDC client secret lookup, validation, and condition cleanup

devguyio:
This PR addresses three issues related to OIDC authentication:

1. Fixed OIDC client secret lookup in oidcsetup controller to use the correct informer (configSecretsLister), namespace (openshift-config), and dynamic secret name from the Authentication CR, instead of hardcoded values.

2. Fixed secret revision validation to compare the TARGET secret (openshift-console/console-oauth-config) with the deployment annotation, following the same pattern as ConfigMap CA trust validation. This ensures proper verification of secret sync status.

3. Added condition cleanup in sync_v400 to properly clear the OIDCProviderTrustedAuthorityConfigGet degraded condition when authentication type changes from OIDC to non-OIDC (e.g., IntegratedOAuth). This prevents the Console Operator from remaining in a Degraded state indefinitely during rollback scenarios.

### DIFF
--- a/pkg/console/controllers/oidcsetup/oidcsetup.go
+++ b/pkg/console/controllers/oidcsetup/oidcsetup.go
@@ -58,6 +58,7 @@ type oidcSetupController struct {
 	authnLister               configv1listers.AuthenticationLister
 	consoleOperatorLister     operatorv1listers.ConsoleLister
 	configConfigMapLister     corev1listers.ConfigMapLister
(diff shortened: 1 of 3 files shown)


everettraven (commented):
Aside from one minor comment, this looks fine to me.

jhadvig (approved):
/lgtm
/approve

PR #1112, opened 2026-02-18
OCPBUGS-77026: DNS optimization, add trailing dots

jotak:
Update getProxyServiceURL and getServiceURL to generate URLs with trailing dots, such as
"my-service.my-namespace.svc.cluster.local.:3000", which makes it unambiguous to the DNS resolver, speeding up DNS resolution by avoiding unnecessary queries.


spadgett (approved):
/lgtm

JoelSpeed:
@spadgett @jhadvig this looks like it should be relatively straightforward as a bug fix, can you assist with the remaining labels/whether we need those and try to get this merged

PR #1144, opened 2026-04-28
OPRUN-4591: Add RBAC for console lifecycle-server API access

perdasilva:
## Summary

- Adds a `ClusterRole` (`system:openshift:console:lifecycle-reader`) and `ClusterRoleBinding` that grant the console ServiceAccount `GET` access to lifecycle-server nonResourceURL paths (`/api/*/lifecycles/*`).
- The lifecycle-server uses `TokenReview` / `SubjectAccessReview` for authn/authz on nonResourceURLs, so the console pod SA needs explicit RBAC to call the API.
- Gated behind `TechPreviewNoUpgrade` feature set and scoped to the `Console` capability.

## Details

The lifecycle-server exposes operator lifecycle data (e.g. update risks, recommended actions) via nonResourceURL endpoints. The console backend needs to query these endpoints to surface lifecycle information in the UI.

Since nonResourceURLs are not covered by standard namespace-scoped RBAC, a dedicated `ClusterRole` with `nonResourceURLs` rules is required. The `ClusterRoleBinding` binds this role to the `console` ServiceAccount in the `openshift-console` namespace.

## Test plan

- [ ] Verify the ClusterRole and ClusterRoleBinding are created on a TechPreviewNoUpgrade cluster
- [ ] Confirm the console SA can `GET /api/*/lifecycles/*` endpoints on the lifecycle-server
- [ ] Confirm the resources are **not** created on a standard (non-TechPreview) cluster

🤖 Commit and PR descriptions generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Added permissions for the OpenShift console to read lifecycle-related information from the platform API.

### DIFF
--- a/manifests/03-rbac-role-cluster-openshift-console-user-operator-lifecycle-reader.yaml
+++ b/manifests/03-rbac-role-cluster-openshift-console-user-operator-lifecycle-reader.yaml
@@ -1,0 +1,45 @@
+# RBAC required for the OpenShift Console to access the lifecycle-server API.
+#
+# The lifecycle-server authenticates callers via TokenReview and authorizes
+# them via SubjectAccessReview on nonResourceURLs. The console backend uses
+# its pod ServiceAccount token to authenticate, so that SA needs permission
+# to GET the lifecycle API paths.
+#
+# These resources should be managed by the console-operator. They are
+# provided here as a reference for development and testing.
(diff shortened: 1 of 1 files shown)


TheRealJon:
One question: Why do we need to use the console service account bearer token to make requests to the lifecycle service? Do we need to escalate privilege like this?

PR #1153, opened 2026-05-11
OCPBUGS-85379: Bump default documentation URL to OCP 4.23

Cragsmann:
**Analysis / Root cause**:

**Solution description**:

**Test setup:**

**Test cases:**

**Browser conformance**:

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (or Epiphany on Linux)

**Additional info:**

**Reviewers and assignees:**

## Summary by CodeRabbit

* **Chores**
  * Updated default documentation links to reference the latest OpenShift Container Platform version.

### DIFF
--- a/pkg/console/subresource/configmap/brand_ocp.go
+++ b/pkg/console/subresource/configmap/brand_ocp.go
@@ -5,5 +5,5 @@ package configmap
 
 const (
 	DEFAULT_BRAND   = "ocp"
-	DEFAULT_DOC_URL = "https://access.redhat.com/documentation/en-us/openshift_container_platform/4.21/"
+	DEFAULT_DOC_URL = "https://access.redhat.com/documentation/en-us/openshift_container_platform/4.23/"
 )


jhadvig:
@jseseCCS do we know if the next version of docs URL will contain 4.23 or 5.0

jseseCCS:
> @jseseCCS do we know if the next version of docs URL will contain 4.23 or 5.0

we don't know yet, no. i pinged some teammates to make sure and if i hear differently i'll let you know.

jseseCCS:
@jhadvig Hey Jakub! 

Asking the obvious, I know, but since 4.23 docs aren't published yet, should we point this to 4.22 for now?

Also, given the uncertainty you mentioned about whether the next version will be 4.23, 5.0, or both, do we have enough clarity to commit to 4.23 here? as i said above, I asked around and will know more soon.

On the hardcoded URL: We handle version numbers on the docs side through attributes and variables. I'll look into how consistently we're applying that for cases like this. Is there an equivalent mechanism on the code side that we might be able to leverage here?

jhadvig:
@jseseCCS fair enough, let's keep this PR open until we know which version we should go with. 
@Cragsmann

jhadvig:
Closing in favour  of https://github.com/openshift/console-operator/pull/1207

PR #1164, opened 2026-05-27
OCPBUGS-38676: Add retry for transient API errors to prevent Degraded blips

sg00dwin:
## Summary

The console ClusterOperator intermittently flips `Degraded=True` during upgrade and CI runs when transient API errors (conflicts, timeouts, connection refused) hit `resourceapply.Apply*()` calls. The error flows directly to the status handler with no retry, setting `Degraded=True` for ~1 minute until the next sync resolves it. This causes CI test failures (8.7% failure rate in OCP 5.0).

This PR adds a shared retry utility and wraps all API write operations that can trigger Degraded conditions, absorbing transient errors before they reach status reporting.

## Changes

- **New**: `pkg/console/controllers/util/retry.go` — shared `RetryOnTransientError()` with backoff (3 steps, 500ms base, 2x factor, ~3.5s max). Retries all errors except permanent ones (Forbidden, Invalid, AlreadyExists, etc.).
- **New**: `pkg/console/controllers/util/retry_test.go` — 10 unit tests covering error classification and retry behavior.
- **New**: `pkg/console/operator/retry.go` — thin wrapper for the operator package.
- **Modified**: Wrapped 16 API write call sites across 10 files:
  - `sync_v400.go` — ConfigMap, ServiceCA, TrustedCA, Deployment, ConsoleConfig, PublicConfig, SessionSecret
  - Individual controllers — PDB, Service, CLIDownloads, DownloadsDeployment, ServiceAccounts, OAuthClients, OAuthClientSecret, UpgradeNotification, StorageVersionMigration
- **Not modified**: Healthcheck controller (already has its own retry from OCPBUGS-24041)

## Testing

- All existing unit tests pass
- New retry utility tests cover: retryable vs permanent error classification, retry-then-succeed, retry exhaustion, generic network error retry
- `make test-unit` passes clean
- `gofmt` / `go vet` clean

## Release Notes

* **Bug Fixes**
  * Enhanced reconciliation reliability with automatic retry handling for transient API failures across multiple console controllers and operator components, improving resilience during temporary network or server issues.

* **Tests**
  * Added comprehensive unit tests for transient error handling and retry logic.


pkg/console/operator/sync_v400.go
```diff
@@ -261,14 +261,27 @@ func (co *consoleOperator) SyncConsoleConfig(ctx context.Context, consoleConfig
...
 		updated := consoleConfig.DeepCopy()
 		updated.Status.ConsoleURL = consoleURL
-		return co.consoleConfigClient.UpdateStatus(ctx, updated, metav1.UpdateOptions{})
+		var result *configv1.Console
+		err := retryOnTransientError(func() error {
+			var e error
+			result, e = co.consoleConfigClient.UpdateStatus(ctx, updated, metav1.UpdateOptions{})
+			return e
+		})
+		return result, err
```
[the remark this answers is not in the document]
sg00dwin: [Claude] Valid observation — if `UpdateStatus` hits a 409 Conflict, the retry will send the same stale `resourceVersion` and fail again. However, the operator is the sole writer of `consoleConfig.Status.ConsoleURL`, so conflicts are extremely rare here. The retry is targeting the errors we actually see in CI logs — timeouts, connection refused, and server unavailable during upgrades.

Re-fetching inside the closure would require a live API `Get` (the lister cache won't have updated in ~3.5s), which changes the read pattern from cache to API. The 60-second reconcile loop self-heals if retries exhaust.

The same reasoning applies to the `SyncServiceCAConfigMap` / `SyncTrustedCAConfigMap` Update paths — the `existing` object comes from the lister cache, so re-fetching from the lister inside the retry would return the same object. A live API `Get` would fix conflict retries but changes the read pattern significantly.

Regarding the Create wrappers: `AlreadyExists` is already in the non-retryable list in `IsRetryableError`, so the retry stops immediately — no wasted attempts. The Create path is only reached when the lister returned `NotFound`, so `AlreadyExists` would indicate a race between cache staleness and the API. Silently swallowing it and fetching the existing object adds complexity for an edge case that self-heals on the next reconcile (60 seconds). The original code before this PR had the same behavior.

pkg/console/operator/sync_v400.go
```diff
@@ -489,13 +514,17 @@ func (co *consoleOperator) SyncServiceCAConfigMap(ctx context.Context, operatorC
...
+			return e
+		})
+		if createErr == nil {
 			klog.V(4).Infoln("service-ca configmap created")
-			return actual, "", err
-		} else {
-			return actual, "FailedCreate", err
+			return actual, "", nil
 		}
+		return actual, "FailedCreate", createErr
```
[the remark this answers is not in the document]
sg00dwin: [Claude] Same reasoning as the reply above — the Create paths here are only reached when the lister returned `NotFound`. `AlreadyExists` is already classified as non-retryable in `IsRetryableError`, so the retry stops immediately on that error with no wasted attempts.

The phantom-success scenario (server-side Create succeeds but client gets a timeout, retry hits AlreadyExists) is a narrow race that self-heals on the next 60-second reconcile. Adding a Get-on-AlreadyExists fallback would add complexity for an edge case that resolves itself within one sync cycle. The original code before this PR had the same behavior — AlreadyExists on Create was always returned as a failure.

pkg/console/operator/sync_v400.go
```diff
@@ -263,14 +262,27 @@ func (co *consoleOperator) SyncConsoleConfig(ctx context.Context, consoleConfig
 		klog.V(4).Infof("updating console.config.openshift.io with url: %v", consoleURL)
 		updated := consoleConfig.DeepCopy()
```
theobarberbany: this needs to be in the closure, otherwise e.g on a 409 every retry will use the stale (conflicting data)

pkg/console/operator/sync_v400.go
```diff
@@ -518,25 +538,34 @@ func (co *consoleOperator) SyncServiceCAConfigMap(ctx context.Context, operatorC
 		return existing, "", nil
 	}
 
-	actual, err := co.configMapClient.ConfigMaps(required.Namespace).Update(ctx, existing, metav1.UpdateOptions{})
-	if err == nil {
+	var actual *corev1.ConfigMap
+	updateErr := retryOnTransientError(func() error {
+		var e error
+		actual, e = co.configMapClient.ConfigMaps(required.Namespace).Update(ctx, existing, metav1.UpdateOptions{})
```
theobarberbany: again, I think if existing is stale we will fail on 409s

pkg/console/operator/sync_v400.go
```diff
@@ -549,12 +578,17 @@ func (co *consoleOperator) SyncTrustedCAConfigMap(ctx context.Context, operatorC
 		return existing, "", nil
 	}
 
-	actual, err := co.configMapClient.ConfigMaps(required.Namespace).Update(ctx, existing, metav1.UpdateOptions{})
-	if err != nil {
-		return actual, "FailedUpdate", err
+	var actual *corev1.ConfigMap
+	updateErr := retryOnTransientError(func() error {
+		var e error
+		actual, e = co.configMapClient.ConfigMaps(required.Namespace).Update(ctx, existing, metav1.UpdateOptions{})
```
theobarberbany: update can also be stale on retry here

pkg/console/controllers/clidownloads/controller.go
```diff
@@ -249,10 +254,15 @@ func ApplyCLIDownloads(ctx context.Context, consoleClient consoleclientv1.Consol
...
 
 	existingCLIDownloadsCopy.Spec = requiredCLIDownloads.Spec
-	actualCLIDownloads, err := consoleClient.Update(ctx, existingCLIDownloadsCopy, metav1.UpdateOptions{})
-	if err != nil {
-		klog.V(4).Infof("error updating %s consoleclidownloads custom resource: %v", cliDownloadsName, err)
-		return nil, "FailedUpdate", err
+	var actualCLIDownloads *v1.ConsoleCLIDownload
+	updateErr := controllersutil.RetryOnTransientError(func() error {
+		var e error
+		actualCLIDownloads, e = consoleClient.Update(ctx, existingCLIDownloadsCopy, metav1.UpdateOptions{})
```
theobarberbany: If anything changes this underneath us we will 409, and the retry doesn't support another get :)

sg00dwin:
@theobarberbany 
  Updated all 4 sites to re-fetch the latest object via a live API Get inside the retry closure before calling
  Update/UpdateStatus. Each retry attempt now uses the current resourceVersion instead of the original stale copy.

  - SyncConsoleConfig: re-fetches consoleConfig before UpdateStatus
  - SyncServiceCAConfigMap: re-fetches configmap and re-applies metadata before Update
  - SyncTrustedCAConfigMap: same pattern
  - ApplyCLIDownloads: re-fetches CLI downloads resource, re-applies spec and metadata before Updat

pkg/console/operator/retry.go
TheRealJon, line 1: What's the purpose of this re-export?

sg00dwin: It was added as a convenience wrapper for the 9 call sites in sync_v400.go, shorter name, no added logic.  If preferred, I can remove it and call `util.RetryOnTransientError` directly instead.

pkg/console/controllers/storageversionmigration/controller.go
```diff
@@ -103,7 +104,9 @@ func (c *StorageVersionMigrationController) syncStorageVersionMigration(ctx cont
 	if !succeeded {
 		klog.V(4).Infof("StorageVersionMigration has not succeeded yet")
 		// Delete the StorageVersionMigration if it has not succeeded yet
-		if err := c.deleteStorageVersionMigration(ctx); err != nil {
+		if err := util.RetryOnTransientError(func() error {
+			return c.deleteStorageVersionMigration(ctx)
```
jhadvig, line 109: Nit: `deleteStorageVersionMigration` doesn't guard against `NotFound`. Since `NotFound` is classified as retryable in `IsRetryableError`, this will retry 3 times if the resource is already gone, then surface a `"FailedDeleteSVM"` error for something that's effectively a success.

Consider adding a `NotFound` guard in `deleteStorageVersionMigration`:

```go
err := c.dynamicClient.Resource(storageVersionMigrationGVR).Delete(ctx, storageVersionMigrationName, metav1.DeleteOptions{})
if apierrors.IsNotFound(err) {
    return nil
}
```

Pre-existing issue.. but still I would suggest to fix it

jhadvig (commented):
Minor nit on the StorageVersionMigration retry — see inline comment.

jhadvig (changes requested):
THanks @sg00dwin 👍 
adding couple of comments

pkg/console/controllers/util/retry.go
```diff
@@ -0,0 +1,51 @@
...
+	Duration: 500 * time.Millisecond,
+	Factor:   2.0,
+	Jitter:   0.1,
+}
+
+// IsRetryableError returns true for errors worth retrying — everything
+// except known permanent errors. This naturally handles both API status
+// errors (apierrors.StatusError) and network-level errors (connection
+// refused, EOF, TLS failures) without explicit net.Error detection.
+func IsRetryableError(err error) bool {
```
jhadvig, line 26: The test explicitly asserts nil is "retryable". While retry.OnError never calls the predicate on nil, this is a confusing semantic. Consider adding `if err == nil { return false }` at the top, or removing the nil test case

pkg/console/controllers/oauthclientsecret/oauthclientsecret.go
```diff
@@ -160,7 +161,10 @@ func (c *oauthClientSecretController) syncSecret(ctx context.Context, clientSecr
 
 	secret, err := c.targetNSSecretsLister.Secrets(api.TargetNamespace).Get("console-oauth-config")
 	if apierrors.IsNotFound(err) || secretsub.GetSecretString(secret) != clientSecret {
-		_, _, err = resourceapply.ApplySecret(ctx, c.secretsClient, recorder, secretsub.DefaultSecret(operatorConfig, clientSecret))
+		err = util.RetryOnTransientError(func() error {
+			_, _, e := resourceapply.ApplySecret(ctx, c.secretsClient, recorder, secretsub.DefaultSecret(operatorConfig, clientSecret))
```
jhadvig, line 165: Here we are passing the same pre-built `clientCopy` on every retry attempt. If the first attempt fails with a conflict, the second attempt will send the same stale `ResourceVersion` and fail again. This needs to re-GET inside the loop like the `CLIDownloads` and ConfigMap update paths do:
```go
  oauthErr := util.RetryOnTransientError(func() error {
      latest, e := c.oauthClient.OAuthClients().Get(ctx, clientCopy.Name, metav1.GetOptions{})
      if e != nil {
          return e
      }
      oauthsub.RegisterConsoleToOAuthClient(latest, consoleURL, secretsub.GetSecretString(sec))
      _, _, e = oauthsub.CustomApplyOAuth(c.oauthClient, latest, ctx)
      return e
  })
```

pkg/console/operator/retry.go
```diff
@@ -0,0 +1,10 @@
+package operator
+
+import (
+	// operator
+	"github.com/openshift/console-operator/pkg/console/controllers/util"
+)
+
+func retryOnTransientError(fn func() error) error {
```
jhadvig: I would remove this... since its just a online wrapper of `RetryOnTransientError`

pkg/console/controllers/upgradenotification/controller.go
```diff
@@ -131,14 +131,19 @@ func (c *UpgradeNotificationController) syncClusterUpgradeNotification(ctx conte
...
 			},
 		}
-		_, err = c.consoleNotificationClient.Create(ctx, notification, metav1.CreateOptions{})
-		if err != nil && !apierrors.IsAlreadyExists(err) {
-			return "FailedCreate", err
+		createErr := util.RetryOnTransientError(func() error {
+			_, e := c.consoleNotificationClient.Create(ctx, notification, metav1.CreateOptions{})
+			return e
+		})
+		if createErr != nil && !apierrors.IsAlreadyExists(createErr) {
```
jhadvig, line 138: `!apierrors.IsAlreadyExists(createErr)` is now partly redundant since `AlreadyExists` will bubble out  immediatel to the `RetryOnTransientError` which will capture it

jhadvig (changes requested):

pkg/console/controllers/oauthclientsecret/oauthclientsecret.go
sg00dwin: Claude's analysis of `CustomApplyOAuth` shows it does a fresh `Get()` internally on each invocation (oauthclient.go:24), and `EnsureObjectMeta` only merges labels/annotations/ownerrefs — not `resourceVersion` (object_merger.go:13-19). So the Update should carry the fresh `resourceVersion` from that internal Get, even when `clientCopy` is reused across retries. 

Adding an explicit Get inside the retry closure would result in two Gets per attempt. Let me know if I'm missing something.

pkg/console/controllers/upgradenotification/controller.go
sg00dwin: Could you clarify what you're suggesting here? Claude's analysis shows that `retry.OnError` returns the `AlreadyExists` error rather than swallowing it — so the outer `!apierrors.IsAlreadyExists(createErr)` check still seems needed to suppress it. Without it, `AlreadyExists` would propagate as `"FailedCreate"`. Am I misreading your comment, or were you suggesting handling it differently (e.g., returning nil for `AlreadyExists` inside the closure)?

pkg/console/controllers/oauthclientsecret/oauthclientsecret.go
jhadvig: You're right.... I verified that `CustomApplyOAuth` does a fresh `Get()` internally on every call (`oauthclient.go:24`), and `EnsureObjectMeta` only merges labels/annotations/ownerrefs without touching `ResourceVersion`. So the `Update()` always carries the fresh `resourceVersion` from that internal `Get`, regardless of what `clientCopy` has. Adding another `Get` in the retry closure would just be a redundant API call. My bad on this one.

pkg/console/controllers/upgradenotification/controller.go
jhadvig: You're right.... I verified that `retry.OnError` returns non-retryable errors directly to the caller (it doesn't swallow them). So `AlreadyExists` will still surface as `createErr`, and the outer `!apierrors.IsAlreadyExists(createErr)` guard is necessary to prevent it from propagating as `"FailedCreate"`. The two layers serve different purposes: the retry filter avoids wasted retries, the outer check avoids treating it as a failure.

pkg/console/controllers/util/retry.go
jhadvig: @sg00dwin still valid

sg00dwin: @jhadvig  I have removed the nil test case

pkg/console/operator/retry.go
sg00dwin: done

jhadvig (approved):
/lgtm

pkg/console/operator/retry.go
TheRealJon: Yeah, I think removing that would be best, tbh.

jhadvig:
manual overide of flaky test

PR #1169, opened 2026-05-29
OCPBUGS-64688: use ObservedGeneration to determine Progressing status

sg00dwin:
## Summary
- Replace `IsAvailableAndUpdated` with `ObservedGeneration < Generation` check for the Progressing condition
- `IsAvailableAndUpdated` used `UpdatedReplicas == Replicas`, which falsely triggers during node reboots when pods are temporarily disrupted — even though the operator made no spec changes
- `ObservedGeneration < Generation` only fires when the operator has updated the deployment spec and the deployment controller hasn't processed it yet
- Extract `checkDeploymentGenerationProgress` helper for testability and add unit tests

## Root Cause
During node reboots (MCO draining nodes), console pods are terminated and replacements are created. During the ~30-second gap, `UpdatedReplicas != Replicas`, causing `IsAvailableAndUpdated` to return false and the operator to report `Progressing=True`. The operator made no changes — the disruption is external.

## Why This Fix
`ObservedGeneration < Generation` precisely detects operator-initiated spec changes:
- **Node reboot (no spec change):** Generation unchanged → Progressing=False ✓
- **Real upgrade (operator updates spec):** Generation bumps → Progressing=True ✓
- **Fresh install:** Generation=1, ObservedGeneration=0 → Progressing=True ✓

The Available condition (`IsAvailable` at line 231) independently monitors pod readiness and is unaffected.

## Test plan
- [x] `make` builds successfully
- [x] `make test-unit` all tests pass
- [x] `gofmt` / `go vet` clean
- [x] Unit tests for `checkDeploymentGenerationProgress`: ObservedGeneration equal, less than, greater than Generation, both zero, and initial rollout
- [x] Manual verification on live OCP 4.22 cluster (node drain stays Progressing=False, spec change triggers Progressing=True)
- [ ] CI e2e upgrade test `[Monitor:legacy-cvo-invariants][bz-Management Console] clusteroperator/console should stay Progressing=False while MCO is Progressing=True` should stop failing (~8% failure rate on Sippy currently)

Assisted-by: Claude Code (Opus 4.6)

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable rollout status: deployments are now marked "progressing" only when the cluster hasn’t observed the latest deployment generation, avoiding noisy status flips caused by fluctuating replica counts during external disruptions.

* **Tests**
  * Added tests covering observed-generation scenarios (equal, less-than, greater-than, and initial rollout) to validate consistent progression reporting.

### DIFF
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -211,13 +211,8 @@ func (co *consoleOperator) sync_v400(ctx context.Context, controllerContext fact
 
 	statusHandler.AddCondition(status.HandleProgressing("SyncLoopRefresh", "InProgress", func() error {
 		version := os.Getenv("OPERATOR_IMAGE_VERSION")
-		// Only report Progressing=True when the deployment is actually rolling out
-		// or the operator version is changing. Do NOT report Progressing just because
-		// resources were updated during reconciliation, as per the API guidelines:
-		// "Operators should not report Progressing when they are reconciling (without action)
-		// a previously known state."
-		if !deploymentsub.IsAvailableAndUpdated(actualDeployment) {
-			return fmt.Errorf("working toward version %s, %v replicas available", version, actualDeployment.Status.AvailableReplicas)
+		if err := checkDeploymentGenerationProgress(actualDeployment); err != nil {
+			return err
 		}
 
 		if co.versionGetter.GetVersions()["operator"] != version {
@@ -754,6 +749,19 @@ func (co *consoleOperator) GetAvailablePlugins(enabledPluginsNames []string) []*
 		return availablePlugins[i].Name < availablePlugins[j].Name
 	})
 	return availablePlugins
+}
+
+// checkDeploymentGenerationProgress returns an error if the deployment controller
+// has not yet processed the latest spec change (ObservedGeneration < Generation).
+// This is used to determine Progressing status without relying on replica counts,
+// which fluctuate during external disruptions like node reboots.
+// See https://issues.redhat.com/browse/OCPBUGS-64688
+func checkDeploymentGenerationProgress(deployment *appsv1.Deployment) error {
+	if deployment.Status.ObservedGeneration < deployment.Generation {
+		return fmt.Errorf("deployment generation %d not yet observed (observed: %d)",
+			deployment.Generation, deployment.Status.ObservedGeneration)
+	}
+	return nil
 }
 
 func getNodeComputeEnvironments(nodes []*corev1.Node) ([]string, []string) {
--- a/pkg/console/operator/sync_v400_test.go
+++ b/pkg/console/operator/sync_v400_test.go
@@ -4,9 +4,12 @@ import (
 	"testing"
 
 	"github.com/go-test/deep"
-	"github.com/openshift/console-operator/pkg/api"
+
+	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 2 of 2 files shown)


pkg/console/operator/sync_v400.go
```diff
@@ -211,13 +211,13 @@ func (co *consoleOperator) sync_v400(ctx context.Context, controllerContext fact
...
-		// "Operators should not report Progressing when they are reconciling (without action)
-		// a previously known state."
-		if !deploymentsub.IsAvailableAndUpdated(actualDeployment) {
-			return fmt.Errorf("working toward version %s, %v replicas available", version, actualDeployment.Status.AvailableReplicas)
+		// Only report Progressing when the deployment controller hasn't yet
+		// processed a spec change made by the operator. Do not use replica
+		// counts — they fluctuate during external disruptions (node reboots)
+		// that the operator did not cause. See https://issues.redhat.com/browse/OCPBUGS-64688
+		if actualDeployment.Status.ObservedGeneration < actualDeployment.Generation {
+			return fmt.Errorf("deployment generation %d not yet observed (current: %d)",
```
Leo6Leo: The parenthetical says "current" but passes ObservedGeneration. Consider changing the wording to make it more accurate:
```suggestion
			return fmt.Errorf("deployment generation %d not yet observed (observed: %d)",
```

pkg/console/operator/sync_v400.go
Leo6Leo, line 1: @sg00dwin Could you add some unit tests for the new behavior?

The old IsAvailableAndUpdated at least had its own unit tests in deployment_test.go. The new inline comparison ObservedGeneration < Generation has no test coverage. A table-driven test with cases like: - ObservedGeneration == Generation → no error (not progressing) - ObservedGeneration < Generation → returns error (progressing) - ObservedGeneration > Generation → no error (edge case, already processed)

would help prevent regressions.

Leo6Leo:
/cc @Leo6Leo 

Spinning cluster to do QE currently

Leo6Leo:
Manual Verification

Deployed the PR build to a live OCP 4.22 cluster (3 control-plane nodes, Azure).

Node drain (should stay Progressing=False): Drained a node hosting a console pod. Pod was evicted and rescheduled. Progressing stayed False throughout — lastTransitionTime never changed.

Spec change (should trigger Progressing=True): Changed logLevel to TraceAll, bumping deployment generation 3→4. Progressing correctly flipped to True ("deployment generation 4 not yet observed (current: 3)"), then resolved to False 2s later.

Both scenarios behave as expected.

Currently just waiting for the CI to past before I give the verify label.

pkg/console/operator/sync_v400.go
theobarberbany: I'd also be considering adding an e2e to one of the upgrade jobs? If we can write one that isn't a giant pain.

sg00dwin: **Unit tests:** Added table-driven unit tests for the ObservedGeneration-based Progressing check covering: equal, less than, greater than, both zero, and initial rollout scenarios.

**Regarding e2e test:** Good idea. There's already a story tracking this under the same epic: [CONSOLE-5188 (e2e testing automation)](https://redhat.atlassian.net/browse/CONSOLE-5188). It's currently unassigned and empty description. Added this suggestion for adding a console-owned e2e assertion to one of the upgrade jobs.

TheRealJon (approved):
/lgtm

jhadvig (approved):
/verified by CI

PR #1174, opened 2026-06-23
CONSOLE-5271: Set olmLifecycleEnabled based on OLMLifecycleAndCompatibility FeatureGate

perdasilva:
**Analysis / Root cause**:
The operator lifecycle metadata UI in the console (cluster compatibility and support phase columns on the Installed Operators page) needs to be gated on the `OLMLifecycleAndCompatibility` cluster FeatureGate. The console frontend cannot safely read the `FeatureGate` resource directly because not all users have RBAC for `config.openshift.io/v1 FeatureGate`. The console-operator, which runs with a service account, is the correct place to read the FeatureGate and pass the result to the console via the ConsoleConfig.

**Solution description**:
- Add `SyncOLMLifecycle()` method that reads the cluster `FeatureGate` resource and checks `status.featureGates[].enabled[]` for the `OLMLifecycleAndCompatibility` gate name
- Add `OLMLifecycleEnabled` field to `ClusterInfo` struct and the `ConsoleServerCLIConfigBuilder`
- Thread `olmLifecycleEnabled` through `SyncConfigMap()` → `DefaultConfigMap()` → both config builder chains
- The generated `console-config.yaml` now includes `clusterInfo.olmLifecycleEnabled: true` when the gate is enabled

**Test setup:**
Requires a cluster with the `OLMLifecycleAndCompatibility` FeatureGate enabled.

**Test cases:**
- Verify `olmLifecycleEnabled: true` appears in the console ConfigMap when the `OLMLifecycleAndCompatibility` gate is enabled in the cluster FeatureGate
- Verify `olmLifecycleEnabled` is absent/false when the gate is not enabled
- Verify graceful fallback when the FeatureGate resource cannot be read
- Unit tests for `DefaultConfigMap` with `olmLifecycleEnabled` true and false

**Browser conformance**:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari (or Epiphany on Linux)

**Additional info:**
Companion PR to openshift/console#16668 which adds the `--olm-lifecycle` bridge flag and reads `window.SERVER_FLAGS.olmLifecycle` in the frontend.

**Reviewers and assignees:**
/assign @perdasilva

## Summary by CodeRabbit

* **New Features**
  * The console operator now reads the `OLMLifecycleAndCompatibility` feature gate and propagates the resulting enablement into the generated console configuration, including updated console-config YAML.
* **Tests**
  * Updated existing configmap tests and added new coverage to verify the `OLMLifecycleMetadataEnabled` toggle is correctly set and reflected after unmarshaling.

### DIFF
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -130,6 +130,12 @@ func (co *consoleOperator) sync_v400(ctx context.Context, controllerContext fact
 		return statusHandler.FlushAndReturn(techPreviewErr)
 	}
 
+	olmLifecycleMetadataEnabled, olmLifecycleMetadataErrReason, olmLifecycleMetadataErr := co.SyncOLMLifecycleMetadata()
+	statusHandler.AddConditions(status.HandleProgressingOrDegraded("OLMLifecycleMetadataSync", olmLifecycleMetadataErrReason, olmLifecycleMetadataErr))
+	if olmLifecycleMetadataErr != nil {
+		return statusHandler.FlushAndReturn(olmLifecycleMetadataErr)
+	}
+
 	cm, cmErrReason, cmErr := co.SyncConfigMap(
 		ctx,
 		set.Operator,
@@ -141,6 +147,7 @@ func (co *consoleOperator) sync_v400(ctx context.Context, controllerContext fact
 		controllerContext.Recorder(),
 		consoleURL.Hostname(),
 		techPreviewEnabled,
+		olmLifecycleMetadataEnabled,
 	)
 	statusHandler.AddConditions(status.HandleProgressingOrDegraded("ConfigMapSync", cmErrReason, cmErr))
 	if cmErr != nil {
@@ -352,6 +359,7 @@ func (co *consoleOperator) SyncConfigMap(
 	recorder events.Recorder,
 	consoleHost string,
 	techPreviewEnabled bool,
+	olmLifecycleMetadataEnabled bool,
 ) (consoleConfigMap *corev1.ConfigMap, reason string, err error) {
 
 	managedConfig, mcErr := co.managedNSConfigMapLister.ConfigMaps(api.OpenShiftConfigManagedNamespace).Get(api.OpenShiftConsoleConfigMapName)
@@ -426,6 +434,7 @@ func (co *consoleOperator) SyncConfigMap(
 		telemetryConfig,
 		consoleHost,
 		techPreviewEnabled,
+		olmLifecycleMetadataEnabled,
 	)
 	if err != nil {
 		return nil, "FailedConsoleConfigBuilder", err
@@ -599,7 +608,7 @@ func (co *consoleOperator) SyncTrustedCAConfigMap(ctx context.Context, operatorC
 func (co *consoleOperator) SyncTechPreview() (techPreviewEnabled bool, reason string, err error) {
 	featureGate, err := co.featureGateLister.Get(api.ConfigResourceName)
 	if err != nil {
-		klog.V(4).Infof("failed to get FeatureGate resource: %v.", err)
+		klog.V(4).Infof("failed to get FeatureGate resource: %v", err)
 		return false, "FailedGet", err
 	}
 
@@ -609,6 +618,29 @@ func (co *consoleOperator) SyncTechPreview() (techPreviewEnabled bool, reason st
 		klog.V(4).Infoln("Console Technology Preview features enabled based on cluster FeatureSet TechPreviewNoUpgrade.")
 	}
 	return techPreviewEnabled, "", nil
+}
+
+// Replace with features.FeatureGateOLMLifecycleAndCompatibility once openshift/api is bumped.
+const olmLifecycleAndCompatibilityFeatureGate configv1.FeatureGateName = "OLMLifecycleAndCompatibility"
+
+// SyncOLMLifecycleMetadata determines if OLM lifecycle metadata features should be enabled
+// by checking if the OLMLifecycleAndCompatibility feature gate is enabled in the cluster.
+func (co *consoleOperator) SyncOLMLifecycleMetadata() (olmLifecycleMetadataEnabled bool, reason string, err error) {
+	featureGate, err := co.featureGateLister.Get(api.ConfigResourceName)
+	if err != nil {
+		klog.V(4).Infof("failed to get FeatureGate resource: %v", err)
+		return false, "FailedGet", err
+	}
+
+	for _, versionedGates := range featureGate.Status.FeatureGates {
+		for _, gate := range versionedGates.Enabled {
+			if gate.Name == olmLifecycleAndCompatibilityFeatureGate {
+				klog.V(4).Infoln("OLM lifecycle metadata features are enabled based on the OLMLifecycleAndCompatibility feature gate")
+				return true, "", nil
+			}
+		}
+	}
+	return false, "", nil
 }
 
 func (co *consoleOperator) SyncCustomLogos(operatorConfig *operatorv1.Console) (error, string) {
--- a/pkg/console/subresource/configmap/configmap.go
+++ b/pkg/console/subresource/configmap/configmap.go
@@ -49,6 +49,7 @@ func DefaultConfigMap(
 	telemeterConfig map[string]string,
 	consoleHost string,
 	techPreviewEnabled bool,
+	olmLifecycleMetadataEnabled bool,
 ) (consoleConfigMap *corev1.ConfigMap, unsupportedOverridesHaveMerged bool, err error) {
 
 	apiServerURL := infrastructuresub.GetAPIServerURL(infrastructureConfig)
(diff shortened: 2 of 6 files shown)


manifests/03-rbac-role-cluster-openshift-console-user-operator-lifecycle-reader.yaml
```diff
@@ -0,0 +1,45 @@
+# RBAC required for the OpenShift Console to access the lifecycle-server API.
```
perdasilva: I don't think we need this

TheRealJon (approved):
/lgtm

perdasilva:
Docs Approver:
cc @jseseCCS

PX Approver:
cc @rh-joshbeverly

TheRealJon (approved):
/lgtm

pkg/console/subresource/configmap/configmap.go
```diff
@@ -106,6 +108,7 @@ func DefaultConfigMap(
 		AuthConfig(authConfig, apiServerURL).
 		Capabilities(operatorConfig.Spec.Customization.Capabilities).
 		TechPreviewEnabled(techPreviewEnabled).
+		OLMLifecycleMetadataEnabled(olmLifecycleMetadataEnabled).
 		ConfigYAML()
 	if err != nil {
 		klog.Errorf("failed to generate user defined console-config config: %v", err)
```
jseseCCS: redundant repetition of "config" ("console-config config"). -->

"failed to generate user-defined console-config: %v"?

also note, "user defined" should be hyphenated bc it's compound modifier + succeeding noun

pkg/console/subresource/configmap/configmap.go
```diff
@@ -66,6 +67,7 @@ func DefaultConfigMap(
 		NodeOperatingSystems(nodeOperatingSystems).
 		CopiedCSVsDisabled(copiedCSVsDisabled).
 		TechPreviewEnabled(techPreviewEnabled).
+		OLMLifecycleMetadataEnabled(olmLifecycleMetadataEnabled).
 		ConfigYAML()
 	if err != nil {
 		klog.Errorf("failed to generate default console-config config: %v", err)
```
jseseCCS: remove redundant dupe "config"? --> 

"failed to generate default console-config: %v"

pkg/console/subresource/configmap/configmap.go
```diff
@@ -49,6 +49,7 @@ func DefaultConfigMap(
 	telemeterConfig map[string]string,
```
jseseCCS, line 49: GLOBAL

compound modifiers preceding nouns should be hyphenated (ex., "user-defined"). 

2. from user/operator perspective, these messages say _that_ something failed, but don't answer "what do I do next?" _strongly_ urge adding at least brief hint about impact or next step.

pkg/console/operator/sync_v400.go
```diff
@@ -612,6 +621,29 @@ func (co *consoleOperator) SyncTechPreview() (techPreviewEnabled bool, reason st
...
+	featureGate, err := co.featureGateLister.Get(api.ConfigResourceName)
+	if err != nil {
+		klog.V(4).Infof("failed to get FeatureGate resource: %v.", err)
+		return false, "FailedGet", err
+	}
+
+	for _, versionedGates := range featureGate.Status.FeatureGates {
+		for _, gate := range versionedGates.Enabled {
+			if gate.Name == olmLifecycleAndCompatibilityFeatureGate {
+				klog.V(4).Infoln("OLM lifecycle metadata features enabled based on OLMLifecycleAndCompatibility feature gate.")
```
jseseCCS: improve readability: add definite article "the". add "are" too in front of "enabled" bc i stumbled over that bit a time or two before i understood its rhythm. at first read it sounds like the features have enabled something. 

-->

"OLM lifecycle metadata features are enabled based on the OLMLifecycleAndCompatibility feature gate."

pkg/console/operator/sync_v400.go
```diff
@@ -612,6 +621,29 @@ func (co *consoleOperator) SyncTechPreview() (techPreviewEnabled bool, reason st
...
 
+// Replace with features.FeatureGateOLMLifecycleAndCompatibility once openshift/api is bumped.
+const olmLifecycleAndCompatibilityFeatureGate configv1.FeatureGateName = "OLMLifecycleAndCompatibility"
+
+// SyncOLMLifecycleMetadata determines if OLM lifecycle metadata features should be enabled
+// by checking if the OLMLifecycleAndCompatibility feature gate is enabled in the cluster.
+func (co *consoleOperator) SyncOLMLifecycleMetadata() (olmLifecycleMetadataEnabled bool, reason string, err error) {
+	featureGate, err := co.featureGateLister.Get(api.ConfigResourceName)
+	if err != nil {
+		klog.V(4).Infof("failed to get FeatureGate resource: %v.", err)
```
jseseCCS: for consistency, use lowercase for "feature gate" and remove trailing period? -->

"failed to get feature gate resource: %v"

pkg/console/operator/sync_v400.go
```diff
@@ -612,6 +621,29 @@ func (co *consoleOperator) SyncTechPreview() (techPreviewEnabled bool, reason st
...
+		return false, "FailedGet", err
```
jseseCCS, line 632: 633: ambiguous for anyone troubleshooting cluster statuses/events. strongly urge being more specific about what resource failed to load. -->

return false, "FailedGetFeatureGate", err

for ex.

pkg/console/operator/sync_v400.go
```diff
@@ -130,6 +130,12 @@ func (co *consoleOperator) sync_v400(ctx context.Context, controllerContext fact
 		return statusHandler.FlushAndReturn(techPreviewErr)
 	}
 
+	olmLifecycleMetadataEnabled, olmLifecycleMetadataErrReason, olmLifecycleMetadataErr := co.SyncOLMLifecycleMetadata()
+	statusHandler.AddConditions(status.HandleProgressingOrDegraded("OLMLifecycleMetadataSync", olmLifecycleMetadataErrReason, olmLifecycleMetadataErr))
```
jseseCCS, line 134: minor terminology check: if status string visible to users, use full word "Synchronization"?

"OLMLifecycleMetadataSynchronization"

not a hill i'd 💀 on though!

jseseCCS (approved):
made just a few teensy tiny comments. conditionally approving in good faith that you'll review, apply as appropriate. 😺

pkg/console/subresource/configmap/configmap.go
perdasilva: Fixed both — dropped the redundant "config" and hyphenated "user-defined". Also fixed the same redundancy on line 73 (`default console-config config`).

pkg/console/subresource/configmap/configmap.go
perdasilva: Fixed — dropped the redundant "config".

pkg/console/subresource/configmap/configmap.go
perdasilva: Hyphenation is fixed.

On the remediation hints — these are operator-internal `klog.Errorf` messages, not user-facing. The existing pattern across the codebase (`sync_v400.go`, `operator.go`, `storageversionmigration/controller.go`, etc.) is consistently `"failed to X: %v"` without next-step guidance. The error detail comes through in `%v`, and the operator's reconciliation loop retries automatically. Adding hints here would be inconsistent with the rest of the codebase and these specific errors already bubble up as return values to the caller. Happy to discuss further if you feel strongly about it though.

pkg/console/operator/sync_v400.go
perdasilva: Fixed — added "are" and "the".

pkg/console/operator/sync_v400.go
perdasilva: Dropped the trailing period (matches the majority convention in this file — 35 log messages without vs 4 with). Keeping "FeatureGate" capitalized though — `starter.go` consistently uses it as the Kubernetes resource Kind name (`"timed out waiting for FeatureGate detection"`, `"FeatureGates initialized"`).

pkg/console/operator/sync_v400.go
perdasilva: `"FailedGet"` is the established convention across this file — it's used in 9 places (lines 108, 184, 190, 537, 581, 613, 633, 695, 776) including the existing `SyncTechPreview` method which does the identical FeatureGate lookup and returns `"FailedGet"` on the same error path. The specificity comes from how the caller consumes the reason (e.g. `HandleProgressingOrDegraded("OAuthClientSecretGet", "FailedGet", err)` — the first arg provides the resource context). Changing just this instance would be inconsistent.

pkg/console/operator/sync_v400.go
perdasilva: Keeping `"Sync"` — every existing condition name in this file uses that abbreviation: `CustomLogoSync`, `TechPreviewSync`, `ConfigMapSync`, `ServiceCASync`, `TrustedCASync`, `DeploymentSync`.

TheRealJon (approved):
/lgtm

PR #1175, opened 2026-06-23
HELM-639: Update Helm download manifests for Helm 3 and Helm 4.

sowmya-sl:
Split the CLI download entries and clarify Helm 4 description relative to Helm 3 so users can distinguish versions.

**Analysis / Root cause**:
Helm 4 has been released recently. Update Command Line Tools to reflect that. 

## Release Notes

* **New Features**
  * Updated the Helm CLI download option to Helm v4 (with the version-specific Helm 4.1.4 mirror link).
* **Bug Fixes**
  * Refreshed the Helm CLI entry details (description and display name) to match Helm v4 instead of Helm v3.

### DIFF
--- a/manifests/07-downloads-helm.yaml
+++ b/manifests/07-downloads-helm.yaml
@@ -1,7 +1,7 @@
 apiVersion: console.openshift.io/v1
 kind: ConsoleCLIDownload
 metadata:
-  name: helm-download-links-v3
+  name: helm-download-links
   annotations:
     include.release.openshift.io/hypershift: "true"
     include.release.openshift.io/ibm-cloud-managed: "true"
@@ -10,28 +10,9 @@ metadata:
     capability.openshift.io/name: Console
 spec:
   description: |
-    Helm 3 is a package manager for Kubernetes applications which enables defining,
+    Helm 4 is a package manager for Kubernetes applications which enables defining,
     installing, and upgrading applications packaged as Helm Charts.
-  displayName: helm - Helm 3 CLI
-  links:
-    - href: 'https://mirror.openshift.com/pub/openshift-v4/clients/helm/latest'
-      text: Download Helm
----
-apiVersion: console.openshift.io/v1
-kind: ConsoleCLIDownload
-metadata:
-  name: helm-download-links-v4
-  annotations:
-    include.release.openshift.io/hypershift: "true"
-    include.release.openshift.io/ibm-cloud-managed: "true"
-    include.release.openshift.io/self-managed-high-availability: "true"
-    include.release.openshift.io/single-node-developer: "true"
-    capability.openshift.io/name: Console
-spec:
-  description: |
-    Helm 4 is the next major release of Helm and, compared to Helm 3, includes
-    updated workflows for defining, installing, and upgrading Helm Charts.
   displayName: helm - Helm 4 CLI
   links:
     - href: 'https://mirror.openshift.com/pub/cgw/helm/4.1.4'
-      text: Download Helm
+      text: Download Helm v4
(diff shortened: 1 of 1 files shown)


manifests/07-downloads-helm.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: console.openshift.io/v1
 kind: ConsoleCLIDownload
 metadata:
-  name: helm-download-links
+  name: helm-download-links-v3
```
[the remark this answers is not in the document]
webbnh: @sowmya-sl, this PR is going to be merged into the `main` branch which will target OCP 5.0 (and 4.23).  In that version (and probably in 4.23), we're only going to provide Helm v4.  So, you should leave the name `helm-download-links` unchanged and update the content with stand-alone text for Helm v4 (similar to the current text for Helm v3, without the comparison to the previous version).  This will produce the correct results going forward, and it will get us around Coderabbitai's concern, here.

Then, when you backport this change to OCP 4.22, you should add a second CRD for Helm v3.  If you like, in that manifest, you can restore the previous description and make the new description for Helm v4 more introductory (like you have here, currently).  But, you should keep the `helm-download-links` for one of the two CRDs and use a new name for the other.

Also, in the backport, the text of the hyperlinks should be different -- e.g., "Download Helm v3" vs. "Download Helm v4".

manifests/07-downloads-helm.yaml
```diff
@@ -16,3 +16,22 @@ spec:
...
+  annotations:
+    include.release.openshift.io/hypershift: "true"
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    capability.openshift.io/name: Console
+spec:
+  description: |
+    Helm 4 is the next major release of Helm and, compared to Helm 3, includes
+    updated workflows for defining, installing, and upgrading Helm Charts.
```
webbnh: I think that the v3 description of Helm is better -- it's a tool for defining, installing, and upgrading _applications_.  (I don't think Helm v4 provides much in the way of new capabilities for manipulating _charts_ -- it provides new ways to use charts to define and maintain **applications**.)

webbnh (changes requested):
There's a context problem with this PR:  this PR is targeting OCP 5.0, but it's carrying a change which is targeting OCP 4.22.

I think you're on the right path -- we should first change the `main` branch (which this PR is targeting) and then backport the change to OCP 4.22 -- but the changes needed for 5.0 and 4.22 are _different_, so you should proceed with the 5.0 change but then do a manual backport to 4.22 with different content.

webbnh:
Also, the PR title is going to need the Jira prefix.

manifests/07-downloads-helm.yaml
sowmya-sl: Do I retain the same text for v4 then?

manifests/07-downloads-helm.yaml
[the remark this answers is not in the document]
sowmya-sl: Changed

manifests/07-downloads-helm.yaml
webbnh: What is there now looks good for OCP 5.0.

When you backport to OCP 4.22, I think something like the text which you originally proposed would be better:  the Helm v3 text should be largely unchanged, and if would probably be good for the Helm v4 entry to carry some "introductory verbiage" (similar to what you had).  (But, only introduce one new CRD -- keep the old one for the second entry.)

sowmya-sl:
Accidentally added some labels here. Please ignore.

jhadvig (approved):

jseseCCS:
note: this merged without docs sign-off. reviewed the diff now, just Helm 3 to Helm 4 version bump on download manifest (description, displayName, download link). no style guide issues in strings themselves. the follow-up PR #1197 added Helm 3 download links back + Helm 4. i left review notes there already.

PR #1178, opened 2026-06-29
OCPBUGS-93982: skip generation check when deployment was just updated

sg00dwin:
## Summary

During upgrades, the console operator updates the console deployment spec 2-4 times (new image, config changes). Each update bumps `Generation`, but `ObservedGeneration` lags by 257ms-1s until the deployment controller processes it. The generation check (introduced by [PR #1169](https://github.com/openshift/console-operator/pull/1169) for [OCPBUGS-64688](https://issues.redhat.com/browse/OCPBUGS-64688)) sees this gap and reports `Progressing=True` — even though the operator itself just caused the gap.

These sub-second blips coincide with MCO's `Progressing=True` window, failing the OTA invariant test (`clusteroperator/console should stay Progressing=False while MCO is Progressing=True`) in ~5% of upgrade jobs.

## Changes

`SyncDeployment` already calls `ApplyDeployment`, which returns a `changed` boolean — but it was being discarded (removed by [PR #1074](https://github.com/openshift/console-operator/pull/1074) for [OCPBUGS-64685](https://issues.redhat.com/browse/OCPBUGS-64685)). This PR re-introduces it:

- **`SyncDeployment`**: Captures `changed` from `ApplyDeployment` and returns it to the caller
- **`sync_v400`**: Skips the generation check when `changed=true` (self-inflicted gap, guaranteed transient)

When `changed=false`, the generation check runs normally — catching genuine deployment rollout delays. The next sync loop always runs with `changed=false` if no new changes are needed, so persistent gaps are still detected.

## Context

This is the 4th fix in a chain of Progressing-related bugs:

| Bug | Problem | Fix |
|-----|---------|-----|
| [OCPBUGS-36830](https://issues.redhat.com/browse/OCPBUGS-36830) / [OCPBUGS-64685](https://issues.redhat.com/browse/OCPBUGS-64685) | Any resource change → false Progressing | Removed change tracking, used replica count |
| [OCPBUGS-64688](https://issues.redhat.com/browse/OCPBUGS-64688) | Node reboots → false Progressing (30s+) | Switched to generation check |
| **[OCPBUGS-93982](https://issues.redhat.com/browse/OCPBUGS-93982)** | **Self-inflicted spec updates → false Progressing (<1s)** | **Guard generation check with `changed` flag** |

## Tests

- [ ] New unit tests: `TestDeploymentProgressingSkippedWhenChanged` — 4 cases covering changed/unchanged x gap/no-gap
- [ ] Existing unit tests: `TestDeploymentProgressingByGeneration` — 5 cases unchanged and passing
- [ ] `make test-unit` passes (all packages, gofmt, govet)
- [ ] CI e2e: verify no `SyncLoopRefreshProgressing` blips during upgrade window

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deployment status handling to avoid brief, misleading “Progressing” updates right after changes are applied.
  * Deployment updates now report more consistently when a change was made, helping status checks reflect the latest state.
  * Added coverage to ensure progress reporting behaves correctly across updated and unchanged deployments.

### DIFF
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -185,7 +185,7 @@ func (co *consoleOperator) sync_v400(ctx context.Context, controllerContext fact
 		return statusHandler.FlushAndReturn(servingCertErr)
 	}
 
-	actualDeployment, depErrReason, depErr := co.SyncDeployment(
+	actualDeployment, depChanged, depErrReason, depErr := co.SyncDeployment(
 		ctx,
 		set.Operator,
 		cm,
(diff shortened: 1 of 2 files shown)


jhadvig (approved):
@sg00dwin LGTM 👍

Clean fix, the guard logic makes sense.... when the operator itself caused the generation bump, reporting Progressing=True for that sub-second lag is just noise. And it's self-correcting since the next sync loop runs with `changed=false` and catches any real persistent gap.

One thing worth considering from CodeRabbit's review, extracting the guard into a small helper like `deploymentProgressingErr(depChanged, deployment)` so the test exercises the real logic instead of duplicating the `if !depChanged` branch inline. Not a blocker, but would be a nice cleanup.

/lgtm

sg00dwin:
/verified by TestDeploymentProgressingSkippedWhenChanged, TestDeploymentProgressingByGeneration, @sg00dwin

PR #1179, opened 2026-06-30
OCPBUGS-67134: add grace period before reporting Available=False

sg00dwin:
## Summary
- During disruptive CI tests (node reboots, drains), console deployment replicas briefly drop to zero for ~10 seconds before self-recovering
- The operator immediately reports `Available=False` with `Deployment_InsufficientReplicas`, triggering OTA invariant test failures
- This fix adds a 15-second grace period that suppresses the false alarm when the deployment was recently available, while still reporting genuine outages promptly

## What changed
- Track when the deployment was last seen available (`lastDeploymentAvailableTime` on the operator struct)
- Extract `evaluateDeploymentAvailability()` method that checks the grace window before reporting `Available=False`
- Grace period (15s) is kept short so genuine production outages are still reported promptly — long enough to absorb the observed ~10s blips but not so long that it masks real problems
- First startup with no prior healthy state reports immediately — no risk of masking real outages

## Related
- [OCPBUGS-67134](https://issues.redhat.com/browse/OCPBUGS-67134)
- Epic: [CONSOLE-5185](https://issues.redhat.com/browse/CONSOLE-5185)
- Follows the same operator-level tolerance pattern as [#834](https://github.com/openshift/console-operator/pull/834) (route health retry), [#1164](https://github.com/openshift/console-operator/pull/1164) (degraded blips), [#1169](https://github.com/openshift/console-operator/pull/1169) (progressing/ObservedGeneration)

## Test plan
- [x] 7 new unit tests covering: available deployment, first-sync unavailable, grace period suppression, grace period expiry, recovery after blip, boundary condition, error message format
- [x] All existing tests pass
- [x] `make test-unit` clean (go test, gofmt, govet)
- [ ] After merge: monitor Sippy for ~5-7 days, then remove origin invariant test exception for `console` + `Deployment_InsufficientReplicas`

Co-Authored-By: Claude Opus 4.6

### DIFF
--- a/pkg/console/operator/operator.go
+++ b/pkg/console/operator/operator.go
@@ -94,6 +94,8 @@ type consoleOperator struct {
 	trackables trackables
 
 	monitoringDeploymentLister appsv1listers.DeploymentLister
+
+	lastDeploymentAvailableTime time.Time
 }
 
 type trackables struct {
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -8,6 +8,7 @@ import (
 	"slices"
 	"sort"
 	"strings"
+	"time"
 
 	// kube
 	appsv1 "k8s.io/api/apps/v1"
@@ -45,6 +46,13 @@ import (
 	utilsub "github.com/openshift/console-operator/pkg/console/subresource/util"
 	telemetry "github.com/openshift/console-operator/pkg/console/telemetry"
 )
+
+// deploymentAvailableGracePeriod is the duration the operator tolerates zero
+// available replicas before reporting DeploymentAvailable=False. This absorbs
+// brief disruptions (e.g. node reboots during conformance-serial tests) that
+// take all replicas offline for ~10 seconds. Kept short so genuine production
+// outages are still reported promptly.
+const deploymentAvailableGracePeriod = 15 * time.Second
 
 // The sync loop starts from zero and works its way through the requirements for a running console.
 // If at any point something is missing, it creates/updates that piece and immediately dies.
@@ -221,13 +229,7 @@ func (co *consoleOperator) sync_v400(ctx context.Context, controllerContext fact
 		return nil
 	}()))
 
-	statusHandler.AddCondition(status.HandleAvailable(func() (prefix string, reason string, err error) {
-		prefix = "Deployment"
-		if !deploymentsub.IsAvailable(actualDeployment) {
-			return prefix, "InsufficientReplicas", fmt.Errorf("%v replicas available for console deployment", actualDeployment.Status.ReadyReplicas)
-		}
-		return prefix, "", nil
-	}()))
+	statusHandler.AddCondition(status.HandleAvailable(co.evaluateDeploymentAvailability(actualDeployment)))
 
 	// if we survive the gauntlet, we need to update the console config with the
 	// public hostname so that the world can know the console is ready to roll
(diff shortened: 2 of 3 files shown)


jhadvig:
@sg00dwin good investigation on this one, the root cause analysis is solid and the tests are well written 👍

I've been thinking about this more though, and I'm not sure suppressing `Available=False` for 2 minutes is the right fix here. `Available` is deliberately given no inertia in library-go - it's meant to be a real-time signal of "can users use this right now?". Reporting `Available=True` when replicas are actually at zero is lying about cluster state, and CVO/monitoring rely on that signal during upgrades and alerting.

The real question is: why are *all* replicas going down during a single node reboot? If they're colocated on the same node, that's the root cause we should fix.

A few alternatives worth exploring:

1. **Pod topology spread constraints / anti-affinity** - ensure console replicas land on different nodes so a single node disruption can't take all of them out. This fixes the actual problem rather than masking it.

2. **PodDisruptionBudget** - if there isn't one already, a PDB with `minAvailable: 1` would prevent voluntary drains from evicting all replicas simultaneously.

3. **Fix it at the test/invariant layer** - if conformance-serial tests do disruptive node operations that are *expected* to cause brief disruptions, the invariant tests should tolerate brief `Available=False` during those windows rather than the operator hiding it. Could be worth discussing with the TRT team.

4. **Propose `Available` inertia in library-go** - if this is a real gap that other operators also hit, the proper fix is adding `WithAvailableInertia` to `StatusSyncer` so it's done consistently for everyone, not ad-hoc per operator.

Can you dig into whether we have topology constraints on the console deployment? That feels like the most impactful fix here.

sg00dwin:
> @sg00dwin good investigation on this one, the root cause analysis is solid and the tests are well written 👍
> 
> I've been thinking about this more though, and I'm not sure suppressing `Available=False` for 2 minutes is the right fix here. `Available` is deliberately given no inertia in library-go - it's meant to be a real-time signal of "can users use this right now?". Reporting `Available=True` when replicas are actually at zero is lying about cluster state, and CVO/monitoring rely on that signal during upgrades and alerting.
> 
> The real question is: why are _all_ replicas going down during a single node reboot? If they're colocated on the same node, that's the root cause we should fix.
> 
> A few alternatives worth exploring:
> 
> 1. **Pod topology spread constraints / anti-affinity** - ensure console replicas land on different nodes so a single node disruption can't take all of them out. This fixes the actual problem rather than masking it.
> 2. **PodDisruptionBudget** - if there isn't one already, a PDB with `minAvailable: 1` would prevent voluntary drains from evicting all replicas simultaneously.
> 3. **Fix it at the test/invariant layer** - if conformance-serial tests do disruptive node operations that are _expected_ to cause brief disruptions, the invariant tests should tolerate brief `Available=False` during those windows rather than the operator hiding it. Could be worth discussing with the TRT team.
> 4. **Propose `Available` inertia in library-go** - if this is a real gap that other operators also hit, the proper fix is adding `WithAvailableInertia` to `StatusSyncer` so it's done consistently for everyone, not ad-hoc per operator.
> 
> Can you dig into whether we have topology constraints on the console deployment? That feels like the most impactful fix here.


@jhadvig Thanks for the thorough review!

On suggestions 1 and 2 the console deployment already has both:

- Hard anti-affinity (RequiredDuringSchedulingIgnoredDuringExecution on kubernetes.io/hostname) so pods land on different nodes in HA mode
- PDB with maxUnavailable: 1 preventing voluntary drains from taking both pods down

The blip still happens because the conformance-serial tests do involuntary node reboots, which bypass the PDB. When the test hits the right nodes, both pods go offline for ~10 seconds regardless.

On option 3 - there's already an origin exception demoting this to a flake, but [OTA-362](https://issues.redhat.com/browse/OTA-362) is moving to remove those exceptions rather than add smarter ones. So the test layer is heading in the opposite direction.

The operator-level grace period was modeled after the sibling fixes ([OCPBUGS-24041](https://issues.redhat.com/browse/OCPBUGS-24041), [OCPBUGS-38676](https://issues.redhat.com/browse/OCPBUGS-38676), [OCPBUGS-64688](https://issues.redhat.com/browse/OCPBUGS-64688)), but your point about Available being intentionally real-time is a fair one. Would you prefer we pursue the library-go path instead — proposing inertia for Available on StatusSyncer so it's handled consistently across operators? Happy to go either direction.

sg00dwin:
> 4\. **Propose `Available` inertia in library-go** - if this is a real gap that other operators also hit, the proper fix is adding `WithAvailableInertia` to `StatusSyncer` so it's done consistently for everyone, not ad-hoc per operator.

Suggesting can we go ahead and merge this pr fix ... and I will open a follow-on issue to track the library-go `WithAvailableInertia` approach so all operators benefit long-term? Also, need to know what component to assign that bug to.

sg00dwin:
Verified on a CI cluster with the PR changes deployed — no regressions observed.

Cluster: OCP on AWS (ci-ln-2xqz9vt)

Steps performed:

1. Logged into the cluster and confirmed console operator healthy (Available=True, 2/2 replicas)
2. Reviewed the 3 changed files — new `evaluateDeploymentAvailability()` method adds a 15s grace period before reporting Available=False with InsufficientReplicas
3. Ran unit tests — all 19 pass including 7 new tests covering grace period suppression, expiry, first-sync, recovery, boundary, and error format
4. Verified consumer tests (deployment.IsAvailable, telemetry) — no regressions
5. Full `make test-unit` clean (gofmt, govet, all packages)
6. Checked cluster logs — first-sync correctly reported InsufficientReplicas immediately (no prior healthy state, no grace period applied)

/verified by @sg00dwin and Claude

PR #1182, opened 2026-07-06
OCPBUGS-58422: Handle disabled Ingress capability in HyperShift

stefanonardo:
## Summary

When Ingress capability is disabled in HyperShift, the console operator crashes repeatedly because it expects IngressController CRDs and Routes that don't exist. This PR makes the operator handle the ingress-disabled case gracefully.

### Changes

- **starter.go**: Detect ingress-disabled at startup via direct API calls to Infrastructure and ClusterVersion resources. Skip creating route and health check controllers, which prevents the IngressController informer from being created and eliminates reflector errors. If Ingress capability is later enabled, the operator restarts automatically (poll-and-restart pattern, same as OLM).
- **route/controller.go**: Reorder sync to check `IsExternalControlPlaneWithIngressDisabled()` before fetching IngressController (defensive).
- **oauthclients.go**: Add infrastructure and clusterVersion listers. Early-return when ingress is disabled and no ConsoleURL is configured.
- **clidownloads/controller.go**: Add infrastructure and clusterVersion listers (from existing configInformer factory). Early-return when ingress is disabled and no ClientDownloadsURL is configured, clearing the OCDownloadsSyncDegraded condition.
- **sync_v400.go**: Early-return when ingress is disabled and no ConsoleURL is configured, clearing the SyncLoopRefreshDegraded condition.

### Related

- openshift/hypershift#8933 — removes the HyperShift validation that blocks Ingress-only disable (depends on this PR)

### Test plan

- [x] `make test-unit` passes, including new tests for the poll-and-restart mechanism
- [x] Verified on a HyperShift hosted cluster on AWS with `--disable-cluster-capabilities Ingress` (using openshift/hypershift#8933 to remove the Ingress/Console coupling validation):
  - Before the fix: console operator logs show continuous `route "console" not found`, `IngressController not found`, and `FailedRouteGet` errors; ClusterOperator reports `Degraded=True`
  - After deploying the fix: console operator pod runs stable with 0 restarts, startup log confirms `Ingress capability is disabled in external control plane topology, skipping route and health check controllers`, zero route/IngressController errors in the logs, and `OAuthClientsControllerDegraded` condition is cleared from the ClusterOperator status

## Summary by CodeRabbit

* **New Features**
  * Added background polling for external control plane environments where ingress is disabled; routing components now start automatically once ingress becomes enabled (triggering a restart).

* **Bug Fixes**
  * OAuth client and CLI download syncing now detect ingress-disabled scenarios and exit early with updated progressing/degraded status, skipping route/ingress URL resolution.
  * Route syncing no longer resolves the default ingress controller when ingress is disabled, and route controllers aren't started in that mode.

### DIFF
--- a/pkg/console/controllers/oauthclients/oauthclients.go
+++ b/pkg/console/controllers/oauthclients/oauthclients.go
@@ -57,6 +57,8 @@ type oauthClientsController struct {
 	consoleOperatorLister       operatorv1listers.ConsoleLister
 	routesLister                routev1listers.RouteLister
 	ingressConfigLister         configv1lister.IngressLister
+	infrastructureConfigLister  configv1lister.InfrastructureLister
+	clusterVersionLister        configv1lister.ClusterVersionLister
 	targetNSSecretsLister       corev1listers.SecretLister
 }
 
@@ -67,6 +69,8 @@ func NewOAuthClientsController(
 	consoleOperatorInformer operatorv1informers.ConsoleInformer,
 	routeInformer routev1informers.RouteInformer,
 	ingressConfigInformer configv1informers.IngressInformer,
+	infrastructureConfigInformer configv1informers.InfrastructureInformer,
+	clusterVersionInformer configv1informers.ClusterVersionInformer,
 	targetNSsecretsInformer corev1informers.SecretInformer,
 	oauthClientSwitchedInformer *util.InformerWithSwitch,
 	recorder events.Recorder,
@@ -81,6 +85,8 @@ func NewOAuthClientsController(
 		consoleOperatorLister:       consoleOperatorInformer.Lister(),
 		routesLister:                routeInformer.Lister(),
 		ingressConfigLister:         ingressConfigInformer.Lister(),
+		infrastructureConfigLister:  infrastructureConfigInformer.Lister(),
+		clusterVersionLister:        clusterVersionInformer.Lister(),
 		targetNSSecretsLister:       targetNSsecretsInformer.Lister(),
 	}
 
@@ -138,6 +144,19 @@ func (c *oauthClientsController) sync(ctx context.Context, controllerContext fac
 	var consoleURL *url.URL
 
 	if len(operatorConfig.Spec.Ingress.ConsoleURL) == 0 {
+		infrastructureConfig, err := c.infrastructureConfigLister.Get(api.ConfigResourceName)
+		if err != nil {
+			return err
+		}
+		clusterVersionConfig, err := c.clusterVersionLister.Get(api.VersionResourceName)
+		if err != nil {
+			return err
+		}
+		if util.IsExternalControlPlaneWithIngressDisabled(infrastructureConfig, clusterVersionConfig) {
+			statusHandler.AddConditions(status.HandleProgressingOrDegraded("OAuthClientSync", "", nil))
+			return statusHandler.FlushAndReturn(nil)
+		}
+
 		routeName := api.OpenShiftConsoleRouteName
 		routeConfig := routesub.NewRouteConfig(operatorConfig, ingressConfig, routeName)
 		if routeConfig.IsCustomHostnameSet() {
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -61,6 +61,14 @@ func (co *consoleOperator) sync_v400(ctx context.Context, controllerContext fact
 	)
 
 	if len(set.Operator.Spec.Ingress.ConsoleURL) == 0 {
+		clusterVersionConfig, err := co.clusterVersionLister.Get(api.VersionResourceName)
+		if err != nil {
+			return statusHandler.FlushAndReturn(err)
+		}
+		if controllersutil.IsExternalControlPlaneWithIngressDisabled(set.Infrastructure, clusterVersionConfig) {
+			return statusHandler.FlushAndReturn(nil)
+		}
+
 		routeName := api.OpenShiftConsoleRouteName
 		routeConfig := routesub.NewRouteConfig(updatedOperatorConfig, set.Ingress, routeName)
 		if routeConfig.IsCustomHostnameSet() {
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -70,6 +70,15 @@ func (co *consoleOperator) sync_v400(ctx context.Context, controllerContext fact
 	)
 
 	if len(set.Operator.Spec.Ingress.ConsoleURL) == 0 {
+		clusterVersionConfig, err := co.clusterVersionLister.Get(api.VersionResourceName)
+		if err != nil {
+			return statusHandler.FlushAndReturn(err)
+		}
+		if controllersutil.IsExternalControlPlaneWithIngressDisabled(set.Infrastructure, clusterVersionConfig) {
+			statusHandler.AddConditions(status.HandleProgressingOrDegraded("SyncLoopRefresh", "", nil))
+			return statusHandler.FlushAndReturn(nil)
+		}
+
 		routeName := api.OpenShiftConsoleRouteName
 		routeConfig := routesub.NewRouteConfig(updatedOperatorConfig, set.Ingress, routeName)
 		if routeConfig.IsCustomHostnameSet() {
--- a/pkg/console/operator/sync_v400.go
+++ b/pkg/console/operator/sync_v400.go
@@ -75,7 +75,7 @@ func (co *consoleOperator) sync_v400(ctx context.Context, controllerContext fact
 			return statusHandler.FlushAndReturn(err)
 		}
 		if controllersutil.IsExternalControlPlaneWithIngressDisabled(set.Infrastructure, clusterVersionConfig) {
-			statusHandler.AddConditions(status.HandleProgressingOrDegraded("SyncLoopRefresh", "", nil))
+			statusHandler.AddConditions(status.HandleProgressingOrDegraded("WaitingForConsoleURL", "", nil))
 			return statusHandler.FlushAndReturn(nil)
 		}
 
--- a/pkg/console/starter/starter.go
+++ b/pkg/console/starter/starter.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"syscall"
 	"time"
 
 	// kube
@@ -15,6 +16,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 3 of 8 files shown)


pkg/console/starter/starter.go
```diff
@@ -246,6 +246,19 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
...
+	}
+	clusterVersionConfig, err := configClient.ConfigV1().ClusterVersions().Get(ctx, "version", metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	ingressDisabled := util.IsExternalControlPlaneWithIngressDisabled(infrastructureConfig, clusterVersionConfig)
+	if ingressDisabled {
+		klog.Info("Ingress capability is disabled in external control plane topology, skipping route and health check controllers")
+	}
+
```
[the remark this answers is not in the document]
stefanonardo: the check uses direct API calls, not informers, so Infrastructure and ClusterVersion are always available at that point.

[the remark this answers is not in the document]
jhadvig: The ingressDisabled check in starter.go is a one-shot startup gate... route and health-check controllers are never created when ingress is disabled. If Ingress capability is later enabled (via `additionalEnabledCapabilities` on ClusterVersion), CVO won't restart the console-operator pod since its Deployment is annotated with Console, not Ingress. The route controllers would remain unstarted until the pod is manually restarted or a cluster upgrade rolls the deployment.

The sync-time checks in oauthclients, clidownloads, and sync_v400 would pick up the change dynamically (they use informer listers), but without the route controller running there are no routes to resolve.

I think the cleanest fix would be to add a small watcher here in the console-operator's starter.go ... or check how other operators are dealing with this issue.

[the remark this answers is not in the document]
stefanonardo: @jhadvig you're right, I added a poll-and-restart function which is also used for OLM

sg00dwin:
@stefanonardo 
/lgtm I gave this PR a review...  
The poll-and-restart pattern matches the existing OLM precedent, the per-controller guards are consistent with each controller's own status patterns, and the utility function is well-tested. No blocking issues found.  You'll still need a repo owner approval.

pkg/console/starter/starter.go
```diff
@@ -693,7 +719,15 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
...
 		storageversionmigrationController,
-	} {
+	}
+	if !ingressDisabled {
+		controllers = append(controllers,
+			consoleRouteController,
+			downloadsRouteController,
+			consoleRouteHealthCheckController,
+		)
+	}
```
logonoff, line 729: Can you check my AI slop and verify that those controllers don't need to be in the list

<details>
<summary>
Slop report
</summary>
# console-operator: Network & Ingress Dependency Summary

## Direct Network Calls

Only **one** controller makes real outbound network connections:

| Controller | What It Does |
|---|---|
| `consoleRouteHealthCheckController` | HTTPS GET to the live console route URL. Custom `http.Client` with cluster CA pool. Up to 10 retries, 5s timeout per request, 1s between retries. |

All others interact exclusively through the kube-apiserver (informers, listers, typed clients). Several import `net/url` for URL parsing only. `resourceSyncer` imports `net/http` but only to serve an **inbound** debug handler — it makes no outbound calls.

---

## Controllers That Break If Ingress Is Disabled

"Ingress disabled" means the cluster router is gone (Routes never reach `Admitted=True`) and/or `config.openshift.io/v1/ingresses/cluster` is missing.

| Controller | What Breaks | Condition Set |
|---|---|---|
| **`consoleRouteHealthCheckController`** | Probes console route URL over HTTPS; if the route is never admitted, all 10 retries fail → **ClusterOperator `Available=False`** (most visible user-facing failure) | `RouteHealthDegraded=True`<br>`RouteHealthAvailable=False`<br>reason: `RouteNotAdmitted` |
| **`consoleOperator` (`sync_v400`)** | Calls `GetActiveRouteInfo` → `IngressURI`; on failure **aborts the entire sync loop** — ConfigMap, Deployment, `console.config.openshift.io` Status.ConsoleURL, and `console-public` ConfigMap all go stale or unset | `SyncLoopRefreshDegraded=True`<br>reason: `FailedIngress` |
| **`consoleRouteController`** | Console `Route` never transitions to `Admitted`; hostname never derivable downstream | `ConsoleDefaultRouteSyncDegraded=True`<br>reason: `FailedAdmitDefaultRoute` |
| **`downloadsRouteController`** | Downloads `Route` never admitted | `DownloadsDefaultRouteSyncDegraded=True`<br>reason: `FailedAdmitDefaultRoute` |
| **`oauthClientController`** | `redirectURIs` on `OAuthClient/console` never updated with the admitted console URL → **OAuth login to the console fails** | `OAuthClientSyncDegraded=True`<br>(via `WithSyncDegradedOnError`) |
| **`cliDownloadsController`** | `ConsoleCLIDownload/oc-cli-downloads` CR never created/updated; `oc` CLI download links broken | No named condition — bare error, sync retries silently |
| **`consoleServiceController`** | If `Ingresses/cluster` object is entirely missing, service creation fails | `ServiceSyncDegraded=True` |
| **`downloadsServiceController`** | Same as above | `ServiceSyncDegraded=True` |

---

## Controllers Not Affected by Ingress

| Controller | Reason Unaffected |
|---|---|
| `migrationCleanupController` | Deletes stale Deployment/Service/Secret only |
| `resourceSyncer` | Syncs ConfigMaps/Secrets via kube-apiserver only |
| `clusterOperatorStatus` | Aggregates conditions written by other controllers |
| `logLevelController` | Manages operator log level only |
| `managementStateController` | Manages operator management state only |
| `configUpgradeableController` | Checks unsupported config overrides only |
| `configObserver` | Watches FeatureGate/ClusterVersion via informers only |
| `consoleServiceAccountController` | Applies ServiceAccount objects only |
| `downloadsServiceAccountController` | Applies ServiceAccount objects only |
| `consolePDBController` | Applies PodDisruptionBudget only |
| `downloadsPDBController` | Applies PodDisruptionBudget only |
| `oauthClientSecretController` | Generates and writes OAuth client secret only |
| `oidcSetupController` | Reads `authentication/cluster`, syncs ConfigMaps only |
| `cliOIDCClientStatusController` | Updates authentication status conditions only |
| `upgradeNotificationController` | Reads ClusterVersion, manages ConsoleNotification only |
| `staleConditionsController` | Removes stale operator status conditions only |
| `storageversionmigrationController` | Manages StorageVersionMigration CRDs only |
| `downloadsDeploymentController` | Manages the downloads Deployment only (reads infra config, not ingress) |

---

## Designed Escape Hatches

### 1. Explicit URL overrides (bypasses `IngressURI` calls)

Setting these fields on the Console operator config skips route admission checks entirely:

| Field | Bypasses `IngressURI` in |
|---|---|
| `operatorConfig.Spec.Ingress.ConsoleURL` | `consoleOperator`, `oauthClientController`, `consoleRouteHealthCheckController` |
| `operatorConfig.Spec.Ingress.ClientDownloadsURL` | `cliDownloadsController` |

### 2. `IsExternalControlPlaneWithIngressDisabled()` (HyperShift)

When the cluster is HyperShift (`ExternalTopologyMode`) **and** the `Ingress` capability is disabled in `ClusterVersion.Status.Capabilities`:

- **`consoleRouteController` / `downloadsRouteController`** — short-circuit with `nil` (no route created, no error set)
- **`consoleServiceController` / `downloadsServiceController`** — switch service type from `ClusterIP` → `NodePort` so pods are reachable without a router

logonoff (commented):
Logic checks out I think

spadgett (approved):
/approve

pkg/console/starter/starter.go
stefanonardo: everything looks in place

pkg/console/operator/sync_v400.go
```diff
@@ -70,6 +70,14 @@ func (co *consoleOperator) sync_v400(ctx context.Context, controllerContext fact
...
 
 	if len(set.Operator.Spec.Ingress.ConsoleURL) == 0 {
+		clusterVersionConfig, err := co.clusterVersionLister.Get(api.VersionResourceName)
+		if err != nil {
+			return statusHandler.FlushAndReturn(err)
+		}
+		if controllersutil.IsExternalControlPlaneWithIngressDisabled(set.Infrastructure, clusterVersionConfig) {
+			return statusHandler.FlushAndReturn(nil)
+		}
+
```
jhadvig, line 81: The `oauthclients.go` and `clidownloads/controller.go` early-returns explicitly clear their conditions (`HandleProgressingOrDegraded("OAuthClientSync", "", nil)` and `HandleDegraded("OCDownloadsSync", "", nil)`).... this one just returns nil without clearing anything. Probably fine since ingress-disabled is a creation-time state on HyperShift, but worth keeping consistent with the other controllers.

pkg/console/starter/starter.go
```diff
@@ -237,6 +239,23 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
...
+	infrastructureConfig, err := configClient.ConfigV1().Infrastructures().Get(ctx, api.ConfigResourceName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	clusterVersionConfig, err := configClient.ConfigV1().ClusterVersions().Get(ctx, api.VersionResourceName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	ingressDisabled := util.IsExternalControlPlaneWithIngressDisabled(infrastructureConfig, clusterVersionConfig)
+	if ingressDisabled {
```
jhadvig, line 251: The PR description mentions "manifests: Gate openshift-ingress-operator namespace Role and RoleBinding with Console+Ingress capability annotation" but I don't see any manifest changes in this PR. I also couldn't find any console-operator manifests that target the `openshift-ingress-operator` namespace in the repo. Is the description stale, or are those changes still needed?

jhadvig (commented):
Nice approach overall @stefanonardo, the strategy of skipping route/healthcheck controllers entirely at startup and polling for ingress enablement is clean. Two things I noticed:

pkg/console/starter/starter.go
stefanonardo: oh yeah, I need to update the description. That role was deleted in the meanwhile in another commit.

pkg/console/operator/sync_v400.go
stefanonardo: I'll add it to make it consistent, however it's not a real bug as you explained

pkg/console/starter/starter.go
stefanonardo: updated

pkg/console/controllers/oauthclients/oauthclients.go
```diff
@@ -67,6 +69,8 @@ func NewOAuthClientsController(
 	consoleOperatorInformer operatorv1informers.ConsoleInformer,
 	routeInformer routev1informers.RouteInformer,
 	ingressConfigInformer configv1informers.IngressInformer,
+	infrastructureConfigInformer configv1informers.InfrastructureInformer,
+	clusterVersionInformer configv1informers.ClusterVersionInformer,
```
jhadvig, line 73: Both of these informares we need to add to the `WithInformers()` method in

pkg/console/operator/sync_v400.go
```diff
@@ -70,6 +70,15 @@ func (co *consoleOperator) sync_v400(ctx context.Context, controllerContext fact
 	)
 
 	if len(set.Operator.Spec.Ingress.ConsoleURL) == 0 {
+		clusterVersionConfig, err := co.clusterVersionLister.Get(api.VersionResourceName)
+		if err != nil {
+			return statusHandler.FlushAndReturn(err)
+		}
+		if controllersutil.IsExternalControlPlaneWithIngressDisabled(set.Infrastructure, clusterVersionConfig) {
+			statusHandler.AddConditions(status.HandleProgressingOrDegraded("SyncLoopRefresh", "", nil))
```
jhadvig: nit: Lets make the reason more specific

```suggestion
			statusHandler.AddConditions(status.HandleProgressingOrDegraded("WaitingForConsoleURL", "", nil))
```

jhadvig (commented):
two small nits, otherwise looks good to go 👍

pkg/console/controllers/oauthclients/oauthclients.go
stefanonardo: These informers are intentionally not in `WithInformers()`. They're only used for the `IsExternalControlPlaneWithIngressDisabled` gating check  which is a static condition at runtime. Adding them to `WithInformers()` would trigger a resync on every ClusterVersion/Infrastructure update event, which combined with `WithSyncDegradedOnError` writing back to operator status can cause unnecessary resync storms. I actually tried adding them during development and had to revert because of resync storms.

jhadvig: ok, so we will stick to to 1minut resync

jhadvig (approved):
/lgtm
/approve

PR #1197, opened 2026-07-24
OSDOCS-20110: Add Helm 3 CLI download links alongside Helm 4

sowmya-sl:
Helm 3 remains available for existing workflows. This adds a separate ConsoleCLIDownload resource so both versions appear on the CLI downloads page.

image

**Reviewers and assignees:**

## Summary by CodeRabbit

* **New Features**
  * Added Helm 3 CLI download information, including operational-continuity guidance and a link to the latest Helm 3 client bundle.
* **Updates**
  * Updated the Helm 4 CLI download description to highlight Helm 4.
  * Updated the Helm 4 download link to the newer Helm 4.2.3 client bundle.

### DIFF
--- a/manifests/07-downloads-helm.yaml
+++ b/manifests/07-downloads-helm.yaml
@@ -29,8 +29,10 @@ metadata:
     capability.openshift.io/name: Console
 spec:
   description: |
-    Helm 3 is a package manager for Kubernetes that simplifies deploying and managing
-    applications using Helm Charts. Helm 3 remains available for existing workflows.
+    Helm 3 is a package manager for Kubernetes applications which enables defining,
+    installing, and upgrading applications packaged as Helm Charts. Helm 3 continues
+    to be provided for operational continuity, but it will not be compatible with
+    future releases of OCP and so customers are encouraged to upgrade to Helm 4.
   displayName: helm - Helm 3 CLI
   links:
     - href: 'https://mirror.openshift.com/pub/openshift-v4/clients/helm/latest'
--- a/manifests/07-downloads-helm.yaml
+++ b/manifests/07-downloads-helm.yaml
@@ -10,8 +10,8 @@ metadata:
     capability.openshift.io/name: Console
 spec:
   description: |
-    Helm 4 is a package manager for Kubernetes applications which enables defining,
-    installing, and upgrading applications packaged as Helm Charts.
+    You can use Helm 4 to define, install, and upgrade applications
+    packaged as Helm Charts.
   displayName: helm - Helm 4 CLI
   links:
     - href: 'https://mirror.openshift.com/pub/cgw/helm/4.2.3'
@@ -29,10 +29,10 @@ metadata:
     capability.openshift.io/name: Console
 spec:
   description: |
-    Helm 3 is a package manager for Kubernetes applications which enables defining,
-    installing, and upgrading applications packaged as Helm Charts. Helm 3 continues
-    to be provided for operational continuity, but it will not be compatible with
-    future releases of OCP and so customers are encouraged to upgrade to Helm 4.
+    You can use Helm 3 to define, install, and upgrade applications
+    packaged as Helm Charts. Helm 3 is provided for operational
+    continuity but will not be compatible with future releases of OCP.
+    We encourage you to upgrade to Helm 4.
   displayName: helm - Helm 3 CLI
   links:
     - href: 'https://mirror.openshift.com/pub/openshift-v4/clients/helm/latest'
--- a/manifests/07-downloads-helm.yaml
+++ b/manifests/07-downloads-helm.yaml
@@ -14,5 +14,24 @@ spec:
     installing, and upgrading applications packaged as Helm Charts.
   displayName: helm - Helm 4 CLI
   links:
-    - href: 'https://mirror.openshift.com/pub/cgw/helm/4.1.4'
+    - href: 'https://mirror.openshift.com/pub/cgw/helm/4.2.3'
       text: Download Helm v4
+---
+apiVersion: console.openshift.io/v1
+kind: ConsoleCLIDownload
+metadata:
+  name: helm3-download-links
+  annotations:
+    include.release.openshift.io/hypershift: "true"
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    capability.openshift.io/name: Console
+spec:
+  description: |
+    Helm 3 is a package manager for Kubernetes that simplifies deploying and managing
+    applications using Helm Charts. Helm 3 remains available for existing workflows.
+  displayName: helm - Helm 3 CLI
+  links:
+    - href: 'https://mirror.openshift.com/pub/openshift-v4/clients/helm/latest'
+      text: Download Helm v3


manifests/07-downloads-helm.yaml
```diff
@@ -16,3 +16,22 @@ spec:
   links:
     - href: 'https://mirror.openshift.com/pub/cgw/helm/4.1.4'
```
webbnh: Consider updating this to 4.2.3.

(Presumably, this would be better done in a separate PR, but the additional hassle doesn't seem worthwhile, given the narrow scope and low complexity of this PR.)

manifests/07-downloads-helm.yaml
```diff
@@ -16,3 +16,22 @@ spec:
...
+  annotations:
+    include.release.openshift.io/hypershift: "true"
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    capability.openshift.io/name: Console
+spec:
+  description: |
+    Helm 3 is a package manager for Kubernetes that simplifies deploying and managing
+    applications using Helm Charts. Helm 3 remains available for existing workflows.
```
webbnh: Just out of curiosity, why does the first sentence use verbiage different from the Helm 4 text?

Regardless, given that we're including that second sentence, we might want to go all-in and say something like "Helm 3 continues to be provided for operational continuity, but it will not be compatible with future releases of OCP and so customers are encouraged to upgrade to Helm 4."  (I suggest conferring with Marina and Jocelyn on detailed content and wording.)

webbnh (approved):
This looks OK to me, but I've got a suggestion, a question, and a proposed revision (I'll let you guess which is which 😉).

/lgtm

manifests/07-downloads-helm.yaml
mkumku: So, we are concerned that having it on the console will imply support, we simply should redirect to the compatibility matrix then: https://access.redhat.com/articles/7145580 .
I would say, under both sections:
For Helm CLI compatibility with OpenShift releases refer to article https://access.redhat.com/articles/7145580.

webbnh: > we are concerned that having it on the console will imply support

@mkumku, we expect that this documentation is version-specific -- e.g., the change proposed here is specifically for OCP 5.0.  We will backport this change to OCP 4.22, and presumably it will propagate automatically to OCP 4.23.  For OCP 5.2 (and, perhaps for 5.1), we will remove the Helm v3 section.

So, while I have no objection to adding a reference to a _support_ statement, I don't think it's necessary; and, I don't think it's really useful to make a statement about _compatibility_ -- what we are advertising here is **_supposed to be_** compatible with the platform where it is being advertised!

webbnh: I don't know how these changes relate to the content served from https://console.redhat.com/openshift/downloads -- that content seems to be version independent, which poses exactly the compatibility challenges that you are raising.  If that content comes from here, then, yeah, we definitely need pointers from here to both compatibility and support documentation.  😛

sowmya-sl: So do I add a link to compatibility matrix or not? 😕

sowmya-sl: @mkumku @webbnh ?

mkumku: Hi Sowmya, since this is the OCP cluster, where you know the version of the cluster and you can adjust the version of Helm you are pointing to, then probably you can skip the link to the support matrix. 

My only question to you, since we cannot use `latest` for Helm v4, do you plan updating this link with each z-stream release of Helm? 
Maybe we can simply point to this directory: https://mirror.openshift.com/pub/cgw/helm/ . This directory will contain all the available Helm v4 binaries. And then this will remove the need of submitting a new PR with each z-release. And then it will make sense to share the compatibility matrix page.
I will leave this between you and @webbnh to decide what makes more sense from development workflow. Both options work for me.

webbnh (commented):
This looks good for now.  Let's merge it as is and then open another PR to work on the Helm v4 download link:  we should make it point to the directory which contains all the versions, and we should change the link text to something like "Helm v4 Downloads Directory".

/lgtm

webbnh:
@sowmya-sl (or @spadgett or anyone else with write access to the repo...), I believe that the Jira for this PR is [OSDOCS-20110](https://redhat.atlassian.net/browse/OSDOCS-20110).  If you could fix the title and ask the bot to confirm it, we can move this PR closer to merging....

jhadvig (approved):
/lgtm
/approve

manifests/07-downloads-helm.yaml
```diff
@@ -14,5 +14,26 @@ spec:
...
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    capability.openshift.io/name: Console
+spec:
+  description: |
+    Helm 3 is a package manager for Kubernetes applications which enables defining,
+    installing, and upgrading applications packaged as Helm Charts. Helm 3 continues
+    to be provided for operational continuity, but it will not be compatible with
+    future releases of OCP and so customers are encouraged to upgrade to Helm 4.
```
jseseCCS: 32-35: This descr mixes voice and is passive in places. --> active, second person --> "We continue to provide Helm 3 for operational continuity. However, it will not be compatible with future releases of OCP, so we encourage you to upgrade to Helm 4." 

MUST they upgrade to helm 4? or if they stick with helm 3, is it that they don't lose access to anything but they just won't be able to use newer versions of OCP?

last, if OCP isn't spelled out elsewhere in this file, expand it on first use.

manifests/07-downloads-helm.yaml
```diff
@@ -14,5 +14,26 @@ spec:
     installing, and upgrading applications packaged as Helm Charts.
```
jseseCCS: This descr is passive and third person. bc the new Helm 3 description below will likely mirror this text, worth fixing both for active voice and second person. for example --> "You can use Helm 4 to define, install, and upgrade applications packaged as Helm Charts."

manifests/07-downloads-helm.yaml
```diff
@@ -14,5 +14,26 @@ spec:
...
+    capability.openshift.io/name: Console
+spec:
+  description: |
+    Helm 3 is a package manager for Kubernetes applications which enables defining,
+    installing, and upgrading applications packaged as Helm Charts. Helm 3 continues
+    to be provided for operational continuity, but it will not be compatible with
+    future releases of OCP and so customers are encouraged to upgrade to Helm 4.
+  displayName: helm - Helm 3 CLI
+  links:
+    - href: 'https://mirror.openshift.com/pub/openshift-v4/clients/helm/latest'
```
jseseCCS, line 38: Helm 4 link above (line 17) points to specific version (4.2.3), but this Helm 3 link points to "latest." Intentional? If so, worth a brief note explaining why, for consistency?

jseseCCS (changes requested):
6 comments

manifests/07-downloads-helm.yaml
webbnh: It's a fair question.  The answer is that the project/product has historically been distributed via a `latest` link to save the customer from having to update their configuration and to save us from having to update the documentation.  However, RelEng won't allow us to have a `latest` link for the Helm v4 binaries (and they are only continuing the one for Helm v3 because customers quickly report it when it is broken), and, now that we support multiple versions of Helm and OCP, respectively, their reasoning is not unsound.  But, in terms of minimizing impact and labor, I'm inclined to continue the current practice of using the `latest` link for Helm v3 for as long as it lasts (which is not much longer, in release terms) and to use another approach for v4.

All that said, I don't really want to attract attention to the seeming inconsistency...I'd rather just go with the flow and let it quietly fade away.

manifests/07-downloads-helm.yaml
sowmya-sl: Changed description for active voice. 
For OCP, the whole thing is seen on OCP documentation. So I believe that the user does not need to know what it is.

manifests/07-downloads-helm.yaml
sowmya-sl: That's the plan in future. https://github.com/openshift/console-operator/pull/1197#pullrequestreview-4831565204

jhadvig (approved):
/lgtm
/approve

Spinning a verification cluster ATM to verify the fix

jhadvig:
The change is there:

Screenshot 2026-08-04 at 15 09 27


/verified by @jhadvig 

manifests/07-downloads-helm.yaml
```diff
@@ -10,9 +10,30 @@ metadata:
     capability.openshift.io/name: Console
 spec:
   description: |
-    Helm 4 is a package manager for Kubernetes applications which enables defining,
-    installing, and upgrading applications packaged as Helm Charts.
+    You can use Helm 4 to define, install, and upgrade applications
```
jseseCCS, line 13: this tells user what they can do with Helm 4 but drops expl of what Helm 4 is. keep both? -->Helm 4 is a package manager for Kubernetes. You can use it to define, install, and upgrade applications packaged as Helm Charts.

manifests/07-downloads-helm.yaml
```diff
@@ -10,9 +10,30 @@ metadata:
...
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    capability.openshift.io/name: Console
+spec:
+  description: |
+    You can use Helm 3 to define, install, and upgrade applications
+    packaged as Helm Charts. Helm 3 is provided for operational
+    continuity but will not be compatible with future releases of OCP.
+    We encourage you to upgrade to Helm 4.
```
jseseCCS, line 35: 32-35: A few things: 

- "We encourage you to upgrade to Helm 4" switches to 1st person/breaks 2nd-person voice elsewhere. It's also language that Red Hat reserves for partner policy guides and various certifications, not prod docs.. 

--> Upgrade to Helm 4 for continued compatibility with future OpenShift Container Platform (OCP) releases. (or similar) 

- we also can't use "we":

Image

- in your version, OCP isn't spelled out. if this is the only place it's used, --> "OpenShift Container Platform (OCP)" (or on first use).
- "will not be compatible with future releases of OCP" doesn't say which release or why. if that's known, much more helpful to name it, or link to where user can find out.
- "operational continuity" might not be clear to all users. change to plainer phrasing --> "...so your existing setup continues to work." ?

jseseCCS (commented):
back to you

jseseCCS:
**Note:** While this pull request was merged prior to docs sign-off, a couple key IBM/Red Hat style guide requirements remain unaddressed. This merge should not be construed as docs-approved, and I hope that at least some of the remaining style updates will be resolved in a future PR. cc @opayne1

manifests/07-downloads-helm.yaml
sowmya-sl: Please suggest a suitable string not using "We". Does "You are encouraged to upgrade to Helm 4  .." work?
 


> in your version, OCP isn't spelled out. if this is the only place it's used, --> "OpenShift Container Platform (OCP)" (or on first use).

But this whole page sits on the OCP console. I believe there are enough references to the full form on the Console

> "will not be compatible with future releases of OCP" doesn't say which release or why. if that's known, much more helpful to name it, or link to where user can find out.

Maybe a link to compatibility matrix works?

jseseCCS: what about "You can use Helm 3 to define, install, and upgrade applications packaged as Helm charts. Helm 3 is still available for operational continuity, but it will not be compatible with future releases of OpenShift Container Platform (OCP). Upgrade to Helm 4 for continued compatibility with future OCP releases."?

PR #1201, opened 2026-07-30
OCPBUGS-99943: set API-server-defaulted fields on deployment specs to prevent excessive update events

sg00dwin:
## What

Set API-server-defaulted fields explicitly in both deployment YAML templates (`console-deployment.yaml`, `downloads-deployment.yaml`) and in the non-HA rolling update strategy code to prevent pathological `DeploymentUpdated` events.

**Fields added to YAML templates:**
- `progressDeadlineSeconds: 600`
- `revisionHistoryLimit: 10`
- `dnsPolicy: ClusterFirst`
- `serviceAccount: <name>`

**Code change:**
- `withStrategy()` now sets `MaxSurge: "25%"` and `MaxUnavailable: "25%"` for non-HA topologies instead of an empty struct

## Why

The console-operator generates pathological `DeploymentUpdated` events in a tight reconciliation loop, causing **79% of test failures** (15/19 runs) in `periodic-ci-openshift-release-main-nightly-5.0-e2e-vsphere-ovn-techpreview-serial` with a **1.9% overall pass rate**.

Library-go's `ApplyDeploymentWithForce` replaces the entire deployment spec on every metadata-triggered update (e.g., when a ConfigMap ResourceVersion changes due to plugin registration). Fields the operator didn't set are written as null, and the API server re-defaults them — creating unnecessary spec churn that doubles the update event count per trigger. With ~14 plugin registrations in a TechPreview test run, this produces ~30 events, exceeding the pathological threshold.

Setting these fields explicitly makes the operator's desired spec match what the API server already stores, eliminating the null-field overwrite. Each legitimate trigger now produces exactly 1 deployment update instead of ~2.

All values are existing API server defaults — no behavioral change.

## Verification

- Spec hash verified locally: required spec hash matches API-server-defaulted spec hash for both HA and SingleNode topologies
- All unit tests pass (`make test-unit`)
- Build succeeds

Fixes [OCPBUGS-99943](https://issues.redhat.com/browse/OCPBUGS-99943)

Assisted by: Claude Code (Opus 4.6)

## Summary by CodeRabbit

- **Improvements**
  - Deployment rollouts now use explicit progress deadlines and retain a defined number of previous revisions.
  - Standard rolling updates now use balanced percentage-based surge and availability settings for smoother deployments.
  - Console and downloads workloads now apply their designated service accounts consistently.
  - Deployment networking and service account configuration are more explicit, improving predictability across environments.

### DIFF
--- a/bindata/assets/deployments/console-deployment.yaml
+++ b/bindata/assets/deployments/console-deployment.yaml
@@ -7,6 +7,8 @@ metadata:
     app: console
     component: ui
 spec:
+  progressDeadlineSeconds: 600
+  revisionHistoryLimit: 10
   selector:
     matchLabels:
       app: console
@@ -23,10 +25,12 @@ spec:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
         openshift.io/required-scc: restricted-v2
     spec:
+      dnsPolicy: ClusterFirst
       nodeSelector:
         node-role.kubernetes.io/master: ""
       restartPolicy: Always
       serviceAccountName: console
+      serviceAccount: console
       schedulerName: default-scheduler
       securityContext:
         runAsNonRoot: true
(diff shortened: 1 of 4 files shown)


jhadvig (approved):
/lgtm
/approve

